PR #568, opened 2018-12-03
add `free` command to see docker memory profile when its run

dougli1sqrd:


cmungall:
Any closer to this @dougli1sqrd? Maybe the SPARQL can be written in a more efficient way? Or should we just have a special purpose procedure in ROBOT, cc @balhoff

balhoff:
What’s this related to? I don’t see anything linked.

balhoff:
I can't find any way to speed up the SPARQL.

balhoff:
I did notice that the `robot` commands in the MONDO makefile don't seem to have `ROBOT_JAVA_ARGS` specified. That didn't make a difference on my laptop, but is it possible the Travis default max heap is smaller?

dougli1sqrd:
The latest commit cleans up the travis config. At this point this travis run should clear and we should be good.

PR #826, opened 2019-09-18
Patch foodon unsat

matentzn:
Removes unsats from foodon import and removes foodon import from make pipeline (temporarily). There is another PR to undo this change when https://github.com/FoodOntology/foodon/issues/69 is fixed.


matentzn:
New foodon does not have this problem, redoing import is enough.

PR #828, opened 2019-09-18
Update foodon_import.owl

matentzn:
updated foodon import to latest version using `make imports/foodon_import.owl` (and obo)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,7 @@ before_script:
   - cd src/ontology
 
 script:
-  - echo "SKIPPING TRAVIS REAL BUILD TO FORCE SUCCESS"
-  #- ./run.sh make test
+  - ./run.sh make test
 
 # whitelist
 branches:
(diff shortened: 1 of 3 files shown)


matentzn:
(I also reactivated the travis job to see wether refreshing restores the passing test)

.travis.yml
```diff
@@ -31,8 +31,7 @@ before_script:
   - cd src/ontology
 
 script:
-  - echo "SKIPPING TRAVIS REAL BUILD TO FORCE SUCCESS"
-  #- ./run.sh make test
+  - ./run.sh make test
```
cmungall: I believe @ShahimEssaid deliberately skips the travis test. I will merge this PR and we can take out the test later if need be

matentzn: Yeap sorry,did that again.forgot that this was the whole reason we had the jenkins.

ShahimEssaid:
There is no harm in re-enabling Travis. The main reason I commented that
out is to make sure the Travis build succeeds every time so the overall
status is only based on Jenkins' status. Travis used to fail due to timeout
and this was always failing the overall status (the worse of multiple
checks is the overall status) and it was forcing someone to open the status
marker to be able to notice if Jenkins' status is success.

matentzn:
Moving discussion back [here](https://github.com/monarch-initiative/mondo/issues/657)

PR #876, opened 2019-10-08
Add analysis.md to docs

deepakunni3:
Add analysis.md and artifacts/all-mondo-disease-terms.tsv to docs in preparation for 'How many rare disease are there?' commentary.


jmcmurry:
looks great, thanks

PR #971, opened 2019-11-15
Issue 808 fix do syn clashes

cmungall:
see #808


cmungall:
@nicolevasilevsky  can you do a quick sanity check, don't spend > 2 mins thx

PR #1851, opened 2020-08-08
making the Mondo pipline executable with ODK

matentzn:
Changes:

1. Added 4 files in the import directory that were previously uncommitted. Maybe not necessary, but maybe it is, and since all the other files are committed (for the other ontologes), lets commit these as well to avoid confusion.
2. More than 30 files added to .gitignore. 
3. I added an empty d2p.pro, so when someone runs the pipeline, this wont show up in their git diff. You can always overwrite this if you want @cmungall
3. I had to overwrite 5 make goals for various reasons. Please review Makefile in diff carefully. I will try to fix the two imports, but right now, that is what it is.


src/ontology/Makefile
```diff
@@ -1002,3 +1002,21 @@ bg-clean:
 
 succeed:
 	echo make succeeded
+	
+
+###### NICO MIGRATION: overwriting unsuccessful goals!
+
+../sparql/construct-ecs-from-xrefs.sparql:
```
cmungall, line 1009: this is unnecessary - you only needed to do this as you didn't have the sparql folder mounted in Docker

src/ontology/Makefile
```diff
@@ -1002,3 +1002,21 @@ bg-clean:
...
+	echo "STRONG WARNING STRONG WARNING make $@ is skipped!" && touch $@
+	
+d2p.pro:
```
cmungall, line 1012: harmless but uneccessary. You can simply remove all mention of this. It is not in the build graph for make all

src/ontology/Makefile
```diff
@@ -1002,3 +1002,21 @@ bg-clean:
...
+	echo "STRONG WARNING STRONG WARNING make $@ is skipped!" && touch $@
+	
+reports/semantic-xref-pairs.tsv:
+	echo "STRONG WARNING STRONG WARNING make $@ is skipped!" && touch $@
+	
+imports/ncbitaxon_import.owl:
```
cmungall, line 1018: but we need to be able to make this!!!

src/ontology/Makefile
```diff
@@ -1002,3 +1002,21 @@ bg-clean:
...
+	echo "STRONG WARNING STRONG WARNING make $@ is skipped, because of out of memory errors with 16GB memory!" && touch $@
+	
+imports/hgnc_import.owl:
```
cmungall, line 1021: we need a ticket for this, this is important

for now we can simply use an archive version. My copy in mirror/ is fine, I can send

src/ontology/Makefile
matentzn: d2p.pro -> reports/semantic-xref-pairs.tsv -> $(ADDITIONAL_REPORT_FILES) -> $(REPORT_FILES) -> $(ARTEFACTS)

src/ontology/Makefile
matentzn: We will, I am on it.

src/ontology/Makefile
matentzn: Fixing this as we speak see slack.

matentzn: https://github.com/monarch-initiative/mondo/pull/1853

src/ontology/Makefile
matentzn: See email, not sure thats the real problem here.

src/ontology/Makefile
matentzn: Fixed.. Keeping the max memory very low actually helped (probably hit some docker max im my case because I had other stuff running).

PR #1852, opened 2020-08-08
addig SO import

matentzn:
see #1848


src/ontology/imports/so_terms.txt
```diff
@@ -0,0 +1,18 @@
...
+http://purl.obolibrary.org/obo/SO_0001268
+http://purl.obolibrary.org/obo/SO_0001272
+http://purl.obolibrary.org/obo/SO_0001411
+http://purl.obolibrary.org/obo/SO_0001500
+http://purl.obolibrary.org/obo/SO_0001637
+http://purl.obolibrary.org/obo/SO_0001877
+http://purl.obolibrary.org/obo/SO_1000008
+http://purl.obolibrary.org/obo/SO_1000029
+http://purl.obolibrary.org/obo/SO_1000037
+http://www.geneontology.org/formats/oboInOwl#ObsoleteClass
```
cmungall, line 18: kinda odd

matentzn: Gets regenerated every time, and your current pipeline matches anything with so in the id.. I dont think its a problem though..

PR #1953, opened 2020-09-05
imports mechanism: remove-seed and cosmetics

matentzn:
Changes:
1. Adding a _global_ remove list that will remove entities from _all_ imports
2. Refreshed ro_import for the purpose of illustration
3. Added a way to SKIP import generation during make process
4. Added a tiny fix for subsets (ODK standard)


matentzn:
See #1942

cmungall (approved):

PR #2033, opened 2020-09-24
MAJOR: Changing mondo.owl

matentzn:
This PR:

- removes equivalencies.owl from mondo.owl
- introduces a new artefact called mondo-with-equivalents.owl (which corresponds to the old mondo.owl)
- renames merged.owl (an intermediate artefact) to reasoned-plus-equivalents.owl (as indicated by Makefile todo item)
- turns mondo.owl into a merged product, getting rid of imports altogether
- removing sparql violation tests on monarch.owl - because of the merge, all tests fail now; the tests on the edit file and monarch.obo should be sufficient, and the remaining errors will be dealt with by the new quality control workflow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,6 @@ src/patterns/imports/seed_sorted.txt
 src/ontology/mondo-qc.*
 src/ontology/tmp/
 src/ontology/mondo_current_release.owl
+src/ontology/mondo-with-equivalents.*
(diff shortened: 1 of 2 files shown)


matentzn:
- [ ] Remember to update monarch.owl
- [ ] Remember to offer equivalencies.owl as a separate release product

cmungall (approved):

PR #2091, opened 2020-10-06
OLD: Omim makefile 1339

matentzn:


matentzn:
@cmungall Old branch, I would like to delete this (some 800 lines of undocumented changes to mondo.obo)

PR #2268, opened 2020-11-07
Mondo - add tags in

matentzn:
This is the first draft of adding the pattern tagging in. 

Here a very interesting example:
image


cmungall:
great!

I'm approving but of course this will also require extending catalog, and registering a new purl folder

cmungall (approved):

PR #2573, opened 2021-01-29
fixes all remaining patterns

matentzn:


matentzn:
@nicolevasilevsky somehow there is a conflict now; did you edit src/patterns/scratch/genetic.yaml?

You should only edit this branch here! :)

PR #2611, opened 2021-02-04
Mass removal of related syns where there are exact syns

matentzn:
After a review by @nicolevasilevsky we decided to go ahead with clearing out the related synonyms where there are exact ones.

We are doing it in two steps. This PR:
1. Gets rid of all related synonyms where there is an exact synonym
2. Moves the annotations over (note there are some spurious other changes because I ran "NORM")

I would suggest this:
1. @nicolevasilevsky you review this with some care (I would say 20 min at least)
2. You pass on your observations to @cmungall who gives the final ok.


cmungall (approved):

nicolevasilevsky (approved):

cmungall:
all looks good. go ahead and merge when you need to. probably want to do it
early to avoid merge conflicts

PR #2622, opened 2021-02-05
Fix imports by removing individuals

matentzn:
This PR removes individuals from all imports, unless they are explicitly used an a logical definition. Good idea?

Review in particular the makefile.
ENVO and ECTO are refreshed according to the new make recipe.


matentzn:
adresses https://github.com/monarch-initiative/mondo/pull/2622

cmungall (approved):

PR #2655, opened 2021-02-11
Issue 2618 Predisposes to X should never be subclass to X

jiaola:
Addresses issue #2618


jiaola:
The validation sparql found some ones that violate the rule. Not sure who should fix the ontology.

Issue #2619, opened 2021-02-05
Mondo Pattern Library 2.0


matentzn:
All our logical patterns can be found [here](https://github.com/monarch-initiative/mondo/tree/master/src/patterns/dosdp-patterns). We want to achieve two things:

1. Standardise the patterns a bit better
2. Create beautiful, user friendly documentation for each of them

## Standardisation

Standardisation can be a bit tedious, and once we have a schema, @nicolevasilevsky and I can help with this.

- [X] Standardising the pattern order. (I have done that, the order should be ["pattern_name", "pattern_iri", "description", "contributors", "classes", "relations", "annotationProperties", "vars", "name", "annotations", "def", "equivalentTo"])
- [x] Standardising the pattern variable names. As you can see [here](https://github.com/monarch-initiative/mondo/blob/baa5a0310510f841ebe077191b3922233504ab77/src/patterns/dosdp-patterns/adenocarcinoma.yaml#L25), some patterns have ugly variable names, we really want to standardise. Standardise means: if the slot is a `disease` the variable should be named `disease` ([example](https://github.com/monarch-initiative/mondo/blob/baa5a0310510f841ebe077191b3922233504ab77/src/patterns/dosdp-patterns/allergic_form_of_disease.yaml#L31)). If the variable refers to a gene, the variable should be called "gene".
- [ ] Standardising the classes, relations, annotationProperties fields. Standardising means: the slot should use the appropriate id as a value. Good example [here](https://github.com/monarch-initiative/mondo/blob/baa5a0310510f841ebe077191b3922233504ab77/src/patterns/dosdp-patterns/allergic_form_of_disease.yaml#L19).
- [ ] Every pattern should have a good definition. Basically you paste a pattern in slack, and @nicolevasilevsky will come up with a description you can add. 
 
 ## Shiny Pattern Documentation System (dp2docs)
 
 Here, we want to have a beautiful markdown based page in our [readthedocs](https://mondo.readthedocs.io/en/latest/editors-guide/e-design-patterns/#design-patterns-for-neoplasms). We don't know yet the perfect way to do this, but in essence, we want:
 
 - [ ] A way to generate a great overview for patterns like [README.md](https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/README.md), which is generated by [this](https://github.com/monarch-initiative/mondo/blob/master/src/scripts/patterns_create_overview.py) script in our read the docs (a new sub page).
[...]

PR #2728, opened 2021-02-25
Issue 2619

jiaola:
Addressed this checkbox in #2619 

> Standardising the pattern variable names. As you can see here, some patterns have ugly variable names, we really want to standardise. Standardise means: if the slot is a disease the variable should be named disease (example). If the variable refers to a gene, the variable should be called "gene".

Fixed variable names. 
Added pattern_iri in hereditary.yaml
Regenerated README


matentzn (approved):
Nicely done! Will discuss more in the QC call later!

PR #2787, opened 2021-03-09
Issue 2619

jiaola:
Added patterns docs 

Check https://jiaola-test.readthedocs.io/en/latest/editors-guide/patterns/ for a preview


matentzn:
Awesome, looking great! Sooo glad this is happening! Some suggestions:
- instead of `{location(owl:Thing)}`, can we render location [location](http://www.w3.org/2002/07/owl#Thing "owl:Thing") (no curly brackets, linking to the full IRI and a tooltip with the ID?)
- Instead of 
`annotation: exact_synonym(oio:hasExactSynonym)`
`text: {location(owl:Thing)} adenoma`
Maybe render:
`oio:hasExactSynonym:  `[location](http://www.w3.org/2002/07/owl#Thing "owl:Thing")` adenoma`
- Can we get the descriptions to respect line breaks? Example: https://jiaola-test.readthedocs.io/en/latest/editors-guide/patterns/adenocarcinoma/ (the word `Examples:` in the description should be on a new line.
- The overview page should probably be a table with the name and the description (same rendering as the individual pages).

Thaaanks! @nicolevasilevsky if you have further ideas, feel free to comment!

matentzn (approved):
@jiaola This is great!

I am happy with this as a first version of the pattern docs! Let have @cmungall review this as well, maybe he has some more ideas! Very nice.

matentzn:
some more ideas (dont implement, lets wait for Chris to check)
- only put {} when we are talking about a variable (i.e. `%s`). Else just don't add the {}
- Make the table on the overview page fit the width restrictions of the rtd theme

PR #2837, opened 2021-03-22
Terms merged bug

matentzn:
fixes #2837 

The problem here is that `alt_ids` introduced by some automatic process (@balhoff and @nicolevasilevsky say owltools) cause issues in conjunction with the obo serialistion..


matentzn:
Closing in favour of #2838

PR #2920, opened 2021-04-13
Last fixes for ROBOT report

matentzn:
- replaced obsolete NCBITaxon terms with their replacements
- merged/removed the last two duplicate_scope synonyms
- renamed abscess to abscess (disease)

PR #2921, opened 2021-04-13
DO NOT merge: just to run some tests

matentzn:
Testing #2915 #2920 together


matentzn:
This was just a (successful test) wether #2920 actually fixes all remaining issues. It does!

PR #2997, opened 2021-04-30
Construct query (does not work yet)

### LINKED ISSUE #2979: Synonyms that equal a label should be removed
See [here](https://api.triplydb.com/s/5NLruOV_O)

```
prefix owl: <http://www.w3.org/2002/07/owl#>
prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT ?term ?related
WHERE 
{ 
  { 
    ?term oboInOwl:hasRelatedSynonym ?related ;
      rdfs:label ?related ;
      a owl:Class .
      [
         a owl:Axiom ;
           owl:annotatedSource ?term ;
           owl:annotatedProperty oboInOwl:hasRelatedSynonym ;
           owl:annotatedTarget ?related ;
           oboInOwl:hasDbXref ?xref ;
      ] .
    
    FILTER (isIRI(?term) && regex(str(?term), "^http://purl.obolibrary.org/obo/MONDO_"))
  }
}
```

There are about 500 such cases. The solution should be easy:

Just delete the synonym! (Right @cmungall?)

Thank you @kshefchek for reporting.

matentzn:
fixes #2979

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -254,6 +254,9 @@ related_annos_to_exact:
 rm_related_annos_to_exact:
 	$(ROBOT) query --use-graphs false -i $(SRC) --update $(SPARQLDIR)/rm-related-exact-synonym-annotations.ru -o $(SRC)
 
+mark_syns_equal_to_label_for_exclusion:
+	$(ROBOT) query --use-graphs false -i $(SRC) --update $(SPARQLDIR)/construct-exclude-syn-equal-label.sparql convert -f obo -o $(SRC)
+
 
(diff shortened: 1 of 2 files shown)


matentzn:
After discussion with @cmungall we need to move this pipeline to be executed at release time only

matentzn:
@twhetzel we need to decide if we still want to do this.. Maybe put on tech call agenda

twhetzel:
Discussed on 8-Dec-2023 Tech call, but no firm decisions.

PR #2998, opened 2021-04-30
Marking synonyms for exclusion from Mondo release if they are equal to label

matentzn:
related #2996 
related #2979


matentzn:
@cmungall is against this, we will do it instead at release time.

PR #3002, opened 2021-04-30
Create poisoning.tsv

laurenechan:
Moving poisoning terms into new pattern from environmental_stimulus pattern

### DIFF
--- a/src/ontology/imports/ecto_import.owl
+++ b/src/ontology/imports/ecto_import.owl
@@ -11,7 +11,7 @@
      xmlns:swrlb="http://www.w3.org/2003/11/swrlb#"
      xmlns:oboInOwl="http://www.geneontology.org/formats/oboInOwl#">
     <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/mondo/imports/ecto_import.owl">
-        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/mondo/releases/2021-04-30/imports/ecto_import.owl"/>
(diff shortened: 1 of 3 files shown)


cmungall (commented):
what is the difference between poisoning and toxicity?

if this is meaningful we should expand the pattern so make this explicit

I suspect it is not..?

laurenechan:
Toxicity appears to actually be describing the qualities of a substance (e.g. the LD50) or [the quality of being poisonous/toxic.](https://www.merriam-webster.com/dictionary/toxicity) 
For the methotrexate toxicity, that term label may actually be a misnomer as it would be referring to how toxic methotrexate is as a substance, right? 

Versus poisoning refers to[ a substance that through its chemical action usually kills, injures, or impairs an organism](https://www.merriam-webster.com/dictionary/poisoning), so we may actually want to use poisoning as opposed to toxicity when referring to a disease state. 

Arguably though, they are used interchangeably within the literature. Including them in the same pattern does make sense and it's maybe up to user preference as to whether to call a term toxicity or poisoning.

maglott:
I found this useful  https://en.wikipedia.org/wiki/Toxicity
Several sites made the point that toxicity is dose-dependent. I'm sure the usage is fuzzy and I don't know if different rules apply.  For alleles that affect drug reponses, PharmGKB has reported values like 'cyclophosphamide response - Toxicity/ADR' for the phenotype.   I thought previous discussions had suggested that reactions to substances were not in scope for Mondo and rather potentially for HPO.

laurenechan:
What are the upper levels of poisoning/hierarchy we can use? 
intentional vs unintentional exposures, acute vs chronic, biological vs element, are toxicity and poisoning different? 
AI: read through a few review articles to see about systems of classifications of poisoning we can borrow from.

PR #3004, opened 2021-04-30
Create substance_dependence.yaml

laurenechan:


cmungall (changes requested):
This DP has two axes of classification, but it looks like you just intend to have one (the entity)

 - I prefer role names not material names for vars, but still discussing this with @matentzn and @nicolevasilevsky . E. g 'dependence on' rather than 'material entity'
 - Use a more specific class than BFO. Is this ever not a chemical?

when making new yaml requests I recommend using dosdp-tools to see how many instances we have already

cmungall:
ok now I see #3008 it looks like you vary the genus a bit between substance and drug... things get confusing with chebi roles, may take more discussion but I think it should be a fixed genus and we should infer drug dependence

(is tehre any case of cocaine dependence when cocaine is not used as a drug?)

is drug dependence even a useful subclass of substance dependence?

laurenechan:
@cmungall 
> ok now I see #3008 it looks like you vary the genus a bit between substance and drug... things get confusing with chebi roles, may take more discussion but I think it should be a fixed genus and we should infer drug dependence

I've altered the vars to just include the 'stimulus' and have made the genus fixed as substance dependence. 

> (is tehre any case of cocaine dependence when cocaine is not used as a drug?)

If you're asking whether there are literature supported medical applications for cocaine that could lead to dependence, I am not aware of any. But that differs from other chemicals like MDMA that are classified as drugs but are being trialed as therapeutics. Not sure if that is what you were considering...

> is drug dependence even a useful subclass of substance dependence?

From discussions with @nicolevasilevsky , it seems as though there are a lot of xrefs for drug dependence, so likely it is a useful class even just for the sake of harmonization.

PR #3173, opened 2021-06-24
Create idiopathic.yaml

sabrinatoro:
Need to get the idiopathic ID


src/patterns/dosdp-patterns/idiopathic.yaml
```diff
@@ -0,0 +1,46 @@
...
+  vars:
+  - disease
+
+def:
+  text: Idiopathic form of %s.
+  vars:
+  - disease
+
+equivalentTo:
+  text: '%s and ''has modifier'' some ''idiopathic'''
```
cmungall, line 44: very minor: I find it more readable and less error prone to use `"` as outer quote and `'` for quoting of labels

nicolevasilevsky (approved):

PR #3230, opened 2021-07-02
Deleted empty 'consider' axiom on MONDO:0016606.

balhoff:
This was causing a problem for the OWL API OBO parser in dosdp-tools (https://github.com/INCATools/dosdp-tools/issues/339). Also it seems to just be an error.


balhoff:
The axiom was added in d2f24ea34ad3e8ed3340a99a31388cc30a7bac1b.

PR #3233, opened 2021-07-02
Add owl axioms check

matentzn:
Makes sure that we dont accidentally introduce non-obo axioms..


matentzn:
@nicolevasilevsky if test passes feel free to merge

cmungall (approved):

PR #3241, opened 2021-07-04
New pattern matches July 4

matentzn:
@nicolevasilevsky 

I will run these (automatic matches of our logical definitions to patterns) from time to time - you should look through all the diffs and spot oddities, like major unintended changes etc..


cmungall (approved):
super useful, we can just do this automatically with actions in future

PR #3244, opened 2021-07-08
Fix bone resorption definition

### LINKED ISSUE #502: osteoporosis
This is now classified as a bone resorption disease which is wrong. It is a bone turn over disease and lack of new bone formation is prob most important. This is the danger of classifying by pathophysiology!!

kallia-p:
fixes #502


kallia-p:
@nicolevasilevsky The PR Fix bone resorption definition #3244 is simply to fix the definition of bone resorption as suggested by Chris - it doesn't address the proposed re-structuring of bone disease branch.

PR #3345, opened 2021-07-22
monochrom

cmungall:


cmungall:
@matentzn I removed the disabling of echoing (`@`), I like echoing, otherwise impossible to debug - OK?

matentzn:
100% we did that for all of odk repos, just forgot for mondo

nicolevasilevsky (approved):

PR #3372, opened 2021-07-23
added Equivalent axiom to 'autosomal dominant cataract' and added sub…

### LINKED ISSUE #2951: MONDO:0022672  autosomal dominant cataract [Revise subclass]
**Mondo term (ID and Label)**
MONDO:0022672  autosomal dominant cataract

Maybe Move this to be a direct child of  MONDO:0005129 "cataract (disease)" then set it to have children from the MIM PS116200 that are autosomal dominant inheritance? 
(MESH has a supplementary concept for Cataract, autosomal dominant, but it treats it as syn. with cataract 9, multiple types (MONDO:0011413), but this seems incorrect. )

Optionally: Obsolete this term. This GARD page is basically empty, it's hard to tell what disorder this citation is referring to. 

sabrinatoro:
…class

Fix #2951

The following terms were inferred as children based on their subclassification to 'systemic disease with cataract': autosomal dominant Alport syndrome (MONDO:0007086), autosomal dominant rhegmatogenous retinal detachment (MONDO:0016202).

Note: I wasn't sure which 'source' to add


nicolevasilevsky:
hi - sorry I didn't review your nice [spreadsheet](https://docs.google.com/spreadsheets/d/1pF_kJodM50hTW_tV74pJZb6mKFcGoDe9C5vrWXYP5Cc/edit#gid=729586216) earlier. I like option #2 to add the superclassOf axioms, then we'll get autoclassifications. And in answer to your question on the spreadsheet, yes, we can add superclassOf axioms like that.

here is an example:
'early-onset autosomal dominant Alzheimer disease'
image

nicolevasilevsky (changes requested):
per my comment on the 'conversation', could you please add subclassOf axioms instead of asserting the superclasses?
thank you!

sabrinatoro:
changes done

nicolevasilevsky (approved):

Issue #3403, opened 2021-07-29
Import and use SO to axiomize chomosome anomaly 'type'


sabrinatoro:
Some 'partial chromosome deletion' terms use the SO term in subclass: 'has modifier' some chromosomal_deletion.
Assuming we want to follow the same pattern, we should
- [x] import SO terms required for the axiomization ([here](https://docs.google.com/spreadsheets/d/1eECpDgSlbD9R326z2dI9WtG7ud8oZ6Dbyqr2hO2uYyo/edit?usp=sharing )
- [ ] add axioms 'has modifier' some (SO term) to existing terms

sabrinatoro:
Closing this issue as the use of SO to maximize will be address in specific issues for part of the ontology

PR #3409, opened 2021-07-29
added SO terms

sabrinatoro:
Address #3403


sabrinatoro:
@nicolevasilevsky, could you please review this? 
I will address the creation of axioms using SO in separate issues.
Thank you

nicolevasilevsky (approved):

PR #3514, opened 2021-08-12
Addresses #616

rays22:
Added new class 'migratory muscle precursor' as an exercise.


matentzn:
@rays22 thank you! which ticket does this pertain to? This seems like.. an anatomy term? Or did you simply just add _any_ term without reference to a specific ticket, (because I see you making some reference to what seems to be unrelated in your commits).

rays22:
Thanks @matentzn for checking it.
Yes, it is just adding a dummy term to test the pull request process of Mondo. The ticket that the pull request references is indeed unrelated.
It was part of my preparations for the Mondo training yesterday as requested by Nicole in her email.
I was planning to delete both the `excercise1` and `test812` branches after the merge, but I guess I can delete them now so they do not cause any further confusions.

PR #3519, opened 2021-08-13
created new sparql querries

sabrinatoro:


sabrinatoro:
I don't think the second commit was correct based on the tsv.
But I don't know how to "uncommit" it.

sabrinatoro:
Note: 
1) the terms which parents are in the Harrison View but inferred are reported as "not-harrison". They should be reported in the "harrison". 
2) it would be helpful to have the parents of the terms in the report

nicolevasilevsky (approved):
Sabrina - I approved this so you can merge it whenever it is ready.

PR #3744, opened 2021-10-01
OMIM: Changed source URL

joeflack4:
Updates
- src/sources/omim/Makefile
  - Changed url to omim.ttl file
  - Changed command to fetch this file


matentzn:
@joeflack4 One important thing to consider is that the slurp is not automatically successful if the slurp command runs. The important thing to check is that the two files generted in the the `tmp` directory actually _contain the new OMIM terms.. When I tried this with the new version of OMIM you created that was not the case - and the reasons for why this may not be the case can be quite complex in the the slurp pipeline. Can you double check these files actually contain the new OMIM slurped terms (for now, just check they are not empty)?

joeflack4:
Good bringing this up. The output files are in fact empty. I've updated the corresponding issues, and we have some discussion on this on Slack with some examples, so let's pick it up there for now.

PR #3773, opened 2021-10-07
Create disease_has_major_feature.yaml

laurenechan:
I wasn't sure if I should include any of the definition, synonym, or naming fields as they probably aren't relevant for many of the terms that can use this pattern. Hopefully we can discuss and iterate @nicolevasilevsky


matentzn:
Converted to draft because I dont see any action items.

nicolevasilevsky (approved):
@laurenechan I finally got it to pass! I looked at some other examples, it was very finicky.

PR #3786, opened 2021-10-12
Adding explain debug.owl

matentzn:
Adding check to explain debug.owl and having pipeline fail right afterwards.


matentzn:
@nicolevasilevsky bring up at QC call in conjunction with the complicated  chromosomal diseases issues

PR #4096, opened 2021-12-03
Remove ENVO from ECTO

matentzn:
After QC passes you should also double check that this didn't leave stray classes in the mondo class graph in protege.. Like things that floated to the top, especially ECTO terms.


nicolevasilevsky:
looks good!
image

matentzn:
gr8

nicolevasilevsky (approved):

PR #4341, opened 2022-01-21
Rename disease->disorder in labels for terms that conform to patterns/location.yaml

joeflack4:
Close https://github.com/monarch-initiative/mondo/issues/508

Updates
```
    - Ran: fix-disorders-conformsTo-location.ru and updated mondo-edit.obo

    - Added SPARQL query: select-conformsTo-location.sparql
    - Added SPARQL update query: Conforms to pattern/location.yaml and has word ‘disease’ in table. Bind label with regex replace pattern from disease to disorder. Delete old label. Insert new label
    - Updated: select-conformsTo-allergy.sparql: added a prefix
```


joeflack4:
@nicolevasilevsky @matentzn I think Nico said this looked like it might be good?

But for me it looks like there are 2 issues, and I don't know what to do:
1. I don't see that "disorder". Is that an issue? Maybe I don't see it because most of the terms were already "disorder", and most of the diffs I'm seeing are simply selections of terms that conform to the `location.yaml` pattern? In that case, I would imagine that no change should have been made, but...:
2. Everything after the `!` following the label has been deleted. This doesn't seem good to me. What do I need to do to fix this? I don't know how to query for this `!` and the text that follows it, and preserve that.

[diff.txt](https://github.com/monarch-initiative/mondo/files/7915786/diff.txt)
Screen Shot 2022-01-21 at 1 13 32 PM

nicolevasilevsky:
Hi @joeflack4 
Unfortunately, I don't think this worked. I don't see that  MONDO:0000001 disease or disorder was changed to disorder.

The changes you are seeing in the screenshot above are updates to the imports - to hgnc and so. Sometimes I see this where the labels are deletd, I don't know it does this.

joeflack4:
Shucks... Well I'm at a loss as to why these 2 issues are happening.

The query: [src/sparql/update/fix-disorders-conformsTo-location.ru](https://github.com/monarch-initiative/mondo/pull/4341/files#diff-06072dc17155d077799c526249dbae593b657935143e6d6ea09540aa06ae3ed0)
```sparql
prefix dcterms: <http://purl.org/dc/terms/>
prefix owl: <http://www.w3.org/2002/07/owl#>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix xsd: <http://www.w3.org/2001/XMLSchema#>
prefix mondoPatterns: <http://purl.obolibrary.org/obo/mondo/patterns/>

DELETE {
  ?term rdfs:label ?label .
}
INSERT {
  ?term rdfs:label ?new_label .
}
WHERE {
  ?term rdfs:label ?label .
  ?term dcterms:conformsTo mondoPatterns:location.yaml .
  FILTER NOT EXISTS { ?term owl:deprecated "true"^^xsd:boolean . }
  FILTER (isIRI(?term) && regex(str(?term), "^http://purl.obolibrary.org/obo/MONDO_"))
  FILTER(regex(str(?label),"disease"))
  BIND(REPLACE(str(?label), "disease", "disorder") as ?new_label )
}
```

matentzn:
All changes to mondo-edit.obo where spurious - they happened because you used the wrong command for applying the update.. 

This command applies one of your 2 queries:
sh run.sh make update-merge-normalise-fix-disorders-conformsTo-location

As @nicolevasilevsky has said, it does not do anything at the moment. Most likely, the where clause does not match anything. This needs a bit of playing around: I would suggest you make sure the select query that contains the identical WHERE clause works properly first.

joeflack4:
@matentzn Thanks; I am sorry this is not working as expected.

The select statement seems like it should work:
https://api.triplydb.com/s/bSFVrEYKb

The only difference in the `WHERE` clause between the Yasgui one and [the one I'm making for Mondo](https://github.com/monarch-initiative/mondo/pull/4341/commits/0920e807dbfd227740dc1a520fdf661476cc005c#diff-06072dc17155d077799c526249dbae593b657935143e6d6ea09540aa06ae3ed0) are the following lines that were added:
```sparql
  FILTER (isIRI(?term) && regex(str(?term), "^http://purl.obolibrary.org/obo/MONDO_"))
  FILTER(regex(str(?label),"disease"))
  BIND(REPLACE(str(?label), "disease", "disorder") as ?new_label )
```

It doesn't seem like anything complicated is going on here, though... Based on this above, do you have any idea what's wrong?

---

**ODK docker**
Also, I think I need to learn more about the merits of using the ODK docker container via `sh run.sh`. Relying on using a docker container _as an environment_ (rather than as an application/service), as it would appear, is new to me. It makes me feel a bit suspect of it, but perhaps that I need to open my mind a bit more. I'm assuming that the advantage is that it bundles a bunch of different tools that have non-easy installations, and includes tools from different programming languages. If they were all written in the same programming language, then I imagine it would be better to use a package management system (e.g. Maven, PyPi, NPM) to ensure consistent versions. But maybe there are data artefacts in the environment too. Maybe this is something that we can bring up at our Wednesday meeting if we have time.

---

I think I'll probably be pushing this ticket to next week, as #4346 and the mondo paper will be filling up my Mondo time for this week.

matentzn:
This is PR is giving me a bit of trouble. I am trying to wrap my head around it: https://github.com/ontodev/robot/issues/967

Looks like your work is all correct @joeflack4 

Will keep looking into this. The workaround is to use sparql construct together with robot unmerge/merge.

matentzn (approved):

nicolevasilevsky:
Nicole to do:
- [ ] check out branch, open in Protege and make sure everything looks okay and check diff too

nicolevasilevsky (approved):
this is ready to merge when QC passes

PR #4386, opened 2022-01-28
Pattern merge pipeline trial run - poisoning.tsv

laurenechan:
Trying out new pattern merge pipeline from #4385  with the poisoning.tsv pattern file


matentzn:
Wow, I cant believe its worked. Its not so much me that should review it - the question is - does this match your expectations (ignore the small changes to the synonyms, this is because of the normalisation process).

laurenechan:
I think it did work how I anticipated, but I'm seeing that checks have failed. Can you help me understand what the problem is?

matentzn:
I could, but I leave it to @sabrinatoro or @nicolevasilevsky to walk you through that. This has an easy fix, but needs some medical expertise and explanation. So, don't worry about it now! They will look at it on Monday!

nicolevasilevsky:
@laurenechan I'll take a look at this asap but it may not be until later in the week

nicolevasilevsky:
@laurenechan this is the reason for the failure:
`FAIL Rule ../sparql/qc/general/qc-single-child.sparql: 1 violation(s)`
`entity,property,value`
`http://purl.obolibrary.org/obo/MONDO_0024251,http://www.w3.org/2000/01/rdf-``schema#subClassOf,http://purl.obolibrary.org/obo/MONDO_0018020`

the issue is a single child, we have a check for this. You can read more about it in Chris' blog post: https://douroucouli.wordpress.com/2022/01/04/ontotip-avoid-the-single-child-anti-pattern/

I'll look into this and fix it

nicolevasilevsky:
these are the classes in question

image

nicolevasilevsky:
'Minamata disease'  is defined as A neurological syndrome caused by severe mercury poisoning.

It is caused by mercury poisoning, but it is not necessarily a type of mercury poisoning. [Wikipedia](https://en.wikipedia.org/wiki/Minamata_disease) says Minamata disease is a [neurological](https://en.wikipedia.org/wiki/Neurological) disease caused by severe [mercury poisoning](https://en.wikipedia.org/wiki/Mercury_poisoning). 

I'll revise the parent class for this term.

nicolevasilevsky (approved):

nicolevasilevsky:
need to fix conflict and then this can be merged

PR #4591, opened 2022-03-09
query for number of rare diseases in Mondo

sabrinatoro:
@Nico : what this query does is explained in the comment in the query.

And for convenience:
MONDO classes with the axiom "'has modifier' some rare" AND/OR "has subset :‘gard_rare’" AND limited to the 'disease' branch (ie not including terms from the 'disease susceptibility' branch)


sabrinatoro:
@matentzn 
Actually, it looks like what it did is not correct: according to [here](https://mondo.monarchinitiative.org/pages/analysis/), we need to "consider only leaf nodes".
If I understand well, it means that I have to filter out the results which would be terms that have children (ie filter out terms which are parents). I don't know how to do that... help!!!!!! please!!! :-)

matentzn:
Excellent job finding the `^` operator! This looks really nice. :) The queries look great to me!

nicolevasilevsky (approved):

PR #4673, opened 2022-03-24
added 'tags' for Mondo Clinical view.

sabrinatoro:
Addresses #4113
@matentzn can you please review this.
More specifically, can you confirm that this is the correct way to exclude the "non-human animal disease" (MONDO:0005583) class (+children) from the clinical view? Will this exclude the whole class and children? or will it make this whole branch "orphan" (and therefore create reasoning issues)? Thanks!


matentzn (commented):
Hmmm, don't like how this is looks. This is what I said, but i think it's better to use source="MONDO:excludedFromClinicalView".. sorry :/

matentzn:
And no I didn't think of excluding the hole branch this way, just the subclass axiom. If you want to exclude the whole branch, you should create an "antislim", meaning you create a new subset, like "exclude_from_clinical" and annotate the term. Maybe a second subset as well "exclude_branch_from_clinical" which means to exclude a class and all its children.

sabrinatoro:
@matentzn about using source="MONDO:excludedFromClinicalView"
1) would this still be on the SubClassOf annotation? 
2) what about ignoring the non-human disease branch? (see comment above, assuming the source=Mondo:excluded is on the SubClassOf)
we can also discuss this in our meeting.

matentzn:
1. yes
2. see above, use subsets for that

:)

PR #4735, opened 2022-04-07
Fixing some references to OMIM and Orphanet

matentzn:
I dont think this is questionable, but some information is lost (some specific URL parameters). I don't think this should be a problem, hopefully.


nicolevasilevsky (approved):
these changes seem okay to me, do you want to merge this?

matentzn:
Yes, thanks!

PR #4750, opened 2022-04-11
update ICD10CM to ICD10WHO

### LINKED ISSUE #4651: Ensure that ICD10 mappings are ICD10CM, not ICD10WHO.
We so far made the flawed assumption that and an id with the ICD10 prefix is in fact ICD10CM if only the id is a legal ICD10 CM code. However, there is no guarantee this is enough. According to @drseb there could be differences between the exact same code in ICD10WHO and ICD10CM. So we need to make sure that all our codes are not only syntactically legal, but also correct. I do not know how to solve this yet, so it needs some thought. Hopefully this problem is not that huge.

sabrinatoro:
fixes #4651
@matentzn I updated the ICD10CM prefixes to ICD10WHO for the x-ref in which ICD10WHO was more precise.

BUT:
I think that the ICD10CM code were not "wrong", just less precise than the ICD10WHO. When a ICD10CM code exist, shouldn't we keep it and add the ICD10WHO as an additional x-ref?


sabrinatoro:
@matentzn ICD10WHO is not an allowed prefix?

matentzn:
@sabrinatoro you will have to add ICD10WHO to the list of allowed prefixes :D Let's see if you can figure this out by yourself, but if you cant, I will do it for you.. Just for fun!

I trust your judgement on the xrefs: the correct way would be to add the ICD10WHO ones and make them EXACT, and the other ones should probably be narrower or broader..

PR #4757, opened 2022-04-11
added dosql query

hrshdhgd:
- [x] Added a `dosql` query from `sssom` to filter MappingSetDataFrames.
- [x] Sorted columns of DataFrame.
- [x] Sorted rows of DataFrame by column # 1
- [x] Added flags to sort by columns and/or rows


src/ontology/mondo.Makefile
```diff
@@ -441,7 +441,7 @@ tmp/%.sssom.tsv: tmp/mirror-%.json | sssom
 
 $(MAPPINGSDIR)/%.sssom.tsv: tmp/%.sssom.tsv
 	python ../scripts/split_sssom_by_source.py -s $< -m $(METADATADIR)/mondo.sssom.config.yml -o $(MAPPINGSDIR)/
-	cp $< $@
+	sssom dosql -q "SELECT * FROM df WHERE predicate_id IN (\"skos:exactMatch\", \"skos:broadMatch\")" $< -o $@
```
matentzn, line 444: This looks great. Can we:
- Drop skos:broadMatch from that query (sorry my bad)
- call a `sssom sort` on all mappings (with the changes I suggest to CLI, in particular row sorting)

hrshdhgd: Already did bullet 2. Re-running now after implementing bullet 1.

PR #4766, opened 2022-04-12
mondo.sssom.config updates

joeflack4:
mondo.sssom.config updates
- OMIMPS: Deleted a deuplicate mapping. The one that I left behind assumes that the ID to be added onto the prefix will be in integer form, e.g. '100300' rather than 'PS100300'.
- HGNC: Added
- HGNC__2: Added this alternate url.
- HGNC__symbol: Added. Even though the HGNC ID is the canonical identifier, not the symbol, I figured that people may want to review this addition; maybe it's useful. I'm not sure if HGNC__symbol or HGNC_symbol is a better name, but I stuck with __, as alternate IDs appeared to use this syntax in this file.

### DIFF
--- a/src/ontology/metadata/mondo.sssom.config.yml
+++ b/src/ontology/metadata/mondo.sssom.config.yml
@@ -4,6 +4,9 @@ curie_map:
   NCIT: http://purl.obolibrary.org/obo/NCIT_
   DOID: http://purl.obolibrary.org/obo/DOID_
   EFO: http://www.ebi.ac.uk/efo/EFO_
+  HGNC: "https://identifiers.org/hgnc:"
+  HGNC__2: "https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:"
+  HGNC_symbol: https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/
   HP: http://purl.obolibrary.org/obo/HP_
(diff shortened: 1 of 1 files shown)


joeflack4:
@matentzn asked me to add HGNC mappings. I also fixed a duplicate OMIMPS entry.

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -4,6 +4,9 @@ curie_map:
   NCIT: http://purl.obolibrary.org/obo/NCIT_
   DOID: http://purl.obolibrary.org/obo/DOID_
   EFO: http://www.ebi.ac.uk/efo/EFO_
+  HGNC: "https://identifiers.org/hgnc:"
+  HGNC__2: "https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:"
+  HGNC__symbol: https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/
```
matentzn, line 9: Use single underscore here as this may confuse the hacky code. Symbol and ID are different namespaces. I used the double underscore to mean same namespace different iri expansion..

joeflack4: Makes perfect sense. I was wondering if this might be the case. Will make the change and push momentarily.

matentzn (approved):

Issue #4675, opened 2022-03-25
disease to gene relation: distinguishing 'inherited' versus 'acquired' 


sabrinatoro:
**Currently**: 
The disease to gene relation used 
'disease has basis in dysfunction of' (RO:0004020)
Definition : A relation that holds between the disease and a material entity where the physical basis of the disease is a disorder of that material entity that affects its function.

This relation is also associated with the GCA:
'disease has basis in dysfunction of' some gene SubClassOf 'has modifier' some inherited

**Issue**: 
When it is true in most cases, especially for information from omim, the relation between a disease and a gene is not always inherited (e.g. somatic mutations, see also #4232)

**To do**: 

- [x] create new Annotation Properties (or relation) to specify whether the dysfunction of the gene is inherited or acquired.  (names TBD)
- [x] update Mondo GCA to specify that the "inherited" relation to gene has the inference 'SubClassOf 'has modifier' some inherited'
- [x] review current QC (eg qc-omimps-should-be-inherited.sparql requires "has modifier some inherited" when a term has a OMIMPS: x-ref  MONDO:equivalentTo.
- [x] review existing patterns
- [x] add the 'acquired' relation to MONDO:0010438 and MONDO:0014166

sabrinatoro:
from curation call 04/14/2022
- use the relation to Mondo: http://purl.obolibrary.org/obo/RO_0004004 ('has material basis in somatic mutation in') 
- use 'has material basis in germline mutation in' instead of 'disease has basis in dysfunction of'

PR #4800, opened 2022-04-18
acquired/inherited relation with gene

sabrinatoro:
addresses #4675


sabrinatoro:
@matentzn : this PR is failing because : 
FAIL Rule ../sparql/qc/mondo/qc-omimps-should-be-inherited.sparql: 1 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0[10](https://github.com/monarch-initiative/mondo/runs/6100911796?check_suite_focus=true#step:4:10)0244,http://purl.obolibrary.org/obo/RO_0002573,

Background: some omim/omimPS terms could be acquired. Therefore I updated the patterns and the QC to reflect that the relation to a gene could be "has material basis in somatic mutation in" or "has material basis in germline mutation in". I also updated the GCA to indicate that  "has material basis in germline mutation in" some gene (instead of "disease has basis in dysfunction some gene") should also have the inherited modifier. 
 
In this case the QC is specifically to check that a term with omimps has to be "inherited", which is not correct anymore. 
I am not sure how to proceed. Should we 1) remove this QC all together, or 2) update the QC such that it says that IF the omimPS has children with omim "inherited", then it should be inherited? 
Thank you!

matentzn:
Please raise in QC call or next 1:1

sabrinatoro:
@matentzn 
This PR is still failing due to the "FAIL Rule ../sparql/qc/mondo/qc-omimps-should-be-inherited.sparql"
I added the "exclude from" this QC, and we updated the sparql to allow this "exclude from" in another PR (and I checked the sparql query and cannot find anything wrong).
I cannot figure out why I keep on getting the same QC fail. 
Should I do something to make sure that the updated QC is taken into consideration? 
Thank you!

sabrinatoro:
UPDATE: I think there might be another (sneaky) typo in the sparql

matentzn:
too many changes in the diff.. risky. Update from master and make sure you add the sparql query correctly. Share protege screen shot of exclusion annotation?

nicolevasilevsky (approved):

PR #4886, opened 2022-04-29
Create qc-doublewhite.sparql

matentzn:
This check prevents double whitespaces and tabs to make it into any of the Mondo annotations


matentzn:
Thank you @drseb for reporting!

nicolevasilevsky (approved):

PR #4932, opened 2022-05-10
Upgrade HGNC URIs

cthoyt:
This PR upgrades the URI format for HGNC references to use identifiers.org's HTTPS protocol as well as the CURIE-style syntax, rather than the legacy path syntax (which uses a slash)

Specifically, this changes all URIs of the form `http://identifiers.org/hgnc/<HGNC ID>` to `https://identifiers.org/hgnc:<HGNC ID>`.

Alternatively, it might be better to just use CURIE syntax and abstract away the different representations of HGNC terms with URIs anyway (since there are currently 4 different ones imposed by identifiers.org)


matentzn:
@cthoyt proposes this for all HGNC references:

image

This must be coordinated with Monarch as a whole. At least, we need input from @cmungall @kevinschaper on this change.

PS: we cant use "curie syntax" for HGNC as these unfold to OBO purls in their OWL serialisation.

matentzn:
@cthoyt proposes this for all HGNC references:

image

This must be coordinated with Monarch as a whole. At least, we need input from @cmungall @kevinschaper on this change.

PS: we cant use "curie syntax" for HGNC as these unfold to OBO purls in their OWL serialisation.

kevinschaper:
Can you tell me more about what this might affect downstream in Monarch? For the new pipeline, my guess is that this might affect the conversion of the ontology to kgx format? by the time I merge it into the graph, I think the example above would just be an edge with the hgnc id in curie format?  We might also need to address the url mapping that happens in the UI?

kevinschaper:
I checked what I'm getting right now with a kgx conversion of the Jenkins ontology build, and it's none of the above! 

```
id	subject	predicate	object	category	relation	provided_by
urn:uuid:3a8410fa-bba7-4f9d-a51f-419446ba5063	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	biolink:category	biolink:NucleicAcidEntity		biolink:category
urn:uuid:61ae0fc8-571a-4ead-b7c3-8bc66c238e28	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	biolink:category	biolink:OntologyClass		biolink:category
urn:uuid:ff4321ff-2ed2-4da1-96a4-7fbde463db97	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	biolink:category	biolink:GenomicEntity		biolink:category
urn:uuid:3392daf5-2236-482d-8c25-f9b34b0e1e76	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	biolink:subclass_of	SO:0000704	rdfs:subClassOf
urn:uuid:059fca17-b065-435d-b925-280e9a2f83f7	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	biolink:category	biolink:Gene	biolink:category
urn:uuid:9fd90e02-7333-4527-af35-5f829fc9d47c	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	biolink:category	biolink:NamedThing		biolink:category
urn:uuid:f2f97cc6-6587-460a-873e-ef087ade33b1	MONDO:0100267	biolink:related_to	https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:8855	RO:0004020
```

I think that the only edge I actually want is the last one (the rest should just be node properties), but the gene node I have is just `HGNC:8855` - so that's what I'm hoping to see as the subject in the last row. I'm not sure what that means in terms of what belongs in an ontology vs what the conversion should do though. 

@caufieldjh @sierra-moxon you might have a better understanding of what the conversion layer needs / should do?

caufieldjh:
KGX looks up from prefixcommons, and the [HGNC prefixes](https://github.com/prefixcommons/prefixes/blob/a3b86fe0ccacec94ddbf60226e65f3b9515653d1/registry.yaml#L1640) don't use https, so KGX doesn't convert.

matentzn:
TBH I think we should be using `https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:` as the canonical expansion (https://bioregistry.io/registry/hgnc). The problem is probably that OBO format does not like the `#!` in it. So we instead use identifiers.org. The question is what we should do across all our projects - live with the fact that we use different expansions everywhere and somehow remember that we need to harmonise them (like we would for any "external source")?

kevinschaper:
Would it be wildly inappropriate to see if HGNC would be willing to add (& maintain) a shorter url?

matentzn:
I dont think this is realistic.. We need to discuss this with Chris, maybe he has an idea.

cmungall:
For better or worse, the decision we came up with 4 years ago in biolink was to use http identifiers.org URIs, with a slash

https://biolink.github.io/biolink-model/#identifiers

In monarch we conform to biolink, so I think we can close this and take any discussion to the biolink tracker

Remember: we are not just selecting URLs for humans to follow a link to web pages. These are identifiers, and thus must have the property of stability. We cannot go changing them without breaking things at the RDF level. It is not enough to have http redirects for URLs/

We absolutely should not use `https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC` and this has nothing to do with obo format. This URL is clearly not intended to be used as an RDF identifier. The chances of it changing in the future will be high. And then there will be pressure to change the URIs in the OWL and we will be having this conversation again.

The reason we chose identifiers.org with http and slash not colon:

- these URIs were commonly used in many triplestores
- https was not rolled out for identifiers.org
- formally there are issues with using colons in the local name in some W3C specs such as SPARQL (e.g. https://www.w3.org/TR/rdf-sparql-query/#rPN_LOCAL). In practice tools don't seem to follow the spec and are lenient, but we felt it was still too risky
- although identifiers.org carrying no strong typing (aka "ontological commitment") could be seen as a disadvantage it can also be an advantage - see the Schulz conflation pattern

I thought this was recorded somewhere, I will look this out

I am open to changing this, but this should be done globally as part of of a coordinated strategy to give stable semantic web URIs to databases entities with appropriate levels of typing.

As an aside, I think the idea of using URLs as identifiers is really problematic and one of the biggest downsides of the semantic web. It's a clear violation of separation of concerns. I seem to be in the minority here. We had a lot of discussion about this on the #https-for-purls channel on OBO slack but we need to pay to unlock that conversation now....

cmungall:
of course, the CURIEs should look like proper CURIEs in obo format, if that isn't happening let's open a separate issue

joeflack4:
Hmm... I googled "Schulz conflation pattern" and skimmed through a few articles, but I couldn't find a definition for it anywhere.

matentzn:
Schulz wouldnt call it conflation: https://jbiomedsem.biomedcentral.com/articles/10.1186/2041-1480-2-S2-S6

PR #5038, opened 2022-06-13
Use CURIE syntax for HGNC references

cthoyt:
As an alternative to #4932, this PR compresses all URI-based references to HGNC terms into CURIEs and adds an idspace declaration at the top of the document (see line 84) to expand it back to the same URIs as before on conversion.


cthoyt:
@matentzn so the issue is I should remove the `idspace` line? Or just the fact that we can't handle any CURIEs that aren't assumed to correspond to OBO PURLs?

If the latter is the case, can we register HGNC with the OBO PURL system? I already have demonstrated generating a corresponding OWL file for HGNC in https://github.com/pyobo/examples/tree/main/export/hgnc

cmungall:
It’s not ignored by pronto/fastobo!

So I think the solution here is not to edit the edit file. Let’s focus on
the release obo file and ensure it the urls are contracted here.

matentzn:
You are sure this is wise? Does this not make the OBO files vulnerable to processing in OWL API based pipelines (I know we shouldn't use owl api for processing OBO, but matter of fact is we do!) I am like 55 % against this because of the marginal gain offered to obo file users on python compared to the potential problems for ROBOt users..

cthoyt:
@matentzn I think philosophically the reason it's important to use CURIEs in the edit file is it allows all consumers to make their own choices of how to expand them, assuming everyone agrees on the prefix, which in the case of HGNC is pretty noncontroversial.

@cmungall I'm not sure I understand what you want me to do, though

matentzn:
@cthoyt you are scratching a pretty important conceptual issue here though. The whole semantic web side of things explicitly does not want you to "make their own choices to expand them". Given the importance of the whole semweb stack, we absolutely should be prescriptive on the URI expansions for any entity used for logical modelling in an ontology. I know this is a bit confusing if you are not a member of the cabal, but there is no such thing as a "URI expansion" in semantic web (RDF/OWL) land. There are prefix declarations, and these are literally to "shorten" the IRI to something more compact, not to say "if you see x: its intended to be expanded to http://..". The distinction is very important: we cant have Mondo expanding hgnc to idorg-slash curies and HPO to have them expand to idord-colon curies: when integrating these ontologies together, our tech stack will think that they are entirely distinct. 

Here the only thing that is in question is this:

1. `idspace` declarations in OBO are according to OBO spec. (+ 3 points for argument for turning HGNC into CURIEs)
2. `idspace` declarations are ignored by the most important tool in OBO for processing ontologies (-7 points for argument for turning HGNC into CURIEs)
3. OBO tools like pronto benefit from the CURIE representation a lot. ROBOT and other semantic web leaning tools should not consume OBO files as inputs and turn them into owl (OWL always exists, +4 points for argument)

Now the question is: are ROBOT pipelines in fact consuming OBO files, which would cause real issues? The answer to the question is what matters here. I know a bunch of pipelines that turn a file to OBO. Run obo-grepl.pl commands. Turn back to OWL. This would cause `idspace` to be ignored and indeed, to be rewired to some terrible OBO purl.

I am against doing this.

If someone thinks this is a priority, I think we have to:
1. Fix OWLAPI obo parser
2. Wait for 2024 for another OWLAPI upgrade

matentzn:
After some consideration we decided not to merge this PR - for the reason outlined above. 

I think we also should have a sync call again soon - one thing I think that stands at least a bit between us is your notion of "identifier" - for me, due to the tech stack I work with (semweb) a curie string is not an identifier - the combination of a prefix declaration with a curie is a compact notation for a URI. So the question is not - "how does the user want to expand the CURIE?". This question makes no sense in the semantic web. An entity is referred to by exactly one URI. The fact that this is empirically not true does not change this conceptual fact. Hence my insistence to standardise "uri expansions" in bioregistry and much less care for standard prefixes. 

Since technical issues will basically cause the HGNC prefix to be expanded to some utter nonsense (obo purl) if we merge this PR (rather than the declared one), I deem the risk too high. 

mondo.obo does not actually have any HGNC in it - how did you even figure out these were non standard? 

Let me know if this makes sense. I am happy to discuss this in a call - we are due a sync up anyways. 😀 

Thanks as always @cthoyt for encouraging standardisation! I hope this does not discourage you to continue to propose other changes! We are keen to standardise xrefs and prefix usage on dc:source annotations as well, and I think we have come a long way now, but still lots to do!

nicolevasilevsky:
@matentzn can we close this?
Yes, thanks @cthoyt!

matentzn:
> Are there reasons we can't go forwards and add an HGNC entry to purl.obolibrary.org and standardize on that? It's been done for NCIT and NCBITaxon already

This is an option.. I guess @cmungall does not see a clear path on how this would fit into the wider strategy - should all databases without standardised uri expansions get some OBO namespace?

I think its better if we focus on standardising a URI scheme for now rather than getting CURIEs into OBO files. Ontologies do not contain CURIES.. This is some ideosyncracy of the OBO format. Ontologies contain IRIs which can be contracted to CURIEs for readability..

matentzn:
Lets move the discussion to a more visible place: https://github.com/OBOFoundry/OBOFoundry.github.io/discussions/1981

PR #5069, opened 2022-06-17
Fix WHO CM bug

matentzn:
Fixes https://github.com/mapping-commons/sssom-py/issues/260


matentzn:
@nicolevasilevsky this can be merged

PR #5279, opened 2022-08-16
GH actions: skip diff workflow on forks

dhimmel:
closes https://github.com/monarch-initiative/mondo/issues/5255


dhimmel:
Created https://github.com/dhimmel/dump-actions-context to explore the scheduled workflow context

Issue #5270, opened 2022-08-12
Update enum file for exclusion reasons


sabrinatoro:
The enum file (https://github.com/monarch-initiative/mondo/blob/master/src/schema/mondo.yaml) should be updated to add the updated exclusion reasons. 
The exclusion reasons can be found [here](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180) and [here](https://mondo.readthedocs.io/en/latest/editors-guide/exclusion-reasons/)

joeflack4:
I'm on it!

joeflack4:
@sabrinatoro Just wanted to let you know I made some minor edits to the [exclusion reasons tab of that google sheet](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180).
1. Simplified header names
2. Added a comment to the `exclude_children` header
3. Collapsed A-D into a single column (I looked, and there was no reason for these cells to be merged anyway)

I added a version name to the file before and after my changes, just in case you want to revert.

joeflack4:
@sabrinatoro I did a check to see if there were any exclusion reasons listed in the [Mondo LinkML schema enum](https://github.com/monarch-initiative/mondo/blob/master/src/schema/mondo.yaml) but not located in the [GoogleSheet](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180), and I found these:

```
excludeClinicalModifier
excludeConditionsCausedByExternalForce
excludeCausingAgents
excludePresenceOfEnvironmentalMaterials
excludeGeneticCarrier
excludeGeneticFormOfDisease
excludeGroupingClass
excludePhenotypeForMetastasis
excludePhenotypeForTumor
excludeRare
excludeConditionsCausedByExternalFactor
excludeBehavior
excludeTraumaticCause
```

I just want to check; are these missing from the GoogleSheet because they don't need any updates, or do they need to be added to it?

I'm writing a script to update from the GoogleSheet, so future updates to the schema enum should be fairly painless.

sabrinatoro:
@joeflack4 these missing ones are "old" codes that should be deleted (and not used anymore).

joeflack4:
Oh OK, I will delete them then.

PR #5288, opened 2022-08-18
PR: Update enum file for exclusion reasons

joeflack4:
Updates
```
    Feature: Exclusion reasons enum updater
    - Update: mondo.yml: Updated via script; includes new is_a field
    - Update: Script: (i) Now doesn't add any fields where there is no value in cell, (ii) Removed exclude_children field.
```

```
    Feature: Exclusion reasons enum updater
    - Update: Removed warning, and replaced it with operation that will delete any items in the .yml that do not appear in the GoogleSheet source of truth.
```

```
    Feature: Exclusion reasons enum updater
    - Add: Script for this work
    - Add: Dependencies to Python requirements files
    - Update: .gitignore: Exclude CSV file downloaded from GoogleSheet
    - Update: mondo.Makefile: Added PHONY target for this work. Also added PHONY target to update python dependencies.

    Misc
    - Add: Python requirements files
    - Update: .gitignore: Reorganized roughly according to type and alphabetically
    - Update: src/schema/mondo.yaml per using new feature.
```


joeflack4:
I figured there was enough to update that this merited some scripting, rather than doing manually. Will help in the future, too.

requirements-unlocked.txt
```diff
@@ -0,0 +1,2 @@
+pandas
```
joeflack4, line 1: @matentzn I'm not opposed to `poetry` or `tox`, but this native `pip` way of doing things is the fastest / easiest way for me to do this right now.

I hope it's OK to add these dependencies, even though I have only one script that would use them right now.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,10 @@
+"""Updates exclusion reasons enum: LINK_TO_FILE"""
```
joeflack4, line 1: FYI: The script

.gitignore
```diff
@@ -1,34 +1,38 @@
+# for mkdocs' Python virtualenv
```
joeflack4, line 1: FYI: Reorganized `.gitignore`

.gitignore
```diff
@@ -82,14 +87,14 @@ src/ontology/sources/*/*.obo
...
-src/dosdp-tools-*
-src/ontology/.cogs
-src/ontology/imports/omo_terms.txt
-.vscode
-src/ontology/mondo-ncit.owl
-src/ontology/tbd/*
-mondo-tags-reasoned.owl
+src/patterns/imports/seed_sorted.txt
+
+src/scripts/exclusion_reasons.csv
```
joeflack4, line 100: FYI: Added to `.gitignore`. Taken from [Google Sheet](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180).

src/ontology/mondo.Makefile
```diff
@@ -700,4 +700,13 @@ dosdp-merge-%: ../patterns/data/default/%.owl tmp/remove_%.ru
 		make NORM
 		mv NORM $(SRC)
 
-p1: dosdp-merge-acute
\ No newline at end of file
+p1: dosdp-merge-acute
+
+python-install-dependencies:
```
joeflack4, line 705: FYI: New `PHONY` targets for this work.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,61 @@
...
+from pathlib import Path
+from typing import Dict
+
+import pandas as pd
+import yaml
+
+
+GSHEETS_URL = 'https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180'
+PROJECT_DIR = Path(os.path.dirname(__file__)).parent.parent
+DEFAULTS = {
```
joeflack4, line 20: I know defaults discouraged, but this is not `mondo-ingest`; interested more in development speed than convention at this moment.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,61 @@
+"""Updates exclusion reasons enum: src/schema/mondo.yaml
```
joeflack4, line 1: The script.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,77 @@
...
+	# - Format to remove whitespace, leading indentation markers, and unneeded MONDO prefix
+	df['codes'] = df['codes'].apply(lambda x: re.sub('.*MONDO:', '', x))
+
+	# Load current exclusions
+	with open(input_path_mondo_schema, 'r') as stream:
+		mondo_schema = yaml.safe_load(stream)
+	exclusion_dicts: Dict[str, Dict[str, str]] = mondo_schema['enums']['entity_type_enum']['permissible_values']
+
+	# Warning: schema enum items not located in GoogleSheet
+	missing_exclusion_reasons = [x for x in exclusion_dicts if x not in set(df['codes'])]
```
joeflack4, line 42: FYI: Prints a warning if some exclusion reasons not found in GoogleSheet. I did find some. Added comment to: #5270

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,98 @@
+"""Updates exclusion reasons enum: src/schema/mondo.yaml
+
+Resources:
+- Source CSV: https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180
+
+Instructions:
```
joeflack4, line 6: FYI: Added instructions to the top of the file.

src/scripts/exclusion_reasons_enum_updater.py
joeflack4: Update: [Per Sabrina](https://github.com/monarch-initiative/mondo/issues/5270#issuecomment-1220087428), I have removed this warning, and it now instead deletes these:
> these missing ones are "old" codes that should be deleted (and not used anymore). Thank you for checking! I should have communicated that.

src/schema/mondo.yaml
```diff
@@ -1,99 +1,360 @@
 id: https://w3id.org/mondo
 name: mondo
-description: >-
```
joeflack4, line 3: @matentzn @sabrinatoro Updated `src/schema/mondo.yaml`. Please take a look.

I was able to prevent it from re-ordering things, but it still does some code style reformatting, sorry about that. Hope this is ok.

I scripted the edits starting with a map between the fields in the [GoogleSheet](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180) and the `.yml:
```py
SHEET_YML_MAPPINGS = {
	'code_name': 'title',
	'definitions': 'description',
	'notes': 'notes',
	'examples': 'examples',
	'exclude_children': 'exclude_children',
}
```

@sabrinatoro If you want, we can streamline this and make the column names in the GoogleSheet the same as in the `.yml`. But I don't mind doing it this way.

If these mappings are incorrect, though, or if additional fields need to be added, please let me know. The script will ignore any values that these fields previously had in the `.yml`, and will completely overwrite with what is in the GoogleSheet, since it is the source of truth.

@sabrinatoro If you add any new columns to the GoogleSheet, it will add the exact spelling of the column name and use that as the field in the `.yml`.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,95 @@
...
+		'-e', '--input-path-exclusion-reasons', required=False, default=DEFAULTS['input_path_exclusion_reasons'],
+		help='Path to source exclusion reasons CSV to update from.')
+	parser.add_argument(
+		'-s', '--input-path-mondo-schema', required=False, default=DEFAULTS['input_path_mondo_schema'],
+		help='Path to Mondo LinkML schema `.yml` which contains exclusion reasons enum to be updated')
+	kwargs = parser.parse_args()
+	d: Dict = vars(kwargs)
+
+	# Validate
+	if not os.path.exists(d['input_path_exclusion_reasons']):
```
joeflack4, line 87: FYI: If the input CSV is missing, it says where to get it, and what path to put it in.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,95 @@
...
+		del exclusion_dicts[item]
+
+	# Updates
+	for _index, row in df.iterrows():
+		exclusion_reason_name = row['codes']
+		row_d = {k: v for k, v in row.to_dict().items() if k != 'codes'}
+		if exclusion_reason_name not in exclusion_dicts:
+			exclusion_dicts[exclusion_reason_name] = {}
+		for k, v in row_d.items():
+			target_field = SHEET_YML_MAPPINGS.get(k, k)
```
joeflack4, line 62: FYI: `.get(k, k) will add a new column from the GoogleSheet if it does not exist in the `SHEET_YML_MAPPINGS`.

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,95 @@
...
+	# - Strip all whitespace
+	df = df.apply(lambda x: x.str.strip() if x.dtype == 'object' else x)
+
+	# Load current exclusions
+	with open(input_path_mondo_schema, 'r') as stream:
+		mondo_schema = yaml.safe_load(stream)
+	exclusion_dicts: Dict[str, Dict[str, str]] = mondo_schema['enums']['entity_type_enum']['permissible_values']
+
+	# Delete: Any schema enum items not located in GoogleSheet source of truth
+	obsoleted_exclusion_reasons = [x for x in exclusion_dicts if x not in set(df['codes'])]
```
joeflack4, line 51: FYI: Deletes any old exclusion reasons not found in the [GoogleSheet](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180).

src/scripts/exclusion_reasons_enum_updater.py
```diff
@@ -0,0 +1,95 @@
...
+}
+
+
+def run(input_path_exclusion_reasons: str, input_path_mondo_schema: str):
+	"""Runs script"""
+	# Load updated exclusions
+	# - Drops all rows where all cells are empty
+	df = pd.read_csv(input_path_exclusion_reasons).dropna(axis=0, how='all').fillna('')
+	# - Format to remove whitespace, leading indentation markers, and unneeded MONDO prefix
+	df['codes'] = df['codes'].apply(lambda x: re.sub('.*MONDO:', '', x))
```
joeflack4, line 41: @sabrinatoro You can keep the leading indentation pattern; it will remove.

src/scripts/exclusion_reasons_enum_updater.py
matentzn: Just keep in mind that this should be changed in production if we want to keep these scripts transparent.

matentzn (approved):
Super cool @joeflack4 didnt even think you would write an entire script. :)

joeflack4:
Kinda glad I did. For these amount of updates, probably was just about as fast to script it. The multiline formatting and more makes the copy/pasting a little bit more time consuming.

@cmungall Not sure who best to tag on the LinkML team to glance at this, but I realize that we're actually updating a LinkML `.yml` file from a GoogleSheet here, so I'm wondering if SchemaView is something that would have been useful here.

src/scripts/exclusion_reasons_enum_updater.py
joeflack4: Hmm, I understand changing them in production in order to discourage simply running the file without knowing what it does or where the inputs are; but RE: transparency, not sure what you mean specifically.

Perhaps we should leave this as it is now until we have an automated way to get stuff from GoogleSheets to CSV programmatically.

There's a way to do that in Python, but would be nice if there was an easy way to do this from the shell. Maybe there's a way. The Python way I know still involves a lot of code, and multiple JSON files which have the API key in them; kinda messy.

src/schema/mondo.yaml
```diff
@@ -1,99 +1,347 @@
 id: https://w3id.org/mondo
```
joeflack4, line 1: @sabrinatoro @matentzn Changes I just made to scripting and the generated `.yml`:
1. `is_a` added
2. `exclude_children` removed
3. If a cell in the GoogleSheet is empty, there will be no field in the Enum. E.g. if there is nothing in the `notes` field for an enum, the `notes` field will not appear. Previously yesterday, it would have shown: `notes: ''`.

nicolevasilevsky (approved):

joeflack4:
@matentzn Merging this now. It's a small PR and doesn't affect anything else. But if this is too early and you want more time to review in the future, let me know.

PR #5386, opened 2022-09-13
renamed some 'disorder of sexual differentiation' terms

### LINKED ISSUE #4949: difference of sexual differentiation
difference of sexual differentiation

**Suggested new label:**
disorder of sexual development (this is by far the most commonly used term for this class of diseases).
@matentzn can we automatically infer class membership? for instance, 46,xx sex reversal is not a subclass. Should we be annotating this manually?

sabrinatoro:
also reviewed the classification.
Fixes #4949


src/ontology/mondo-edit.obo
```diff
@@ -22740,7 +22740,7 @@ replaced_by: MONDO:0016567
 
 [Term]
 id: MONDO:0001571
```
nicolevasilevsky, line 22742: could you please link the term tracker item to the terms you changed?

(I tried to do this myself, but when I open this branch, I don't see your changes, not sure what is going on).

We should add term tracker items to any changed terms going forward - it's a great way to track the rationale for our changes.

nicolevasilevsky (commented):
see comment

src/ontology/mondo-edit.obo
sabrinatoro: done

src/ontology/mondo-edit.obo
```diff
@@ -22754,10 +22754,11 @@ xref: SCTID:198112004 {source="DOID:12698"}
 xref: SCTID:266646002 {source="DOID:12698"}
 xref: SCTID:4754008 {source="MONDO:equivalentTo", source="DOID:12698"}
 xref: UMLS:C0018418 {source="NCIT:C3073", source="MONDO:equivalentTo", source="DOID:12698"}
-is_a: MONDO:0002145 {source="DOID:12698"} ! difference of sexual differentiation
+is_a: MONDO:0002145 {source="DOID:12698"} ! disorder of sexual differentiation
 is_a: MONDO:0002657 {source="MESH:D006177", source="NCIT:C3073/inferred"} ! breast disorder
 intersection_of: MONDO:0001100 ! hypertrophy of breast
 intersection_of: disease_has_location UBERON:0003101 ! male organism
+property_value: IAO:0000233 https://github.com/monarch-initiative/mondo/issues/4949
```
nicolevasilevsky, line 22761: we have to add type xsd:anyURI with the term tracker items too (this is a new thing)

image

sabrinatoro: what!? oh, I guess it is because it is not included in the AP "conditions"...

sabrinatoro: What is the difference between using the IRI Editor (where you add the IRI) and using "literal" (where you have to pick the xsd:anyURI)? 
I am assuming that in this case I have to use the "literal"?

nicolevasilevsky (approved):
perfect, thanks!

PR #5491, opened 2022-10-11
Filling in `object_label` for *.sssom.tsv file

### LINKED ISSUE #5329: Add `object_label` column to mondo.sssom.tsv
OAK has a method to add labels to a tables on the fly. I don't know if this works for sssom files - it should.

Can we try to modify [this Make goal](https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mondo.Makefile#L442)

add an `object_label` column using  OAK?

This file is my primary concern, but it should work for all sssom files:

https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mappings/mondo.sssom.tsv

This file should be used as the dictionary:

https://github.com/monarch-initiative/mondo-ingest/releases/latest/download/mondo-ingest.owl

hrshdhgd:
Fixes #5329  
- [x] Add oaklib dependency
 - [x] Use OAK to add the `object_label` column n the sssom.tsv

 -  Could use `runoak -i tmp/mondo-ingest.db fill-table test.tsv --allow-missing -o test2.tsv` but `fill-table` accepts `msdf.df` and not `msdf`. Hence wrote python script (`add_object_label.py`) to do the  same


matentzn:
Hey @hrshdhgd,

I dont understand the comment:

> fill-table accepts msdf.df and not msdf

Do you mean that fill table does not preserve the metadata header of an sssom file? If so, the PR here is fine, but make an OAK issue "fill table should respect commented rows" and say two sentences about the SSSOM use case.

hrshdhgd:
Issue opened https://github.com/INCATools/ontology-access-kit/issues/313

matentzn:
thanks!

src/ontology/mondo.Makefile
```diff
@@ -457,8 +457,13 @@ tmp/mirror-efo.json: #mirror/efo.owl
 sssom:
 	python3 -m pip install --upgrade pip setuptools && python3 -m pip install --upgrade --force-reinstall sssom==0.3.17
 
-tmp/%.sssom.tsv: tmp/mirror-%.json | sssom
+.PHONY: oaklib
+oaklib:
```
joeflack4, line 461: This is probably a good approach (I think Nico agrees w/ this as well), updating to the latest version of whatever it is that we rely heavily on which changes frequently. For `mondo-ingest`, I've been simply been `pip install --upgrade`ing everything, but I can change that.

src/scripts/add_object_label.py
```diff
@@ -0,0 +1,41 @@
+from typing import TextIO
```
joeflack4, line 1: Looks simple and like it does the job! I'm guessing that when the [`fill-table`](https://github.com/INCATools/ontology-access-kit/issues/313) issue you opened is fixed, this may be changed from a Python script to an sssom CLI call, as it looks like you suggested.

src/ontology/mondo.Makefile
```diff
@@ -624,6 +629,12 @@ open_%_report:
 mondo_obo:
 	robot convert -i mondo-edit.obo -f obo -o mondo-edit.obo
 
+tmp/mondo-ingest.owl:
+	curl https://github.com/monarch-initiative/mondo-ingest/releases/latest/download/mondo-ingest.owl -L --output $@
```
joeflack4, line 633: I was worried that this wouldn't be updated, but it looks like Nico has been doing monthly updates at least.

joeflack4 (approved):

matentzn (approved):
Looks good - can you commit the updated mappings as well to make sure that they look alright?

hrshdhgd:
Done

joeflack4:
@hrshdhgd Usually I take a first stab with a `git rebase master`. Then if things get too dicey, I do `git rebase --abort`.

I'm surprised so many of the mappings are in conflict. I guess that means that someone updated the mappings on `master` within the last 13 days or so since you started this PR. I'm guessing that if you have just run the command to create the mappings on your branch, your files are probably the more authoritative ones, in which case I'd do something like `git checkout --ours src/ontology/mappings/*` during your rebase.

Looks like the only other file w/ a conflict is `.gitignore`, which should be easy to reconcile manually.

hrshdhgd:
I just pulled the files down from `master` and compared the mappings locally. The only difference is in the `mapping_set_id` where the date of release is different. The data itself is pretty much the same. I think I can sort this out myself. Thanks for the advice @joeflack4 !

matentzn (approved):

PR #5602, opened 2022-11-11
Workflow that runs `ontobot-change-agent` when relevant issue is created/edited.

hrshdhgd:
If an issue is made relevant to editing the ontology file, this workflow will run to trigger a PR. This includes 3 files:
1. The workflow YAML file itself
2. TOML file that dictates dependencies
3. Lock file as source of truth for package versions.


matentzn (approved):
Its fine for now, but ultimately I would prefer we use ODK for this!

matentzn:
(we can already use ODK as I regularly publish dev snapshots with the latest versions)

cmungall (approved):

cmungall:
@matentzn - agreed! Let's go ahead with this PR for now and circle back to ODK

PR #5677, opened 2022-11-30
rare diseases sparql queries

sabrinatoro:
Addresses #5642
A query already existed, I modified it such that there are 3 "new" queries to get rare diseases: 
1 ) get ALL Mondo term for rare disease (count)
2) get ALL Mondo term for rare disease (list)
3) get only leaf concepts Mondo term for rare disease (count)

queries are based on the method used here: https://mondo.monarchinitiative.org/pages/analysis/

### DIFF
--- a/src/sparql/reports/rarediseases_list.sparql
+++ b/src/sparql/reports/rarediseases_list.sparql
@@ -1,0 +1,38 @@
+prefix IAO: <http://purl.obolibrary.org/obo/IAO_>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix oio: <http://www.geneontology.org/formats/oboInOwl#>
+prefix def: <http://purl.obolibrary.org/obo/IAO_0000115>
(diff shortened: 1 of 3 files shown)


src/sparql/reports/rarediseases_list.sparql
```diff
@@ -0,0 +1,38 @@
...
+
+SELECT distinct ?cls ?term_label
+
+WHERE
+{
+  {?cls a owl:Class;
+  rdfs:label ?term_label ;
+          rdfs:subClassOf+ <http://purl.obolibrary.org/obo/MONDO_0000001> .
+
+  ?cls rdfs:subClassOf+ [
```
matentzn, line 18: this only works on mondo.owl, _not_ on mondo-edit.obo. Just to remember.

matentzn (approved):
Just looking over it it _looks_ ok, but I don't know the assumptions on what counts as a rare disease so much..

PR #5721, opened 2022-12-09
Removed poetry dependency and triggers QC checks

hrshdhgd:
- Removed poetry relevant files
- Now when a PR is generated, it triggers QC workflows like any other PR.


matentzn (approved):
> Now when a PR is generated, it triggers QC workflows like any other PR.

How does it do that?

.github/workflows/ontobot.yaml
```diff
@@ -79,3 +67,4 @@ jobs:
           title: ${{ env.PR_TITLE }}
           base: ${{ github.head_ref }}
           branch: ${{ env.branch-name }}
+          token: ${{ secrets.GH_TOKEN }}
```
hrshdhgd, line 70: @matentzn , when the token is passed here, it authorizes this workflow to kickstart any qc workflow(s) for the repository.

matentzn: Very good to know, thanks!

PR #5744, opened 2022-12-15
split x-ref of susceptibility versus 'immunodeficiency 33'

sabrinatoro:
Addresses #5701

Note: running the NORM let to some unexpected changes, but these seems to only be order of annotations.

@nicolevasilevsky, please check the original issue and check that you agree with the splitting of immunodeficiency33 and the susceptibility term. Thank you.


nicolevasilevsky (approved):
this looks good, good call to split out the susceptibility, thanks!

PR #5829, opened 2023-01-26
DO NOT MERGE: Add subset system

matentzn:
- adding tagging system that automatically tags rare diseases with the rare disease subset tag using sparql
- add a rare disease subset that currently does not work

Add to QC agenda.


sabrinatoro:
- the subset tag seems to be added correctly
Note that some "'bearer of' some rare" are found in the SubClass Of (Anonymous Ancestor) section of Protege because it comes from the parent or grand-parent.

- The command above didn't work, but the following one worked (to a point): 
`sh run.sh make subsets/mondo-rare.owl IMP=false MIR=false PAT=false`
I however get an error: 

Exception in thread "main" java.lang.OutOfMemoryError: Java heap space: failed reallocation of scalar replaced objects  
make: *** [Makefile:332: subsets/mondo-rare.owl] Error 1

@matentzn : this is where I am blocked

- note the line in the Makefile: `subsets/mondo-%.owl` (and in the command line). Does it mean that the name of the subset is already "mondo-rare"?

matentzn:
I will wait for the subset command in ROBOT next week 15th February to continue working on this issue. I have shared with you the slack discussion explaining your error - mention this in our next 1:1.

sabrinatoro:
This PR is addressing #5812

matentzn:
@sabrinatoro you may not have enough memory to generate the subset but this would do it:

```
MEMORY_GB=16 sh run.sh make rare_subset
```

In any case, I have checked in the current subset, and looks ok, 12800 Mondo ids are returned by the rare subset query we wrote - the exact number that shows up in the subset.

sabrinatoro:
copying the comment from the commit here so we can see it:

@matentzn : 
- I added a new subset AP: rare_grouping
- I merged a robot template adding the rare_grouping subset to the parent we might want for classification of the rare-subset.

I didn't change any of the sparql queries used to create the subset because. 

Should we have 2 slims:
(1) one with the main classification (including rare and rare_grouping subtypes),  to be used for groups like NORD, where users want to see what "type" of diseases are the rare diseases
(1) one for only rare diseases, to be used for when a list of rare diseases is needed

matentzn:
It will be difficult to manage two subsets because of their interactions (three would be possible though, 1 grouping only, 1 rare only, and 1 combine, but before we do that lets make sure there is a point in having three - Is there?)

matentzn:
I pushed a version of the rare subset with your groupings

sabrinatoro:
Thank you @matentzn 
I added the "rare_grouping" subset tag to more grouping terms (there are a lot of rare disease terms without parents). Could you please re-run the rare subset? Thank you. 

WRT the possibility of having 3 different subsets: let's talk to the users and figure out whether they need a subset with only the rare disease terms or not.

matentzn:
@sabrinatoro 

You can share this link: 
https://raw.githubusercontent.com/monarch-initiative/mondo/ee016331542da52e9e2fbe9dafc81f38bbc77ab5/src/ontology/subsets/mondo-rare.json

With our NORD collaborators (in particular Sean) to start their development process.

Let them know that we will, once they approve the subset, share the correct/official link.

Also, I would recommend setting up NORD with protege so they can download, open and review the OWL file.

PR #5870, opened 2023-02-06
obsoleting group_of_disorder terms that have no metadata and are sourced entirely from ordo.

cmungall:


nicolevasilevsky:
I have a couple questions:
1. before we obsolete terms, we should add 'candidate for obsoletion' and an obsoletion date and wait a couple months before we obsolete the terms, right?
2. do we want to add source annotations to this superclass assertions?

nicolevasilevsky:
@sabrinatoro and I should review all of these reclassifications, right?

I'm looking at the first few and I am not sure that MONDO:0000179 'Neu-Laxova syndrome' should be a child of 'inherited ichthyosis syndromic form'. This does not seem like a great grouping class in general and we should probably consider obsoleting this too.

nicolevasilevsky:
another one- I don't think MONDO:0000212 hypercalcemia, infantile should be a kidney disorder.

cmungall:
> I am not sure that MONDO:0000179 'Neu-Laxova syndrome' should be a child of 'inherited ichthyosis syndromic form'.

This PR doesn't introduce any new entailments

`mondo viz -p i MONDO:0000179  'inherited ichthyosis syndromic form'`

image

We can't fix all problems all at once. The aim of this PR was to start with getting rid of all non-standard groupings which will lead to a smaller number of edges to check. But it may be simpler to start by reviewing the edges

cmungall:
> before we obsolete terms, we should add 'candidate for obsoletion' and an obsoletion date and wait a couple months before we obsolete the terms, right?

that's fine, it's easy to redo this PR any time.

I note that we have some candidates for obsoletion from 2022 - we should just have a script that auto-acts on these.

nicolevasilevsky:
@sabrinatoro I think this can be closed, we should have addressed this with our strategic refinement efforts.

PR #5941, opened 2023-02-22
mapping DO exact

sabrinatoro:


sabrinatoro:
We are having issues with sssom:object_label. 
to do: merge without this AP; we will add it later.

matentzn:
The problem cannot be solved - so we have to go ahead without the object_label unfortunately!

PR #6034, opened 2023-03-08
Add a manually curated seed to pull in terms from modules

matentzn:
This new system allows you now to add some terms to seed file, (imports/manual_seed.txt), which are pulled in when refreshing modules.


nicolevasilevsky (approved):

sabrinatoro:
Thank you @matentzn. I was able to add the NCIT terms I needed. Thanks!

PR #6120, opened 2023-03-24
Change way the mondo ingest dependency is generated

matentzn:
@hrshdhgd can you review carefully?


src/ontology/config/prefixes.csv
hrshdhgd, line 1: So you've refreshed the entire `prefixes.csv` file?

src/ontology/mondo.Makefile
hrshdhgd, line 1: The make file looks good to me.

hrshdhgd (approved):

src/ontology/config/prefixes.csv
matentzn: I copied it from mondo ingest, but turns out i cant use it here because we don't have ODK:dev in mondo.

src/ontology/mondo.Makefile
matentzn: Thanks!

PR #6152, opened 2023-04-01
DO NOT MERGE nephrogenetic classes

pnrobinson:
@matentzn  @cmungall 
I am trying my first mondo PR.
related to https://github.com/monarch-initiative/mondo/issues/5952

I have created a class called nephrogenetic disease and have started to add links to existing terms. In two cases, I added a term called hereditary X because MONDO currently just has X but medically X can have hereditary and other subtypes. For instance, primary focal segmental glomerulosclerosis, and so this is the only way to have a decent hereditary disease subhierarchy.

Before I reach the center of the Earth with this rabbit hole, I would like to ask for some feedback. I am afraid there is no way of doing this with small increments and honestly there are so many other errors that I had to hold back for now.

One issue is that I have added about 10 terms but Protege already incremented by auto term ids to MONDO_0100323 (I started at MONDO_0100000), so I am now officially freaked out.


src/ontology/mondo-edit.obo
```diff
@@ -76573,10 +76573,10 @@ xref: UMLS:C0022680 {source="DOID:898", source="MONDO:notFoundInDiseaseSubset"}
 xref: UMLS:C0085413 {source="NCIT:C84578", source="MONDO:notFoundInDiseaseSubset", xref="MONDO:equivalentTo"}
 xref: UMLS:CN119611 {source="MONDO:equivalentObsolete"}
 is_a: MONDO:0000426 {source="DOID:898", source="MONDO:Redundant"} ! autosomal dominant disease
-is_a: MONDO:0005240 {source="DOID:898", source="EFO:1001496/inferred", source="NCIT:C84578/inferred"} ! kidney disorder
 is_a: MONDO:0005372 ! male infertility
 is_a: MONDO:0019741 {source="EFO:1001496", source="MONDO:Redundant", source="Orphanet:730"} ! familial cystic renal disease
 is_a: MONDO:0020642 {source="NCIT:C84578"} ! polycystic kidney disease
+is_a: MONDO:0100313 {source="DOID:898", source="EFO:1001496/inferred", source="NCIT:C84578/inferred"} ! nephrogenetic disease
```
matentzn, line 76579: @nicolevasilevsky risky this, as the isa changed, but the source annotations moved along with the change..

src/ontology/mondo-edit.obo
```diff
@@ -223804,7 +223805,7 @@ xref: ICD10CM:N04.1 {source="DOID:0111129"}
 xref: MESH:C565831 {source="MONDO:equivalentTo"}
 xref: OMIM:603965 {source="DOID:0111129", source="MONDO:equivalentTo"}
 xref: UMLS:C1858915 {source="MONDO:equivalentTo", source="MONDO:ncbi_mim2gene_medline", source="OMIM:603965"}
-is_a: MONDO:0005363 {source="DC-OMIM:603965", source="DOID:0111129", source="MESH:C565831", source="MONDO:Redundant", source="OMIM:603965"} ! focal segmental glomerulosclerosis
+is_a: MONDO:0100314 {source="DC-OMIM:603965", source="DOID:0111129", source="MESH:C565831", source="MONDO:Redundant", source="OMIM:603965"} ! hereditary focal segmental glomerulosclerosis
```
matentzn, line 223808: Same here (see above comment)

src/ontology/mondo-edit.obo
```diff
@@ -450622,6 +450623,23 @@ intersection_of: MONDO:0100308 ! atactic disorder
 intersection_of: disease_has_basis_in_dysfunction_of UBERON:0004681 ! vestibular system
 created_by: https://orcid.org/0000-0001-5208-3432
 
+[Term]
+id: MONDO:0100313
+name: nephrogenetic disease
+def: "A type of kidney disorder that is characterized Mendelian (autosomal dominant, recessive, or X-linked) or chromosomal etiology and by morphological or functional abnormalities in the kidney or urinary system that dominate the clinical picture." []
+is_a: MONDO:0005240 ! kidney disorder
+property_value: http://purl.org/dc/elements/1.1/date "2023-04-01T19:18:35Z" xsd:dateTime
+created_by: peter
```
matentzn, line 450632: @nicolevasilevsky when you advise @pnrobinson here, it seems Protege is doing it smartly (filling in missing IDS in the middle) but Peter has been assigned an ID range that is overlapping with yours:

https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mondo-idranges.owl#L156

So you need to move the `1` digit one step to the left.

As far as I can see, however, there is no need to redo this PR, but I leave it to you

matentzn (changes requested):
I am not reviewing this PR just advising @nicolevasilevsky on two observations I made looking through it:

- Peter has been assign a wrong ID range
- ISA relations moving around along with their provenance defeats the purpose of provenance

src/ontology/mondo-edit.obo
nicolevasilevsky: thanks @matentzn!
@pnrobinson I assigned you a new ID range

src/ontology/mondo-edit.obo
```diff
@@ -450622,6 +450623,23 @@ intersection_of: MONDO:0100308 ! atactic disorder
...
+
+[Term]
```
nicolevasilevsky, line 450634: @pnrobinson I don't think we need this term, MONDO:0005363 'focal segmental glomerulosclerosis' has a superclass assertion: 'bearer of' some inherited, so it already is the genetic form.

nicolevasilevsky:
> @nicolevasilevsky @matentzn if the general way I am making changes is ok, then I would try to finish up nephrogenetics and at least one additional area before the next release? I could do some work today

Yeah, sounds good! Please go ahead and finish this up. I can review it later.

I made some minor changes on this PR:

1. We should use our ORCID for the created_by annotation. You can set this in your Protege peferences.
2. We add dbxrefs for defintions, I added your ORCID as the dbxref
3. We also add source annotations to the superclass assertions. I added your ORCID to these.
4. I added a logical definition to hereditary pseudohypoaldosteronism (following this pattern: https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/hereditary.yaml)

src/ontology/mondo-edit.obo
nicolevasilevsky: ok! We can do that. :)

pnrobinson:
@nicolevasilevsky what does this mean? "We also add source annotations to the superclass assertions" ?

nicolevasilevsky:
> @nicolevasilevsky what does this mean? "We also add source annotations to the superclass assertions" ?

if you click on the @ symbol next to the superclass, we add a source (often times it is the external terminology ID, or in this case, I added your ORCID, since you determined that this should be th parent.)

image

image

nicolevasilevsky:
I reviewed this and made sure all your revisions have been implemented on a different branch (https://github.com/monarch-initiative/mondo/pull/6207), I will close this.

Issue #5846, opened 2023-01-31
Add mondo-all.sssom.tsv and mondo-xrefs.sssom.tsv to `src/ontology/mappings`


matentzn:
The mapping pipeline currently generates a sssom file with only the `exact` mappings.

For various use cases, including boomer and mondo source sync, we need `mondo-all.sssom.tsv` as well, which simply adds (for all sources allowed in `mondo.sssom.tsv`) also `skos:broadMatch`, `skos:closeMatch`, `skos:narrowMatch`.

We also need mondo-xrefs.sssom.tsv, which contains all mappings where `oboInOwl:hasDbXref` is the predicate, but the `subject_id, object_id` combination _does not appear in `mondo-all.sssom.tsv`.

matentzn:
Also, we should add the correct `mapping_set_id` to the mondo mapping sets.

hrshdhgd:
Mondo Mapping set: http://purl.obolibrary.org/obo/mondo/mappings/mondo.sssom.tsv

Issue #4751, opened 2022-04-11
Mondo mappings should be sorted to reduce diff


matentzn:
All sssom tables when run through split_sssom_by_source should be sorted entirely (like by all columns) and the columns should also be sorted in the sssom preferred order.

Right now, some rows in pandas seem to randomly fluctuate.

matentzn:
Even better, implement https://github.com/mapping-commons/sssom-py/issues/51

can call ssom-py function in split_sssom_by_source

hrshdhgd:
Done: https://github.com/mapping-commons/sssom-py/issues/51

matentzn:
in mondo makefile?

hrshdhgd:
Yes: https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mondo.Makefile#L471

matentzn:
This sorts mondo.sssom.TSV, what about the files that are split out in the python script?

hrshdhgd:
https://github.com/monarch-initiative/mondo/pull/5491

PR #6194, opened 2023-04-11
`mapping_set_id` corrected

hrshdhgd:
Addresses #5846


src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -97,7 +97,7 @@ global_metadata:
   subject_source: http://purl.obolibrary.org/obo/mondo.owl
   license: "https://creativecommons.org/publicdomain/zero/1.0/"
   mapping_date: "2017-09-20"
-  mapping_set_id: "GLOBAL" # Apparently this is the one that matters.
+  mapping_set_id: "http://purl.obolibrary.org/obo/mondo/mappings/mondo.sssom.tsv" # Apparently this is the one that matters.
```
matentzn, line 100: Is this needed when `sssom annotate`?

matentzn (changes requested):
- Many `object_label` columns have disappeared from the split file
- Can we add `object_label` to the main mondo.sssom.tsv file?

src/ontology/metadata/mondo.sssom.config.yml
hrshdhgd: not really. I'll change it.

hrshdhgd:
I think the latest run fixed the missing `object_label` issue. I had to run `make mondo-ingest.owl` and `semsql make tmp/mondo-ingest.db -P config/prefixes.csv` again before `make mappings`.

matentzn (commented):
This looks great! Almost there! Can you 

1. Make sure that we don't have to run anything manually? Maybe update https://github.com/monarch-initiative/mondo/blob/550263a2d62d6a468e2883421de654030c42c5af/src/ontology/mondo.Makefile#L641?
2. Deal with https://github.com/monarch-initiative/mondo/issues/4751? I still see sorting issues in your diff, see for example `src/ontology/mappings/mondo_exactmatch_mesh.sssom.tsv`.

Remember this pipeline is updated with every release, so we need to make sure that when Sabrina runs the release, the files get updated in the correct way.

matentzn (approved):

Issue #5585, opened 2022-11-09
Adding OMIA related inforamtion to MONDO


ImkeTammen:
Dear Sabrina and Nico

I have created a form to facilitate submission of OMIA data to Mondo.

Google Sheet
https://docs.google.com/spreadsheets/d/1L3wnp83F-YCKUpBMVBVv1fjHqU0Rw1g58AjvpE24owA/edit#gid=0

I propose to start with those diseases in OMIA that have a same/similar disease listed in OMIM - where both diseases are caused by mutations in orthologous genes.

Any feedback on the form - in the google sheet or via zoom would be much appreciated.

Imke

ImkeTammen:
Based on discussions in meeting on the 2/12/2022 I have updated the mapping file - see version 2 in:

https://docs.google.com/spreadsheets/d/1L3wnp83F-YCKUpBMVBVv1fjHqU0Rw1g58AjvpE24owA/edit#gid=2111247295

I presented information for two disease in OMIA - can I get some feedback that this is what you had in mind for parent / child? 
I added other columns as discussed and created one line for each 'parent' (OMIA phene entry) and one line for each OMIA phene-species entry)

katiermullen:
Question from @ImkeTammen: I had a bit more of a look at phene-species in regard to characterisation of ‘defect’. There are about 200 entries that are currently listed as ‘unknown’ and I just wanted to check if they should be yes/no.

In this group there are a range of entries relating to animal models of human disease that have been created via genome editing or other forms of genetic modifications. As they are not a ‘naturally’ occurring disease we have sometimes (butnot consistently) listed them as unknown in regards to ‘defect’. Would you want genome edited/transgenic animals listed in MONDO? If yes than I will add a ‘defect – yes’ classification.

From @katiermullen: No, we do not want them in Mondo because they are not naturally occurring diseases. 

Question from @ImkeTammen: Another group of entries currently listed as ‘unknown’ relate to confirmed or suspected disease ‘resistance/susceptibility’ e.g. ‘Spongiform encephalopathy, susceptibility/resistance to’ or  ‘Resistance/susceptibility to lentivirus’ – clearly a quantitative trait – not sure if they should be listed as defect. Resistance to a pathogen is not a defect – susceptibility is. Again would this be something that MONDO would want to include?

From @katiermullen: Please tag the "disease susceptibility" entries in OMIA because we may want to add them later as there is a human disease susceptibility branch in Mondo.

[...]

PR #6273, opened 2023-05-18
add OMIA non-human animal disease with x-species analogs

katiermullen:
This addresses #5585 by adding OMIA non-human animal diseases. Species specific OMIA non-human animal diseases have not yet been added.


katiermullen:
This is an old branch. Going to restart a new one.

PR #6336, opened 2023-06-08
Update ORDO mapping obsolesence status

matentzn:
Related to #6331 

@nicolevasilevsky this is an easy review but should be done carefully.

1. Check 4 random ORDO IDs and make sure they are really obsolete
2. Scroll through the rest of the changes and sanity check

Should be no more than 15 min, probably less. Once this is merge, I do the same for DOID.


src/ontology/mondo-edit.obo
```diff
@@ -93307,7 +93307,7 @@ xref: ICD10CM:K00.0 {source="Orphanet:99798/ntbt", source="Orphanet:99798", sour
 xref: ICD9:520.0 {source="DOID:0050591"}
 xref: MESH:D000848 {source="DOID:0050591", source="MONDO:directSiblingOf"}
 xref: OMIMPS:106600 {source="MONDO:cjm", source="MONDO:equivalentTo"}
-xref: Orphanet:2227 {source="MONDO:equivalentTo", source="DOID:0050591"}
+xref: Orphanet:2227 {source="MONDO:equivalentObsolete", source="DOID:0050591"}
 xref: Orphanet:99798 {source="MONDO:equivalentTo", source="MONDO:preferredExternal", source="DOID:0050591"}
```
cmungall, line 93311: preferredExternal is now lo longer necessary but could be handled in a 2nd pass

cmungall (approved):

src/ontology/mondo-edit.obo
matentzn: https://github.com/monarch-initiative/mondo/issues/6339

matentzn:
I have reviewed this again carefully and reached the conclusion this is fine. @nicolevasilevsky I will merge it, I have spot-checked 5 classes to make sure they are truly rare and scrolled through the diff.

PR #6362, opened 2023-06-16
Add non-human animal disease high level classifications

katiermullen:
This PR adds high level non-human animal disease classifications modeled on human disease classifications in Mondo.

### DIFF
--- a/docs/editors-guide/a-protege-setup.md
+++ b/docs/editors-guide/a-protege-setup.md
@@ -117,3 +117,5 @@ If you edit another ontology besides Mondo, you will need to switch your ID rang
 1. `Date property` http://purl.org/dc/elements/1.1/date
 1. `Date value format` Select ISO-8601
 
+<img width="883" alt="image" src="https://github.com/monarch-initiative/mondo/assets/6722114/ef1a9b25-c7e6-41fd-a6c6-7335e32b78a9">
+


sabrinatoro:
@katiermullen 
let me know if you have questions about how to resolve the QC errors in this PR. 

In addition, could you please add definitions for these new terms according to the pattern [here](https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/nonhuman_disease.yaml).
You can use the "database_cross_reference": MONDO:patterns/nonhuman_disease  (see MONDO:0700049 as an example)
(note: you can add these definitions using a ROBOT template: you just need the ID and the column for definition + xref. Let me know if you need help)

I want to question some terms: these were added directly from human but I don't know if they make sense for non-human (you are the expert, so if you say yes, that's good. I am just wondering)
- 'psychiatric disorder, non-human animal'
- 'occupational disorder, non-human animal'
- 'obstetric disorder, non-human animal'  (this might be ok)
- 'hereditary disease, non-human animal'  (we need to be careful with this one as many diseases in non-human are due to a virus, so the term is ok, but we need to be very careful about the terms we classify underneath this term)
- 'disease related to transplantation, non-human animal'
- 'breast disorder, non-human animal'   (should it be called "mammary gland" instead of "breast"? I don't know whether "breast" is used in non-human)

katiermullen:
@sabrinatoro -

I made the following edits based on your comments:

- 'psychiatric disorder, non-human animal' - keep due to anxiety and related disorders in animals.
- 'occupational disorder, non-human animal' - keep due to service/work animal related disorders.
- 'obstetric disorder, non-human animal' (this might be ok) - keep, same terminology as veterinary medicine. 
- 'hereditary disease, non-human animal' (we need to be careful with this one as many diseases in non-human are due to a virus, so the term is ok, but we need to be very careful about the terms we classify underneath this term). Removed cross-species analog EC. Will need a rule for using this concept.
- 'disease related to transplantation, non-human animal' - transplantation occurs in model organisms and some owned-animals. 
- 'breast disorder, non-human animal' (should it be called "mammary gland" instead of "breast"? I don't know whether "breast" is used in non-human) - changed to "mammary gland"

sabrinatoro (approved):
This is fantastic @katiermullen! Thank you so much! 

@nicolevasilevsky  I am merging this PR.

PR #6422, opened 2023-07-01
Pipeline for british language synonyms

matentzn:


src/ontology/mondo.Makefile
```diff
@@ -853,4 +858,12 @@ update-%-mappings: $(TMPDIR)/new-exact-matches-%.owl
...
 		make NORM
-		mv NORM $(SRC)
\ No newline at end of file
+		mv NORM $(SRC)
+
+.PHONY: help
+help:
+	@echo "$$data"
+	echo "Making sure all english in Mondo in American:"
+	echo "sh run.sh make americanize"
```
joeflack4, line 867: We're all living in America

PR #6433, opened 2023-07-05
Update ontobot.yaml

hrshdhgd:
This PR:
- [x] Install the latest `ontobot-change-agent` in the workflow
- [x] Passes the prefix `MONDO` needed for new term request template processing.


matentzn (approved):
What does -p MONDO do?

PR #6442, opened 2023-07-12
redux of mass obsoletion

cmungall:
Replaces #5870 (although the procedure is slightly different)

Workflow: https://github.com/cmungall/mondo-refactoring

Corresponds to these artefacts:

https://github.com/cmungall/mondo-refactoring/tree/77b5f39bc149290f3d43eef0c07c614d61d602bb

- intersection of ordo-only classes and ancestors of omim-equivs
- some terms cannot be obsoleted until logical-defs referencing them are removed

### DIFF
--- a/src/patterns/linkml/mondo-schema.yaml
+++ b/src/patterns/linkml/mondo-schema.yaml
@@ -1,0 +1,104 @@
+# linkml schema for mondo
+id: https://w3id.org/linkml/examples/mondo
+name: mondo
+prefixes:
+  MONDO: https://purl.obolibrary.org/obo/MONDO_
+
+classes:
+
+  AnatomicalStructure:
+
+  CellType:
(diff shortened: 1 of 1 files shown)


cmungall:
I looked at the first one:

FAIL Rule ../sparql/qc/mondo/qc-omim-subsumption.sparql: 38 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0007078,http://www.w3.org/2000/01/rdf-schema#subClassOf,http://purl.obolibrary.org/obo/MONDO_0019182

It looks like this is just revealing already present issues? The QC check should check for transitive subclasses. The rewiring just revealed an existing issue. http://purl.obolibrary.org/obo/MONDO_0019182 is a grouping for a polygenic trait https://omim.org/entry/601665

cmungall:
Replaced by #6447

PR #6454, opened 2023-07-18
Added keywords to pandas function arguments

hrshdhgd:
Pandas' new version complaints if it is not done this way.

### DIFF
--- a/src/scripts/split_sssom_by_source.py
+++ b/src/scripts/split_sssom_by_source.py
@@ -76,8 +76,8 @@ df["mapping_justification"]="semapv:ManualMappingCuration"
 subject_prefixes_allowed = meta['subject_prefixes']
 relations_allowed = meta['relations']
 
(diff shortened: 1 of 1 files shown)


matentzn (approved):

src/scripts/split_sssom_by_source.py
```diff
@@ -76,8 +76,8 @@ def read_metadata(filename):
 subject_prefixes_allowed = meta['subject_prefixes']
 relations_allowed = meta['relations']
 
-subject_prefixes=set(df['subject_id'].str.split(':', 1, expand=True)[0])
-object_prefixes=set(df['object_id'].str.split(':', 1, expand=True)[0])
+subject_prefixes=set(df['subject_id'].str.split(pat=':', n=1, expand=True)[0])
```
joeflack4, line 79: I wonder if the order of the arguments changed. Or maybe they just prefer the keyword args rather than positional. Either way it is honestly clearer to me what the code is doing this way. That it now works is the most important thing!

PR #6465, opened 2023-07-22
add cross-species analog AP from axioms

### LINKED ISSUE #6464: create AP for cross-species analog
We create axioms connecting non-human animal diseases to human diseases. These axioms are used for classification purposes and are removed before the release (because they are not correct and should not be passed to the children terms). The relation between human and non-human disease seems useful for the Monarch prediction tools though. while removing them from the ontology is the correct thing to do, we should somehow leverage them in the knowledge graph.
Question: do we (should we) export/keep the relation between human and non-human diseases to be used in the Monarch tools? 

Response: we can create an annotation property for this
From Nico: 
you can model the semapv:crossSpeciesExactMatch assertions after the sparql query called “construct-mondo-rare-subset.sparql”.

sabrinatoro:
closes #6464

This PR is to add the AP annotation : `crossSpeciesExactMatch (some Mondo human disease)` to the non-human diseases with a "cross-species analog some 'Mondo human disease'" annotation.

@matentzn could you please review this PR?
In addition, I have specific questions:

1- the sparql-construct includes _only_ `cross-species analog` when it is used in a Subclasses_Of axiom. There are a lot of non-human diseases with the `cross-species analog` in the logical definition. I don't know how to include them (the one in the logical definition)

2- I don't know which one should be used in the sparql-construct:
INSERT (same as construct-rare-subset.sparql)
CONSTRUCT (same as construct-orphanet-rare-subset.sparql)

3- The new AP `<https://w3id.org/semapv/vocab/crossSpeciesExactMatch>` is not currently in Mondo. Should I add it manually (by updating the import?), or will it be added automatically when this construct process is in place?

4- will adding these AP be fully automated? 
a) will it automatically add **or remove** these AP based on subclassOf and logical definitions?
b) when is this going to happen? Every time a PR is submitted or during the release process?

Thank you!!


matentzn:
> 1- the sparql-construct includes only cross-species analog when it is used in a Subclasses_Of axiom. There are a lot of non-human diseases with the cross-species analog in the logical definition. I don't know how to include them (the one in the logical definition)

Its a bit cumbersome.. I have done it now! Your solution was Great!

> 2- I don't know which one should be used in the sparql-construct:
INSERT (same as construct-rare-subset.sparql)
CONSTRUCT (same as construct-orphanet-rare-subset.sparql)

I was not sure before, I made it an INSERT now! If you want, check out the query how it is now!

3- The new AP <https://w3id.org/semapv/vocab/crossSpeciesExactMatch> is not currently in Mondo. Should I add it manually (by updating the import?), or will it be added automatically when this construct process is in place?

> eh... okay. We should probably import it, but.. No worries now. Lets leave it as is!

> 4- will adding these AP be fully automated?
a) will it automatically add or remove these AP based on subclassOf and logical definitions?

YES

> b) when is this going to happen? Every time a PR is submitted or during the release process?

During the release! They wont show up in the edit file, only in the release file.

sabrinatoro:
@matentzn Thank you! 
I assume we will have to wait until the next release to test that these are being added correctly. Is that right?

sabrinatoro:
@matentzn 
The terms with "cross-species analog" in their logical definition do not get the AP when they also have the NCBI taxon in their logical definition. 
Eg: bovine neoplasm does not get the AP "neoplasm", though it has a logical definition: 
Screenshot 2023-07-24 at 1 22 07 PM

I created https://github.com/monarch-initiative/mondo/pull/6477 based on this branch. I hope this is the correct solution (it looked like it worked)

sabrinatoro:
the work was done through https://github.com/monarch-initiative/mondo/pull/6477
Closing this PR.

PR #6466, opened 2023-07-24
Extend OMIM check to look for ancestors rather than only direct parents

matentzn:


twhetzel (approved):
I thought the `rdfs:subClassOf+` returned the direct and indirect subclasses vs. ancestors mentioned in the PR title, but I gather this is the change you want to make so approving.

matentzn:
@twhetzel I have now assigned this PR to @sabrinatoro for triage, as there is quite a bit of manual curation involved into fixing it. @sabrinatoro you let us know if you need help with this! As a reminder, this is the implementation of the following proposition:

We do not want to have one OMIM turn subclass of another OMIM term unless this is _specifically_ reviewed and approved. Before this PR, we only checked cases where the _direct_ parent of an OMIM disease was an OMIM disease. This PR increases the severity of this checking by checking cases where an OMIM disease has an _ancestor_ that is an OMIM disease.

src/sparql/qc/mondo/qc-omim-subsumption.sparql
```diff
@@ -11,7 +11,7 @@ prefix mondoSparqlQcMondo: <http://purl.obolibrary.org/obo/mondo/sparql/qc/mondo
 prefix mondoPatterns: <http://purl.obolibrary.org/obo/mondo/patterns/>
 
 SELECT DISTINCT ?entity ?property ?value WHERE {
-   ?entity rdfs:subClassOf ?parent .
+   ?entity rdfs:subClassOf+ ?parent .
```
joeflack4, line 14: _Nico said_:
>  This PR increases the severity of this checking by checking cases where an OMIM disease has an ancestor that is an OMIM disease.

LGTM!

Issue #5548, opened 2022-11-01
OBO Academy exercise: relabel term


sabrinatoro:
**Mondo term (ID and Label):**
bronchiolitis obliterans with obstructive pulmonary disease

**Suggested new label:**
bronchiolitis obliterans

**!!!! This issue is for ontology training only !!!!!!**

sabrinatoro:
Remember to 
- create a new branch
- create a PR (after you made changes)
- refer to this issue in your PR
- tag Sabrina (or someone else) to review your work

sabrinatoro:
For demo/teaching only. Closing.

PR #6473, opened 2023-07-24
change name for MONDO:0015265

ehurwitz:
Addresses #5548


sabrinatoro:
Great job @ehurwitz. You can see here that your checks have failed. 
We can review why and how to proceed when we talk on Wednesday. 
Thank you!

PR #6477, opened 2023-07-24
Issue 6464 taxonspecific

sabrinatoro:
see comment in https://github.com/monarch-initiative/mondo/pull/6465

I updated the query to include the logical definitions which include "in taxon"


sabrinatoro:
@matentzn : I created this PR from [this PR](https://github.com/monarch-initiative/mondo/pull/6465) in case that was the wrong thing to do.

matentzn:
It looks ok but it needs to be tested if it works!

sabrinatoro:
Thank you @matentzn. It did work when I tested it, but still wanted to make sure I didn't make the query weirder than it needed to be. 

@katiermullen could you please be another pair of eyes and test this PR?
All the cross-species analog axioms (in subclassOf and in logical definitions) should be found as annotation property annotations. Please see the original PR and issue for more details. 
Here is how you can test: 

sh run.sh make mondo.owl MIR=false IMP=false
Then when it's done you open mondo.owl in Protege

Thank you!

nicolevasilevsky (approved):

sabrinatoro:
@nicolevasilevsky I see that you already merged this PR. 
did you test this as mentioned above? (I don't think that Katie has done it yet). 

I "think" it is ok... I guess we'll see soon enough :)

PR #6506, opened 2023-07-28
Rare subset revision

matentzn:
partially addresses #6448

This PR revises our subset approach according to what was discussed in #6448. It does not cover the bearer of some rare axioms, just the way the subsets are generated.

@sabrinatoro requires cursory but not too detailed review. Ideally, look at mondo-edit.obo on this branch and note what you find.


sabrinatoro:
@matentzn 
1) I can see all subsets, including the "inferred_rare" though it currently looks like this: 

Screenshot 2023-07-28 at 8 41 15 AM


2) (probably not this PR, but still mentioning): there is no axiom "bearer_of some rare"
3) (probably not this PR, but still mentioning): there is no terms in the mondo_rare subset. I need to get all the terms with "bearer of some rare" that have a source that is different from orphanet and update these.

matentzn:
@sabrinatoro exactly - do you want to create the inferred_rare subset on this branch here so we have it? The rest will come on a different branch!

sabrinatoro:
@matentzn: inferred_rare subset AP has been created.

sabrinatoro (approved):

matentzn:
We cant merge this yet. We need a better way to determine "rare" from orphanet first.. Because `MONDO:disease` is aligned with ORDO, which means that all its subclasses are, which means that all of Mondo is.. rare. 

@sabrinatoro can you figure out a way how we can get from ORDO a list of all rare diseases? I have now only excluded the to MONDO:disease class, but I am assuming that this wont be enough.

matentzn:
Removing the MONDO:disease class made a big difference. I have now added four tables in this PR:

1. reports/old-rare-diseases.tsv (the rare disease list in the latest release)
2. reports/new-rare-diseases.tsv (the rare disease list if we go ahead with this PR)
3. reports/added-rare-disases.tsv (these is the rare diseases that were previously not in the list, and now are)
4. reports/removed-rare-diseases.tsv (these are the rare diseases that previously were in the list, and are now not anymore)

@sabrinatoro you will need to check out the last two and advice on how to proceed!

sabrinatoro:
> @sabrinatoro can you figure out a way how we can get from ORDO a list of all rare diseases? I have now only excluded the to MONDO:disease class, but I am assuming that this wont be enough.

@matentzn I am not following what the problem is. 
you have removed all the axiom "bearer of some rare"? Or you are not able to remove these axioms? Or we don't know what orphanet class is actually rare?

matentzn:
Our algorithm is: if either Orphanet GARD Nord or mondo say: it's rare, it's rare. The problem is what it means to be rare according to Orphanet. If we say: "it's rare because it has a mapping to ORDO" then we pull in all of Mondo, because the Ordo:disease term is aligned with the mondo:disease root..

sabrinatoro:
Thank you @matentzn 
I need to think about it and do some research.

sabrinatoro:
@matentzn 
I have reviewed the list of terms that were added/removed to the list of rare diseases. For the terms I checked:
- the terms removed from the list had either no source for being "rare", or their source was a Mondo obsoleted terms. Therefore I think it is ok.
- the terms added to the list of rare diseases either come from new subset annotation (mostly from NORD), or were inferred to be rare because their parent is rare. Therefore I think it is ok. 

I think this is ready to merged? 

I have a robot template to add the manually curated "Mondo_rare" subset. Should I add these to this PR, or should I wait for this PR to be merged?

matentzn:
There are actually _much less_ lost ones - I had a bug. So it looks more comforting.

We can merge, but you should go deep in your soul if you want st like this in your rare disease subset:

image

In principle, we could exclude `ordo_group_of_disorder` from the rare subset, if we so wished. Up to you. We can also go first this way, and revise the decision later.

matentzn:
@nicolevasilevsky this is an important PR, but @sabrinatoro and I looked through it a while. It seems like its good to go to my eyes. Do you want to see if you can spot any major issues?

matentzn:
Actually I reviewed it again, and found a small bug - I am fixing it now (nothing really dangerous)

sabrinatoro:
@matentzn I agree, we can probably not include the grouping terms in the rare subset. The only reason to keep them is if they are on the nord/gard list. I'll take a look

matentzn:
Now its all fixed, and it looks even better than earlier..

src/scripts/filter_rare_disease_list.py
```diff
@@ -0,0 +1,26 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Process and filter two TSV files")
+    parser.add_argument("old_rare_list", help="Path to the first input TSV file")
+    parser.add_argument("new_rare_list", help="Path to the second input TSV file")
+    parser.add_argument("new_rare_diseases", help="Path to the first output TSV file")
+    parser.add_argument("lost_rare_diseases", help="Path to the second output TSV file")
+
+    args = parser.parse_args()
+
+    main(args)
```
joeflack4, line 26: ### Argparse style
I use a few different patterns but I prefer `main(**vars(parser.parse_args()))` to `main(args)`.

This will make it so that you are passing normal keyword arguments to `main()`. That way, you can define your parameters in `main()` and you can also call it directly rather than just through the CLI.

It's not a big deal though. What's here is not very verbose, which is also advantageous. If you want you can compare to [migrate](https://github.com/monarch-initiative/mondo-ingest/blob/bugfix-omimps-labels/src/scripts/migrate.py), which has some other differences.

src/scripts/filter_rare_disease_list.py
```diff
@@ -0,0 +1,26 @@
+import argparse
+import pandas as pd
+
+def main(args):
+    # Read the input files into pandas dataframes
+    old_df = pd.read_csv(args.old_rare_list, sep='\t')
+    new_df = pd.read_csv(args.new_rare_list, sep='\t')
+
+    # Filter the dataframes
+    df_added = old_df[~old_df['ID'].isin(new_df['ID'])]
```
joeflack4, line 10: ### Implementation
Implementation is fine to me. Really straightforward and quick.

---

I saw that I am set as a 'code owner'f or Python. @matentzn If you want me to review other stuff here in addition to Python files, let me know.

src/scripts/filter_rare_disease_list.py
matentzn: TBH, I asked ChatGPT to generate this script and didn't even look at it much. I don't like Argparse so much anymore, I prefer click. But for a small ad-hoc script like this where I i just didn't know how to do it outside python, its fine. Thanl you for the feedback though! Very interesting!

src/scripts/filter_rare_disease_list.py
matentzn: @joeflack4 thanks for your review in general! I will assign you if I need a specific review, else I kinda like your eyes on things if you see it fly by, whenever you have time!

PR #6528, opened 2023-08-02
Add QC check for cross species analog diseases

### LINKED ISSUE #6523: Create QC check to catch cross-species analog error
The QC should check for non-human disease terms (MONDO:0005583): 
the logical definition and subclass of annotations using the relation "cross-species analog" (MONDO:0700097) should only use terms from the human disease branch, and no term from the non-human disease branch.

matentzn:
Fixes #6523 

!!!Be careful not to merge with test error in.!!!


matentzn:
@sabrinatoro this is now ready for review and fixing.

matentzn:
Yes didn't even think of that :) that's right!

sabrinatoro:
@matentzn I have updated the QC check to allow the use of the "crossSpeciesExactMatch" property. 
Note that I couldn't figure out which prefix to use for this property, so I used the whole IRI.

matentzn:
Great!

sabrinatoro:
@matentzn, I don't know what the problem is with the query. I am sure it is very simple and something ridiculous like parentheses or something like that. Could you please take a look? Thank you!

matentzn:
Yea!

PR #6560, opened 2023-08-15
Repair UMLS conflicts in Mondo.

joeflack4:
## Overview

The Goal of this PR is to fix the UMLS references in Mondo based on the curated corrections from @kanems. The PR is one-off -> we are removing ca 2600 non-exact UMLS mappings and replacing them by new, updated UMLS mappings provided by MedGen. 

## Tasks
Original tasks as outlined in [Nico's comment](https://github.com/monarch-initiative/mondo-ingest/issues/273#issuecomment-1653431573) in [the issue](https://github.com/monarch-initiative/mondo-ingest/issues/273):
- [x] 1. Remove bad xrefs / conflicts from Mondo as identified by Megan
- [x] 2. Add new xrefs as identified by Megan
- [x] 3. Add new xrefs generated from [MedGen ingest](https://github.com/monarch-initiative/medgen): which will mainly come from the FTP file `MedGenIDMappings.txt`, but will probably require that whole ingest to be working to generate.

Additional tasks that came up during the PR:
- [x] 4. [Remove UMLS:CN# sources](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1297843256) (completed by #6573)
- [x] 5. [Remove `MEDGEN:` xrefs and sources](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1297844406)

## Addresses
- https://github.com/monarch-initiative/mondo-ingest/issues/273


src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,12 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
 rare-disease-reports: reports/old-rare-diseases.tsv reports/new-rare-diseases.tsv
 	python ../scripts/filter_rare_disease_list.py reports/old-rare-diseases.tsv reports/new-rare-diseases.tsv reports/added-rare-disases.tsv reports/removed-rare-diseases.tsv
 
-	
+#############################################
+##### One-time scripts ######################
```
joeflack4, line 268: ## One-time scripting
@matentzn There are enough steps here that I want to codify this rather than do things manually. I also want you/others to be able to check my work too.

I think since this will be a one time thing, perhaps what we can do is at the end of this PR, I can add another commit to remove any make commands / intermediate Python scripts, and leave only the updated `mondo-edit.obo` and the added robot template. What do you think?

src/scripts/collate_bad_medgen_xrefs.py
```diff
@@ -0,0 +1,45 @@
+"""Collate bad medgen xrefs
+
+From excel file containing workseets about MedGen conflicts, create a text file that lists all the bad CUIs for xrefs
+that need to be removed from Mondo.
+
+See also: https://github.com/monarch-initiative/mondo-ingest/issues/273
+
+Prerequisites:
+1. Python package 'openpyxl' to read .xlsx (left out of requirements.txt since this is a temporary script)
```
joeflack4, line 9: Requires `openpyxl`, but didn't put this in `requirements.txt` as I am thinking this process is temporary considering making a final commit that removes it all.

src/scripts/bad_medgen_xrefs_update_analyze_diff.py
```diff
@@ -0,0 +1,87 @@
+"""MedGen conflict updates QC analysis
```
joeflack4, line 1: ### Analysis of `grep -v` results
TLDR: I think this method did not work out.
Probably should instead do (a) SPARQL update, (b), manual/partial editing based on my analysis results (c) remove all xrefs and add from scratch w/ new ingest as originally planned.

---

I analyzed the 5 sheets in [Megan's excel](https://docs.google.com/spreadsheets/d/13CO5yAAhMZD3BhAK8S1GfLF6ZNfDZdcW/edit#gid=61010183). For each sheet, I was able to identify which xrefs to MedGen/UMLS IDs needed to be removed, totaling 2,934.

However, after doing `grep -v` on `mondo-edit.obo` on these IDs, there were more than that number of lines removed, totaling 3,125. Here's a breakdown of my analysis ([report-qc-medgen-conflicts-update-diff.tsv.zip](https://github.com/monarch-initiative/mondo/files/12353911/report-qc-medgen-conflicts-update-diff.tsv.zip)).

```
n_lines_removed 3125
- n_non_xrefs_removals     6
  - synonyms: 4
  - comments: 1
  - relationships: 1
- n_xrefs_removed     3119
  - n_xrefs_removed_cuz_umls_was_a_source   60
  - n_umls_removals 3059
    - n_single_cui_removed_from_xrefs_n_times__1   2819 (1x2819=2819)
    - n_single_cui_removed_from_xrefs_n_times__2   106 (2x106=212)
    - n_single_cui_removed_from_xrefs_n_times__3   8 (3x8=24)
    - n_single_cui_removed_from_xrefs_n_times__4   1 (1x4=4)
```

For each indent line you see here, the indented level below it contains all the different cases that add up to the total on that line.
Out of 3,125 removals, 6 of those were mistakes; they were not xrefs (see below for detail).
3,119 xrefs were removed. However, 60 of those were mistakes; they weren't UMLS/MedGen xrefs. They got removed by `grep` because the MedGen ID was one of the xref sources.
3,059 UMLS/MedGen xrefs were removed. However, sometimes multiple xrefs got removed for a single ID. For 2,819 IDs, that same number of xrefs got removed. 106 IDs were removed 2 times, 8 were removed 3 times, and 1 was removed 4 times _Why this is hard to fix using this `grep -v` approach; hard to explain succinctly_: This is not what Megan/we wanted in most cases. There is only 1 sheet, I believe, where Megan lists out MedGen IDs that no longer exist and that we should remove. But the other sheets talk about specific xrefs on specific Mondo classes. So this number of 3,059 is a problem. We were looking to remove around 2,934 IDs (total number of unique IDs in Megan's excel). It's possible because of that 1 sheet just mentioned, that number could be higher than 2,934. I'd need to re-run this removal pipeline on that sheet alone, but I have a feeling that this difference of excess xrefs removed is likely not coming from the CUIs in that 1 sheet. The crux of it is that even if we solve the issue of non-umls xrefs being removed by the `grep -v` approach, there is still a big problem of too many xrefs being removed here.

<details><summary>Non-xref removals (4 synonyms, 1 comment, 1 relationship)</summary>
<p>

synonym: ""Ullrich disease"" RELATED [DOID:0050558, https://www.ncbi.nlm.nih.gov/gtr/conditions/C0410179/]
relationship: excluded_subClassOf MONDO:0016697 {source=""UMLS:CN201941""} ! low grade ependymoma
comment: Other ontologies classify peripheral neuroepithelioma as a seprate class but NCIT treats these as synonyms and we aim to follow NCIT here. (See HP:0006717, SNOMED:254764001, the ""included"" disease concept on OMIM:612219, and UMLS:C3489398).
synonym: ""polysyndactyly"" EXACT [SCTID:84598000, UMLS:C0265553]
synonym: ""thrombophilia, hereditary"" EXACT [MESH:C540694, UMLS:C2584620]
synonym: ""triple positive breast cancer"" EXACT [UMLS:C5243562]

</p>
</details>

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,38 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	python ../scripts/bad_medgen_xrefs_collate.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -o $@
+
+# Creates a command that, when run, will create mondo-edit.obo.tmp
+tmp/bad-medgen-xrefs-grep-command.sh: tmp/bad-medgen-xrefs.txt
+	python ../scripts/bad_medgen_xrefs_grep_command.py -i tmp/bad-medgen-xrefs.txt -t mondo-edit.obo -o $@
+
+# Creates mondo-edit.obo.tmp by using command
+mondo-edit.obo.tmp: tmp/bad-medgen-xrefs-grep-command.sh
+	@read -r grep_cmd < tmp/bad-medgen-xrefs-grep-command.sh; \
+    eval "$$grep_cmd"
```
joeflack4, line 285: ### How I constructed/ran `grep -v`
1. XLSX -> Single .txt of IDs
2. IDs -> construct grep string
3. Run `eval` on grep string

### Eval
I know it's not usually safe to run `eval`, but this is a temporary thing and I'm thinking I'll do a final commit where I remove some/all of these scripts.

Some alternatives I could have done for this: (i) create the grep command via some ShellScript and have make run it, (b) Run the grep command from Python using `subprocess.

joeflack4:
@hrshdhgd @souzadevinicius Just tagging you here to review, even though this PR is not done yet.

src/scripts/bad_medgen_xrefs_grep_command.py
```diff
@@ -0,0 +1,40 @@
...
+    bad_xref_cuis = []
+    with open(input_file, "r") as file:
+        for line in file:
+            bad_xref_cuis.append(line.strip())
+
+    output_file2 = output_file.replace('.txt', '-grep-command.sh')
+    command_string = "grep -v "
+    for s in bad_xref_cuis:
+        command_string += f"-e {s} "
+    command_string += f"{target} > {target + '.tmp'}\n"
```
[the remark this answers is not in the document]
matentzn: Also, this PR here should _only_ be about UMLS.

src/scripts/bad_medgen_xrefs_update_analyze_diff.py
matentzn: I see your point.. It is unfortunate that this has become such a complex issue. Unfortuntely - I don't know how to help you here, you will just have to figure it out somehow! Use whatever means you want!

For the non-xref matches, I would just delete them in isolation (not -v, just the xrefs inside, manually).

> Probably should instead do (a) SPARQL update, (b), manual/partial editing based on my analysis results (c) remove all xrefs and add from scratch w/ new ingest as originally planned.

(a) may be much more cumbersome than you think, because of axiom annotations
(b) yes
(c) I don't think this is easily possible as we may have valid UMLS cross references that are not in Medgen at all..

src/ontology/mondo.Makefile
joeflack4: That's better! Honestly IDK if eval is any less safer, but your recommendation is just more straightforward / simpler. IDK why I didn't think of that. Changed~

src/scripts/collate_bad_medgen_xrefs.py
joeflack4: Nico [thinks it's fine to keep these one-time scripts / goals, etc](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1295660735), so I decided to add this to the requirements file. Maybe it's not entirely necessary, but it's only 1 dependency, with few transitive dependencies.

src/scripts/bad_medgen_xrefs_grep_command.py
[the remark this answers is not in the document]
joeflack4: 1. UMLS
Yes, I was thinking about prepending "UMLS" to these search strings. Probably better to do that. But just FYI I checked the results, and only UMLS:<ID> cases were affected; no other namespaces.
2. `xref: {s}`
Yes, I was thinking about that. That is a good solution to some of [the `grep -v` problems](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1295270141). I'm about to read [your reply](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1295695067) to that. If we choose a different route than `grep -v`, then this point is moot. But if we stay with `grep -v`, I'll make this change.
3. Replace CUIs instead of delete
Maybe. But still doesn't address many [problems](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1295270141). I think what we decide there will decide how we do this.
<details><summary>Some more, maybe redundant thoughts</summary>
<p>

Not all CUIs need to be replaced; some just need to be removed. So we could do replacements where needed (that entails something like SPARQL UPDATE) and deletions where needed. But for the removals, sometimes they're being removed in too many places.
Replacing them here is also in conflict with 2 other options: (a) I just wanted to remove all of these xrefs and bring in new ones from the ingest anyway, (b) assuming not 'b', part '2' of [the OP](https://github.com/monarch-initiative/mondo/pull/6560#issue-1850743276), is about your instruction to provide a robot template to introduce these additions as a 2nd step, which will have same effect as doing the find/replace of CUIs using something like SPARQL UPDATE.

</p>
</details>

src/scripts/bad_medgen_xrefs_update_analyze_diff.py
joeflack4: No problem. I need to / should be able to figure out implementations / workarounds. 

I was keen on 'a' SPARQL UPDATE, and aware of the axiom annotations. I was thinking that they would all be the same "shape", so that my query would only need 2 blocks: (i) to remove the xref, and (ii) to remove the axiom annotation. But I think I'll trust your instinct that (a) if I'm correct, implementing this might be more complicated than it seems, or (b) I'm not correct, and there are different shapes / types of axiom annotations, requiring even more blocks in my query.

I hear you on 'c'. I didn't think of that.

So yes, I think I'll go with 'b', manual manipulations, and figure it out as I go.

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,37 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+#############################################
+##### One-time scripts ######################
+#############################################
+# MedGen conflicts update
+## - main pipeline
+tmp/July2023_CUIReports_FromMedGentoMondo.xlsx:
+	mkdir -p tmp && wget "https://github.com/monarch-initiative/mondo-ingest/files/12029712/July2023_CUIReports_FromMedGentoMondo.xlsx" -O $@
+
+tmp/bad-medgen-xrefs.txt: tmp/July2023_CUIReports_FromMedGentoMondo.xlsx
+	python ../scripts/bad_medgen_xrefs_collate.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -o $@
```
joeflack4, line 276: - Consider using this for find/replace https://github.com/monarch-initiative/mondo/blob/master/src/scripts/clean-british-english.py
- Delete all `Medgen:` refs. can do manually

joeflack4: Nico will handle these 6 cases:
```
non_xref_removals_content     "synonym: ""Ullrich disease"" RELATED [DOID:0050558, https://www.ncbi.nlm.nih.gov/gtr/conditions/C0410179/]
relationship: excluded_subClassOf MONDO:0016697 {source=""UMLS:CN201941""} ! low grade ependymoma
comment: Other ontologies classify peripheral neuroepithelioma as a seprate class but NCIT treats these as synonyms and we aim to follow NCIT here. (See HP:0006717, SNOMED:254764001, the ""included"" disease concept on OMIM:612219, and UMLS:C3489398).
synonym: ""polysyndactyly"" EXACT [SCTID:84598000, UMLS:C0265553]
synonym: ""thrombophilia, hereditary"" EXACT [MESH:C540694, UMLS:C2584620]
synonym: ""triple positive breast cancer"" EXACT [UMLS:C5243562]
```

src/scripts/bad_medgen_xrefs_update_analyze_diff.py
joeflack4: Strategy:
**n_single_cui_removed_from_xrefs_n_times__2+** (~125): Joe will add back xrefs that were accidentally removed.
**n_xrefs_removed_cuz_umls_was_a_source (60)**: Joe will remove manually.
**n_non_xrefs_removals (6)**: Nico removing manually.

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,38 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+
+tmp/mondo-edit.obo.tmp.diff: mondo-edit.obo.tmp
+	diff mondo-edit.obo mondo-edit.obo.tmp > $@
+
+tmp/report-qc-medgen-conflicts-update-diff.tsv: tmp/mondo-edit.obo.tmp.diff
+	python ../scripts/bad_medgen_xrefs_update_analyze_diff.py -i tmp/mondo-edit.obo.tmp.diff -o $@
+
+# address-medgen-conflicts: https://github.com/monarch-initiative/mondo-ingest/issues/273
+.PHONY: address-medgen-conflicts
+address-medgen-conflicts: tmp/report-qc-medgen-conflicts-update-diff.tsv mondo-edit.obo.tmp-stricter
```
joeflack4, line 296: Nico will handle these 6 cases:
```
non_xref_removals_content     "synonym: ""Ullrich disease"" RELATED [DOID:0050558, https://www.ncbi.nlm.nih.gov/gtr/conditions/C0410179/]
relationship: excluded_subClassOf MONDO:0016697 {source=""UMLS:CN201941""} ! low grade ependymoma
comment: Other ontologies classify peripheral neuroepithelioma as a seprate class but NCIT treats these as synonyms and we aim to follow NCIT here. (See HP:0006717, SNOMED:254764001, the ""included"" disease concept on OMIM:612219, and UMLS:C3489398).
synonym: ""polysyndactyly"" EXACT [SCTID:84598000, UMLS:C0265553]
synonym: ""thrombophilia, hereditary"" EXACT [MESH:C540694, UMLS:C2584620]
synonym: ""triple positive breast cancer"" EXACT [UMLS:C5243562]
```

Options:
a. Nico will make changes in a commit here (not preferred; I tend to force push and amend commits)
b. Nico will make changes in master and Joe will rebase (preferred)
c. Nico will make changes after this PR has been merged.

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,32 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	python ../scripts/bad_medgen_xrefs_collate.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -o $@
+
+tmp/bad-medgen-xrefs-grep-command.sh: tmp/bad-medgen-xrefs.txt
+	python ../scripts/bad_medgen_xrefs_grep_command.py -i tmp/bad-medgen-xrefs.txt -t mondo-edit.obo -o $@
+
+mondo-edit.obo.tmp: tmp/bad-medgen-xrefs-grep-command.sh
+	sh $<
+
+tmp/mondo-edit.obo.tmp.diff: mondo-edit.obo.tmp
+	-diff mondo-edit.obo mondo-edit.obo.tmp > $@
```
joeflack4, line 285: `-diff`: If there is a difference between two files, `diff` returns an error code. As a result, this goal wasn't running, since diff "failed":
```make
tmp/report-qc-medgen-conflicts-update-diff.tsv: tmp/mondo-edit.obo.tmp.diff
```
Adding `-` to the front of `diff` fixed that problem.

src/scripts/bad_medgen_xrefs_grep_command.py
```diff
@@ -0,0 +1,40 @@
...
+    """Create grep command"""
+    bad_xref_cuis = []
+    with open(input_file, "r") as file:
+        for line in file:
+            bad_xref_cuis.append(line.strip())
+
+    output_file2 = output_file.replace('.txt', '-grep-command.sh')
+    command_string = "grep -v "
+    for s in bad_xref_cuis:
+        command_string += f'-e "xref: UMLS:{s}" '
```
joeflack4, line 19: In `grep -v`, changed `-e {ID}` to `-e "xref: UMLS:{ID}"`

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,35 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+
+tmp/report-qc-medgen-conflicts-update-diff.tsv: tmp/mondo-edit.obo.tmp.diff
+	python ../scripts/bad_medgen_xrefs_update_analyze_diff.py -i tmp/mondo-edit.obo.tmp.diff -o $@
+
+.PHONY: address-medgen-conflicts-aug2023
+address-medgen-conflicts-aug2023: tmp/report-qc-medgen-conflicts-update-diff.tsv
+	grep -v 'source="UMLS:CN' mondo-edit.obo.tmp > mondo-edit.obo.tmp2.tmp
```
joeflack4, line 292: Added a 2nd `grep -v`. See reason and ramifications in [this comment](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1297818697).

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,35 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	grep -v MEDGEN: mondo-edit.obo.tmp2.tmp > mondo-edit.obo.tmp3.tmp
```
joeflack4, line 293: Added a 3rd `grep -v`
@matentzn This is the removal of all lines that have the text `MEDGEN:`, as you requested at our meeting today.
You recommended that this be done manually, but this was so easy/fast to do programmatically that I preferred to do it that way. I think it's much better doing it programmatically as well, just in case there are any merge conflicts or something else that would necessitate that I re-run this. 

There were about 30 removals.

<details><summary>Diff</summary>
<p>

```diff
2196d2195
< xref: MEDGEN:546803 {source="UMLS:C0275757"}
2213d2211
< xref: MEDGEN:537148 {source="UMLS:C0259737"}
2346d2343
< xref: MEDGEN:540779 {source="UMLS:C0267556"}
2992d2988
< xref: MEDGEN:575966 {source="UMLS:C0343965"}
3485d3480
< xref: MEDGEN:547234 {source="UMLS:C0276286"}
3533d3527
< xref: MEDGEN:547314 {source="UMLS:C0276386"}
13413d13406
< xref: MEDGEN:508459 {source="UMLS:C0149948"}
16128d16120
< xref: MEDGEN:11567 {source="UMLS:C0038041"}
26833d26824
< xref: MEDGEN:509829 {source="UMLS:C0155174"}
69691d69681
< xref: MEDGEN:208928 {source="UMLS:C0854912"}
151847,151848d151836
< xref: MEDGEN:358137 {source="UMLS:C1868139"}
< xref: MEDGEN:881357 {source="UMLS:C4054549"}
172389d172376
< xref: MEDGEN:82812 {source="UMLS:C0268501"}
350178d350164
< synonym: "XYLT1-CDG - xylosyltransferase 1 congenital disorder of glycosylation" EXACT [MEDGEN:C4750849]
371444d371429
< xref: MEDGEN:231044 {source="UMLS:C1332210"}
404531d404515
< xref: MEDGEN:508876 {source="UMLS:C0152081"}
404569d404552
< xref: MEDGEN:508879 {source="UMLS:C0152087"}
404588d404570
< xref: MEDGEN:57754 {source="UMLS:C0152222"}
404625d404606
< xref: MEDGEN:102349 {source="UMLS:C0156307"}
404825d404805
< xref: MEDGEN:60214 {source="UMLS:C0206708"}
408322d408301
< xref: MEDGEN:447610
409362d409340
< xref: MEDGEN:575892 {source="UMLS:C0343875"}
433950d433927
< def: "Deficient blood distribution to the limbs caused by narrowing or obstruction of the lumen of the peripheral arteries." [MEDGEN:115991]
434966d434942
< xref: MEDGEN:419036 {source="UMLS:C2931284"}
434999d434974
< xref: MEDGEN:75782
435090d435064
< xref: MEDGEN:452941 {source="UMLS:C1291560"}
435118d435091
< xref: MEDGEN:6300 {source="UMLS:C0025306"}
435162d435134
< xref: MEDGEN:418965 {source="UMLS:C2931030"}
435219d435190
< xref: MEDGEN:224821 {source="UMLS:C1279186"}
441678d441648
< xref: MEDGEN:233758 {source="UMLS:C1334968"}

```

</p>
</details>

src/ontology/mondo.Makefile
```diff
@@ -264,7 +264,35 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	sh $<
+
+tmp/mondo-edit.obo.tmp.diff: mondo-edit.obo.tmp
+	-diff mondo-edit.obo mondo-edit.obo.tmp > $@
+
+tmp/report-qc-medgen-conflicts-update-diff.tsv: tmp/mondo-edit.obo.tmp.diff
+	python ../scripts/bad_medgen_xrefs_update_analyze_diff.py -i tmp/mondo-edit.obo.tmp.diff -o $@
+
+.PHONY: address-medgen-conflicts-aug2023
+address-medgen-conflicts-aug2023: tmp/report-qc-medgen-conflicts-update-diff.tsv
```
joeflack4, line 291: @matentzn I've now committed changes to `mondo-edit.obo` if you want to examine.

matentzn:
- `xref: GTR:AN1112503 {source="UMLS:CN706304"}` is removed, but should not. Remember #6573! This PR needs to be merged first before your pipeline is run!
- You remove stuff like 
    - `+def: "Deficient blood distribution to the limbs caused by narrowing or obstruction of the lumen of the peripheral arteries." [MEDGEN:115991]`
    - `+synonym: "XYLT1-CDG - xylosyltransferase 1 congenital disorder of glycosylation" EXACT [MEDGEN:C4750849]`
    - Please do not remove these! only `^xref:` matches!
- If a row contains one of the following strings, do not remove it: `MONDO:preferredExternal`, `https://orcid.org` (indicates someone has manually removed it)
- Question: for all the UMLS CUIs we remove, will we get one back from the pipeline?

src/ontology/mondo.Makefile
```diff
@@ -268,8 +268,55 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
 rare-disease-reports: reports/old-rare-diseases.tsv reports/new-rare-diseases.tsv
 	python ../scripts/filter_rare_disease_list.py reports/old-rare-diseases.tsv reports/new-rare-diseases.tsv reports/added-rare-disases.tsv reports/removed-rare-diseases.tsv
 
-	
+#############################################
+##### One-time scripts ######################
+#############################################
+# MedGen conflicts (Aug 2023) pipeline
```
joeflack4, line 274: _Nico wrote:_
- `xref: GTR:AN1112503 {source="UMLS:CN706304"}` is removed, but should not. Remember #6573! This PR needs to be merged first before your pipeline is run!
- You remove stuff like 
    - `+def: "Deficient blood distribution to the limbs caused by narrowing or obstruction of the lumen of the peripheral arteries." [MEDGEN:115991]`
    - `+synonym: "XYLT1-CDG - xylosyltransferase 1 congenital disorder of glycosylation" EXACT [MEDGEN:C4750849]`
    - Please do not remove these! only `^xref:` matches!
- If a row contains one of the following strings, do not remove it: `MONDO:preferredExternal`, `https://orcid.org` (indicates someone has manually removed it)
- Question: for all the UMLS CUIs we remove, will we get one back from the pipeline?

joeflack4: Hey @matentzn, good review. I copy/pasted your comment here because I think, to reduce clutter and ensure everything is completed, it's a good idea to make as many comments possible from the "files changed" tab, so that that we get the "resolve conversation" button. I also enumerated your points.

src/scripts/medgen_conflicts_add_xrefs.py
```diff
@@ -0,0 +1,79 @@
...
+    # - recommended action: Nothing. Removals done in previous step. Good xref ref'ed in sheet still exists.
+    pass
+
+    # Sheet 5/5: MondoXrefs_NotinMedGen
+    # - summary: These terms do not exist in MedGen.
+    # - recommended action: Nothing. Removals done in previous step. No other actions from this sheet.
+    pass
+
+    # Output: List of CUIs of bad xrefs
+    out_df = pd.DataFrame([{'mondo_id': 'ID', 'umls_id': 'A oboInOwl:hasDbXref'}] + rows)
```
matentzn, line 62: How does the `rows` variable get populated?

src/scripts/medgen_conflicts_add_xrefs.py
```diff
@@ -0,0 +1,79 @@
...
+    # Sheet 2/5: WrongCUIxref_withCurrCUI
+    # - summary: MedGen team has suggested us better mappings to use on these Mondo IDs.
+    # - recommended action:
+    df = book.parse("WrongCUIxref_withCurrCUI")
+    pass
+
+    # Sheet 3/5: 1MondoID_to1CUI
+    # - summary: MedGen has identified better Mondo IDs to map to these CUIs. 
+    # - recommended action:
+    df = book.parse(">1MondoID_to1CUI")
```
matentzn, line 48: Doesnt this overwrite the provious `df=`, row 42?

src/ontology/mondo.Makefile
matentzn: Re 1 and 2: I guess we should have clarified that "remove xrefs" means: if the xref appears in an axiom annotation (in wavy or squared brackets) remove the axiom annotation with the xref, not the whole row in the OBO file, if the xref appears in a term level `xref: ...` statement, remove the whole row. 

Thanks for the clarifications on 4, I ready your comments. I trust you to basically have an idea of what you are doing now, and that removing the 2/5 means removing _wrong_ xrefs. I don't have enough time to ensure that these 2/5 are considered wrong by Megan; I will trust it is. So I am fine as you are doing it then!

src/scripts/medgen_conflicts_add_xrefs.py
joeflack4: I haven't gotten to this yet. There is a section in the script with some instructions for each worksheet. I'm not even sure if this is the implementation that I'll pick, but it'll probably work. There's a 'todo' that describes my thoughts. I'll probably do a `.to_dict()`, do a transform, then add to `rows`.

src/scripts/medgen_conflicts_add_xrefs.py
joeflack4: This I think is a minor detail. When I load multiple temporary dataframes like this, sometimes I overwrite them further down, and sometimes I give them different names, which helps if I want to have multiple of them open in the debugger.  In this case (and in the other xref removal script), I just overwrite the variable.

src/scripts/bad_medgen_xrefs_grep_command.py
joeflack4: I think Nico has had a chance to review this, and I don't think there is anything of consequence here. Instructions and  implementation have changed slightly.

src/ontology/mondo.Makefile
joeflack4: @matentzn Today I made fixes / updates regarding topics 1-3 above.

**Rebased after your #6573**
Regarding topic 1 above, this might have been an oversight/confusion on my part. I've left the `UMLS:CN` removals to your PR. I reset my changes to `mondo-edit.obo` and rebased.

**MEDGEN removals: undid non-xrefs**
Regarding topic 2 above, that feels a bit dumb on my part to have removed every line if "MEDGEN:" simply appeared in it; I had an inkling you might have meant to remove MEDGEN xrefs lines and source annotations separately.

See code snippet below for how this is now implemented. For MEDGEN xrefs, the whole lines are removed. I then was going to implement the logic for situations like `source="MEDGEN:`. However, there was only 1 such case and it was already being removed because it was one of Megan's conflicts.

**Python reimplementation**
Because of the nature of topic 3 above (exceptions), I re-implemented the whole make/ShellScript method of doing these removals (`grep -v`) to pure Python. Also seems lot easier / more straightforward in Python, and greatly increases accuracy.

https://github.com/monarch-initiative/mondo/blob/9fd9171bad755556d93b1e34ee32e46d9211c478/src/scripts/medgen_conflicts_removals.py#L82-L94

<details><summary>Updated metrics table</summary>
<p>

These results look good. `n_single_cui_removed_from_xrefs_n_times__1` is 2,818 this time; was previously 2,819. I' guessing this is because I rebased from your recent PR, which must have removed 1 of these already.

I didn't add an entry to this table for `n_medgen_source_evidence_xref_annotations_removed`. But there were 0 anyway. All of these ones that would have been removed this way ended up being removed as a consequence of Megan's UMLS: xref removals anyway.

|metric_name                               |val |
|------------------------------------------|----|
|n_lines_removed                           |3086|
|n_xrefs_removed                           |3086|
|n_non_xrefs_removals                      |0   |
|n_medgen_removals                         |28  |
|n_umls_removals                           |3058|
|n_single_cui_removed_from_xrefs_n_times__1|2818|
|n_single_cui_removed_from_xrefs_n_times__2|106 |
|n_single_cui_removed_from_xrefs_n_times__3|8   |
|n_single_cui_removed_from_xrefs_n_times__4|1   |


</p>
</details> 

**Can increase specificity of removals, if you want**
You said that it was alright for any CUIs / UMLS IDs Megan mentioned in her conflicts file, that I can remove them even if they end up getting removed in multiple places. However, she did mention specific classes where this was a problem. Now that I have changed `grep -v` --> pure Python, I can change it to remove only the xrefs on these classes. If I do this, I can re-run the analysis table above to show that all CUIs were only removed from 1 xref.

**Removal exceptions MONDO:preferredExternal & https://orcid.org**
You said:
> If a row contains one of the following strings, do not remove it: MONDO:preferredExternal, https://orcid.org (indicates someone has manually removed it)

I'm only doing 3 kinds of deletes. Here they are, as well as a count of how many times I found either of these strings:
1. source="MEDGEN: - 0
2. xref: MEDGEN: - 0
3. Medgan's "xref: UMLS:" conflicts - 23

So, when it comes to my removals of anything MEDGEN/UMLS related, the only time I'm finding these strings (in the xref axiom annotations) is for cases where Megan specifically asked me to remove an xref.

However, I take it that you do not actually want me to override her, so (as can be seen in my Python snippet above), I did not implement this exception logic for those conflicts.

Also, just to clarify--what do you mean by "it" in "indicates someone has manually removed it" in your quote above? I think it means "UMLS or MEDGEN source= evidence axiom annotations". Is that correct?

<details><summary>Examples of such rows I found</summary>
<p>

1. `xref: UMLS:C1970879 {source="MONDO:equivalentTo", source="MONDO:ncbi_mim2gene_medline", source="MONDO:preferredExternal", source="OMIM:300636"}`
2. `is_a: MONDO:0000762 {source="DC-OMIM:300869", source="https://orcid.org/0000-0002-4142-7153"} ! syndrome caused by partial chromosomal duplication`

</p>
</details>

src/scripts/medgen_conflicts_removals.py
```diff
@@ -0,0 +1,117 @@
+"""MedGen conflicts: Removals
```
joeflack4, line 1: FYI: New Python script for removals

src/ontology/mondo.Makefile
matentzn: This looks more or less like I would expect now. I have gathered the 900 more risky removals in this spreadsheet: https://docs.google.com/spreadsheets/d/1y86gfXydQE0wOfQzddohN-TWWHUJ11QH7IJg9n_PA5c/edit#gid=0

I need to be convinced that all of these are really wrong. Can you tell me, according to the data (I know you told me before, but I needed to ask the question again):

**Does Megan consider all of these wrong one way or another?**

I am still not sure if we should remove UMLS xrefs only from specific Mondo IDs, or from all of them. Maybe now it is time to organise a call with Megan?

Let me know when you have the ROBOT template together, this will give me some important clues to advice you better. Thanks!!

joeflack4: Glad this is better. Yes, Megan considers all of these problematic and should be removed some way in Mondo. As you understand, for many of them she is only pointing out that they need to be removed on a single class. But I will continue removing all of their xrefs for now.

Perhaps the best way to see what is recommended for each sheet is in my code comments:
https://github.com/monarch-initiative/mondo/blob/9fd9171bad755556d93b1e34ee32e46d9211c478/src/scripts/medgen_conflicts_removals.py#L31-L61

This is a distillation of what I've written on each sheet in my personal notes, which is basically a distillation of what Megan has written in [README_July2023_CUIReports_FromMedGentoMondo.txt](https://github.com/monarch-initiative/mondo-ingest/files/12029711/README_July2023_CUIReports_FromMedGentoMondo.txt) ([her original comment](https://github.com/monarch-initiative/mondo-ingest/issues/273#issuecomment-1632774012)).

Feel free to schedule a call if you want. Personally what she's asking seems pretty clear to me; I don't feel the need for any clarification. But perhaps you feel that need, or maybe we could evaluate reasoning / ask other questions that you may have. Below is my schedule for the next two weeks.

<details><summary>Joe's availability</summary>
<p>

Screen Shot 2023-08-27 at 4 37 50 PM

Screen Shot 2023-08-27 at 4 37 59 PM


</p>
</details>

src/scripts/medgen_conflicts_add_xrefs.py
```diff
@@ -0,0 +1,94 @@
...
+TEMPLATES_DIR = PROJECT_DIR / "src" /  "templates"
+TMP_DIR = ONTOLOGY_DIR / "tmp"
+INPUT_FILE = str(TMP_DIR / "July2023_CUIReports_FromMedGentoMondo.xlsx")
+OUTPUT_FILE = str(TEMPLATES_DIR / "ROBOT_addMedGen_fromConflictResolution.tsv")
+
+
+def _rows_from_common_sheet_format(df: pd.DataFrame) -> List[Dict]:
+    """From worksheets having same common format, get prefixed xrefs for the namespaces we're looking to cover"""
+    df['xref_id'] = df['xref_id'].apply(
+        lambda x: f'MEDGENCUI:{x}' if x.startswith('CN')  # "CUI Novel"
```
joeflack4, line 29: ## Xref additions: From Megan's conflicts file
@matentzn I parsed through Megan's conflicts file and there are 3 sheets where she thinks xrefs should be replaced. I did this as separate steps: (i) the removals, as you have reviewed previously, and (ii) additions here as a robot template.

I've highlighted here the algorithm for prefixing each of these ids. I've used the same algorithm as the one we're using in the MedGen ingest:

https://github.com/monarch-initiative/medgen/blob/main/src/medgen2obo.pl#L161-L172

It chooses a different prefix based on if its a CN, a CUI, or a UID. And also as we did in the MedGen ingest, if it's a CUI, we add 2 xrefs: one under the `UMLS` namespace, and another under the `MEDGENCUI` namespace.

joeflack4: ## Next & final step: xref additions from MedGen mappings
As I wrote in [the OP](https://github.com/monarch-initiative/mondo/pull/6560#issue-1850743276), my next step is: 
> 3. Add new xrefs generated from [MedGen ingest](https://github.com/monarch-initiative/medgen): which will mainly come from the FTP file MedGenIDMappings.txt, but will probably require that whole ingest to be working to generate.

This depends on the medgen ingest working properly, and it's in pretty good shape I think, so I don't anticipate much longer until this is done too.

joeflack4: ## Run command to add these additions?
Is this something you'd like me to do, or would you like to do it, or will this automatically be done via some pipeline?

I'm not familiar with the exact command to do this; maybe `--merge-before` ([docs](http://robot.obolibrary.org/template#merging))?
Would this be run on `mondo-edit.obo`?

src/ontology/mondo.Makefile
```diff
@@ -268,8 +269,42 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	python ../scripts/medgen_conflicts_removals.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -I mondo-edit.obo -o $@
+
+# - MedGen conflicts: adds
+.PHONY: address-medgen-conflicts-aug2023-adds
+address-medgen-conflicts-aug2023-adds: $(TEMPLATES_DIR)/ROBOT_addMedGen_fromConflictResolution.tsv $(TEMPLATES_DIR)/ROBOT_addMedGen_fromIngest.tsv
+
+$(TEMPLATES_DIR)/ROBOT_addMedGen_fromConflictResolution.tsv: tmp/July2023_CUIReports_FromMedGentoMondo.xlsx
+	python ../scripts/medgen_conflicts_add_xrefs.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -o $@
 
+$(TEMPLATES_DIR)/ROBOT_addMedGen_fromIngest.tsv:
```
joeflack4, line 306: ## Next xref overlap analysis
You asked for another report that you called a "static analysis", where we report CUIs/CNs/UIDs that are `in_mondo`, `in_medgen`, or `both`. To do that, these removals/additions [need to be applied](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1306730253). Then you recommended 1 way to get my fresh input for that `mondo.sssom.tsv` was via [this `mondo-ingest` goal](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/mondo-ingest.Makefile#L341).

I'll leave it to you as to whether you want to do this (a) ahead of [Next & final step: xref additions from MedGen mappings](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1306729478), or (b) after that step.

joeflack4, line 306: ## Next & final step: xref additions from MedGen mappings
As I wrote in [the OP](https://github.com/monarch-initiative/mondo/pull/6560#issue-1850743276), my next step is: 
> 3. Add new xrefs generated from [MedGen ingest](https://github.com/monarch-initiative/medgen): which will mainly come from the FTP file MedGenIDMappings.txt, but will probably require that whole ingest to be working to generate.

This depends on the medgen ingest working properly, and it's in pretty good shape I think, so I don't anticipate much longer until this is done too.

joeflack4: @matentzn As long as everything looks good to you, this is ready to merge. Everything on my end [per the instructions from your comment in the issue](https://github.com/monarch-initiative/mondo-ingest/issues/273#issuecomment-1653431573) have been done. The "Next xref overlap analysis" above is the only remaining related immediate follow-up on my todo list.

joeflack4: @joeflack4 I've now added `ROBOT_addMedGen_fromIngest.tsv`. This is the same file that is called `medgen-xrefs.robot.template.tsv` in the MedGen ingest.

Why do I rename it here? Just to be consistent with what's currently in the `templates/` dir:
```
ROBOT-partial-chromosome.tsv
ROBOT-partial-duplication.tsv
ROBOT_AML_NCIT_for_MONDO_2021-09-14.tsv
ROBOT_AddClinGenPref.tsv
ROBOT_CreateNewTerm.tsv
ROBOT_MergeTag.tsv
ROBOT_ObsoleteTag.tsv
ROBOT_addCongenitalSubclass-Mondo4069.tsv
ROBOT_addSubClassOf_Mondo3680.tsv

# And now also: 
ROBOT_addMedGen_fromConflictResolution.tsv
ROBOT_addMedGen_fromIngest.tsv
```

src/ontology/mondo.Makefile
joeflack4: Nico says don't worry about it. Global scale in mondo for this kind of thing; outdated refs. manually

src/scripts/medgen_conflicts_removals.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # Convert to strings and sort
+    bad_xref_ids: List[str] = sorted([str(x) for x in bad_xref_ids])
+    return bad_xref_ids
+
+
+def run(input_conflicts: str = INPUT_CONFLICTS, input_obo: str = INPUT_OBO, output: str = OUTPUT_FILE):
+    """Run"""
+    # Vars
+    exceptions = ['MONDO:preferredExternal', 'https://orcid.org']
```
joeflack4, line 75: ##  Removal exceptions `MONDO:preferredExternal` & `https://orcid.org`
@matentzn Update 8/31: We just met today and I'm re-looking at the comments we didn't look over. This is a copy/paste of my [original comment](https://github.com/monarch-initiative/mondo/pull/6560#discussion_r1306588345).

The main point:
Is that you told me not to remove cases where I see these strings. But the only time they come up is when Megan specifically is saying that I should remove the xref. I'm assuming you want me to follow Megan's instructions in these cases.

---

You said:
> If a row contains one of the following strings, do not remove it: MONDO:preferredExternal, https://orcid.org (indicates someone has manually removed it)

I'm only doing 3 kinds of deletes. Here they are, as well as a count of how many times I found either of these strings:
1. source="MEDGEN: - 0
2. xref: MEDGEN: - 0
3. Medgan's "xref: UMLS:" conflicts - 23

So, when it comes to my removals of anything MEDGEN/UMLS related, the only time I'm finding these strings (in the xref axiom annotations) is for cases where Megan specifically asked me to remove an xref.

However, I take it that you do not actually want me to override her, so (as can be seen in my Python snippet above), I did not implement this exception logic for those conflicts.

Also, just to clarify--what do you mean by "it" in "indicates someone has manually removed it" in your quote above? I think it means "UMLS or MEDGEN source= evidence axiom annotations". Is that correct?

src/ontology/mondo.Makefile
joeflack4: @matentzn Marking this one resolved. I think the only important one left here is my last comment. You were just feeling unsure of some of these removals and maybe wanted to meet with Megan. I explained the removals above; I don't really see a need to meet with her at this time.

src/scripts/medgen_conflicts_add_xrefs.py
```diff
@@ -0,0 +1,100 @@
+"""MedGen conflicts: Add new xrefs
```
joeflack4, line 1: Hey @nicolevasilevsky. The PR requires you to review because you are a code owner. I think Nico has this one covered but if you have time and want to glance, feel free!

From `.github/CODEOWNERS`:

```
* @nicolevasilevsky
*.sparql @matentzn
*.py      @joeflack4
```

@matentzn @nicolevasilevsky To me, this looks like Nicole will be required on every single PR. Is this what we want?

src/scripts/medgen_conflicts_removals.py
```diff
@@ -0,0 +1,146 @@
+"""MedGen conflicts: Removals
```
joeflack4, line 1: ## Only removing xrefs on specific Mondo IDs
@matentzn Per your instruction, I followed Megan's original directive and removed xrefs w/ on specific Mondo IDs.

I re-ran this diff analysis, and it looks good. Only xrefs are removed, and unlike before, any xrefs removed are only being removed on 1 term.


|metric_name                               |val |
|------------------------------------------|----|
|n_lines_removed                           |2669|
|n_xrefs_removed                           |2669|
|n_non_xrefs_removals                      |0   |
|n_medgen_removals                         |0   |
|n_umls_removals                           |2669|
|n_single_cui_removed_from_xrefs_n_times__1|2669|

matentzn: great, i will take it from here! Thank you!

src/scripts/medgen_conflicts_add_xrefs.py
joeflack4: ## `mondo-edit.obo` conflict
@matentzn When you get around to reviewing this, there may be a conflict with `mondo-edit.obo`. I had been rebasing this branch, resetting that file, and then re-running `sh run.sh make address-medgen-conflicts-aug2023-deletes`. When you get to reviewing this, let me know and I can do that again for you, or you can do what I just described or some variation of it.

src/ontology/mondo.Makefile
```diff
@@ -268,8 +269,45 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	mkdir -p tmp && wget "https://github.com/monarch-initiative/mondo-ingest/files/12029712/July2023_CUIReports_FromMedGentoMondo.xlsx" -O $@
+
+# - MedGen conflicts: analysis
+tmp/mondo-edit.obo.tmp.diff: mondo-edit.obo.tmp
+	-diff mondo-edit.obo mondo-edit.obo.tmp > $@
+
+tmp/report-qc-medgen-conflicts-update-diff.tsv: tmp/mondo-edit.obo.tmp.diff
+	python ../scripts/medgen_conflicts_removals_diff_analysis.py -i tmp/mondo-edit.obo.tmp.diff -o $@
+
+tmp/report-qc-medgen-conflicts-adds-diff.tsv: $(TEMPLATES_DIR)/ROBOT_addMedGen_fromConflictResolution.tsv $(TEMPLATES_DIR)/ROBOT_addMedGen_fromIngest.tsv
```
joeflack4, line 291: ## Megan's 278 additions absent from ingest
@matentzn Added script w/ goal to analyze any xref addition discrepancies between what is coming from our ingest and what's coming from curated conflicts, with the main purpose be to see if there's anything recommended from curation that is not coming in through the ingest.

The result is 35,567 additions coming in through the ingest that were not among Megan's recommended replacements. A high number is to be expected as she was just curating some bad Mondo xrefs and making recommended replacements.

However, Megan made 278 recommendations that did not appear in the ingest. I only checked 1 case, and it is found in the FTP's `MedGenIDMappings.txt`; it was just not mapped explicitly to Mondo. I expect the other terms will also appear. My guess here is that Megan's ad hoc process to create these recommendations for us involved a lot of manual work and was decoupled from whatever process generates `MedGenIDMappings.txt`. Perhaps she derived these mappings via proxy (in the one case I looked at, the term was mapped to an OMIM term).

**Suggested action**
I suggest we bring this list to Megan and, ideally, for each term, evaluate whether (a) they should update `MedGenIDMappings.txt` w/ the mapping, or (b) see if she made the recommendation in error.

[report-qc-medgen-conflicts-adds-diff.tsv.zip](https://github.com/monarch-initiative/mondo/files/12564166/report-qc-medgen-conflicts-adds-diff.tsv.zip)

src/ontology/mondo.Makefile
```diff
@@ -268,8 +269,45 @@ reports/%-rare-diseases.tsv: $(ONT)-base.owl reports/%-rare-diseases.txt
...
+	python ../scripts/medgen_conflicts_removals.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -I mondo-edit.obo -o $@
+
+# - MedGen conflicts: adds
+.PHONY: address-medgen-conflicts-aug2023-adds
+address-medgen-conflicts-aug2023-adds: tmp/report-qc-medgen-conflicts-adds-diff.tsv
+
+$(TEMPLATES_DIR)/ROBOT_addMedGen_fromConflictResolution.tsv: tmp/July2023_CUIReports_FromMedGentoMondo.xlsx
+	python ../scripts/medgen_conflicts_add_xrefs.py -i tmp/July2023_CUIReports_FromMedGentoMondo.xlsx -o $@
 
+$(TEMPLATES_DIR)/ROBOT_addMedGen_fromIngest.tsv:
```
joeflack4, line 309: ## [Robot template: UID sources #13](https://github.com/monarch-initiative/medgen/pull/13#pullrequestreview-1618699434)
Updated MedGen ingest repo with [this PR](https://github.com/monarch-initiative/medgen/pull/13#pullrequestreview-1618699434), which, for the robot template it is generating, added a `source_id` column so that for any UIDs we are ingesting, we include the source UMLS CUI they are mapped to.

Uploaded the new file to [releases](https://github.com/monarch-initiative/medgen/releases/tag/2023-09-03), then ran `make ../templates/ROBOT_addMedGen_fromIngest.tsv -B` to update the template in this repository.

src/scripts/medgen_conflicts_add_xrefs.py
matentzn: Thank you, I will take it from here!

src/ontology/mondo.Makefile
matentzn: @joeflack4 can you turn this into an issue link it to this comment and resolve it here. Since these are "false negatives" xrefs (i.e. missing mappings) and not "false positive" mappings (wrong mappings), lets deal with these in a second step.

src/scripts/medgen_conflicts_removals.py
matentzn: Do you know without much work how often the case happens that a removed xref has an orcid? Dont do a lot work, I can see it from the diff as well.

joeflack4: If it's useful info, in the 'details' above, I wrote:
> I'm only doing 3 kinds of deletes. Here they are, as well as a count of how many times I found either of these strings:
> source="MEDGEN: - 0
> xref: MEDGEN: - 0
> Medgan's "xref: UMLS:" conflicts - 23

As for _orcid_ specifically, looks like just 2 cases:
`git diff master src/ontology/mondo-edit.obo | grep orcid | grep -- -xref`
```
-xref: UMLS:C1512224 {source="DOID:5310", source="MONDO:equivalentTo", source="MONDO:notFoundInDiseaseSubset", source="https://orcid.org/0000-0001-5493-2602"}
-xref: UMLS:C4755316 {source="MONDO:equivalentTo", source="https://orcid.org/0000-0001-5493-2602"}
```

src/ontology/mondo.Makefile
joeflack4: Sounds good! Here's the issue created: https://github.com/monarch-initiative/mondo-ingest/issues/366

Also, I went ahead and let Megan know about this. You can also see that and her response: [here](https://github.com/monarch-initiative/mondo-ingest/issues/273#issuecomment-1712658861)

src/ontology/mondo.Makefile
matentzn: I have removed all these MEDGEN mappings for now - this PR is now focussing on the UMLS conflicts

matentzn:
@sabrinatoro @twhetzel due to its size and the huge time our stakeholders have been waiting for this, we should prioritise merging this now that the release is done.

You do not need to concern yourself with any of the code - reviewing this will take a lot longer than what it is worth, and this is a one off PR for fixing a lot of imprecise UMLS mappings. 

To review this you could go to the main branch, run git pull and run: `git checkout bugfix-medgen-conflicts -- mondo-edit.obo`. And review in GitHub desktop. The changes look very sane to me and I am inclined to trust @kanems with the task of cleaning up the UMLS references moving forward.

src/scripts/medgen_conflicts_removals.py
joeflack4, line 1: ## Some things not removed? @matentzn 
@matentzn This still a problem? Could you double check my understanding?

<details><summary>Details</summary>
<p>

[_Nico wrote](https://github.com/monarch-initiative/mondo/commit/cef1663f5aa4ba27eb10be5b2393735f1f02e8cd)_:
> @joeflack4 Why were they these not removed by the remove step?

@matentzn It looks like you corrected a problem [in this commit](https://github.com/monarch-initiative/mondo/commit/cef1663f5aa4ba27eb10be5b2393735f1f02e8cd), but I'm not sure if the problem remains and there is something I need to solve?

If it still needs my fixing, could you double check my understanding and clarify if the following is what's happening?
1. There's a robot template that shows things that should be removed from `mondo-edit.obo`
2. After running pipeline, they're still not removed.

</p>
</details>

sabrinatoro:
@matentzn 
The changes in mondo-edit.obo make sense. And I also trust the review by @kanems 
Quick question: did I see correctly that some UMLS x-ref were removed? (I think it makes complete sense, I just want to confirm). Otherwise, I approve the mondo-edit.obo part of this PR.

sabrinatoro (approved):

twhetzel:
Working through these changes and I do see that for Mondo IDs in "July2023_CUIReports_FromMedGentoMondo.xlsx" the `database_cross_refererence` and `exactMatch` annotations are removed for these UMLS ref_types. 

Given that in a prior PR, UMLS xrefs with "CN" were removed, was that only for "equivalentTo"? I saw one with "equivalentObsolete" and wondering if that is expected (screenshot attached from mondo-edit.obo from this PR).
Screenshot 2023-12-13 at 5 09 54 PM

src/scripts/medgen_conflicts_removals.py
matentzn: Sorry, I am creating more work than its worth here. Lets leave it; something in the "conflict removal" step is a bit incomplete, probably because it focusses on combinations of MONDO ids with Xrefs.

In this commit, `UMLS:C1858991` was referenced from a another class (not in the MedGen table) and therefore caused a clash after the fixed xrefs where added in.

Lets not worry about this; it looks mostly good, I would rather we have a decision about how to curate UMLS moving forward!

matentzn:
@sabrinatoro 

> Quick question: did I see correctly that some UMLS x-ref were removed? (I think it makes complete sense, I just want to confirm). Otherwise, I approve the mondo-edit.obo part of this PR.

Yes, this is the point of this PR: removing "wrong" xrefs and replacing them by better ones (supplied by @kanems and MedGen)

@twhetzel 

> Given that in a prior PR, UMLS xrefs with "CN" were removed, was that only for "equivalentTo"? I saw one with "equivalentObsolete" and wondering if that is expected (screenshot attached from mondo-edit.obo from this PR).

Very good observation, very thorough. I will make a new PR which will just remove _all_ CN xrefs, which are basically meaningless for us. They will be replaced by MEDGEN identifiers instead. I have reduced the scope of this PR to 

1. Remove wrong UMLS reference (wrong according to @kanems)
2. Add right UMLS references (right according to @kanems)

We will have a standardised system for this in the future, for all resources curating xrefs. For now, we just fix a lot of xrefs at once, and then have something more gradual moving forward (basically, whenever MedGen reports a better xref, we implement it).

twhetzel:
> I will make a new PR which will just remove all CN xrefs
Is there another commit to this PR that does this or did this mean another ticket would be created to handle these other "CN" UMLS xrefs?

twhetzel (approved):
I have an existing question inline about the future work to remove _all_ CN type UMLS xrefs, but do not want to hold up the PR until then since I think that is intended to be addressed in a separate, future ticket.

PR #6573, opened 2023-08-17
Remove places where UMLS:CN was used as evidence

matentzn:
There are 59 places where UMLS:CN was used as evidence. CN are temporary CUIs minted by Medgen, and we have no way to track them - so we should not use them as evidence. They are not in UMLS.


matentzn:
cc @joeflack4 FYI

joeflack4:
Hmm I thought UIDs were more transient than CNs (CUI Novels), but you are probably right.

Just to clarify, what precisely am I looking to delete?

I'm thinking it is this pattern?
`xref: XYZ:123 {source="UMLS:CN12345"}`

Or is there an IAO or some other predicate that has some meaning like "has_evidence"?

matentzn:
? This is a pull request that already deals with the issue! No need for you to do anything!

joeflack4:
Yeah maybe I should have just deleted that question; I had wrote later:
> edit: Oh, this is a PR, not an issue, lol.I see what you have done.

sabrinatoro (approved):

PR #6590, opened 2023-08-24
Weekly automated mappings update

hrshdhgd:
Update mappings in Mondo from external sources.


matentzn:
Some issues:

1. Do not include GARD mappings
2. Run normaliser

PR #6575, opened 2023-08-17
OMIA Non-Human Animal Diseases

katiermullen:
Address issue #5585

PR #6601, opened 2023-08-30
added omia terms

sabrinatoro:
#6575 is definitely problematic. I am re-creating the PR (trying first by adding all the terms before adding in smaller chuncks)


sabrinatoro:
Resolving conflicts leads to weird diffs. Closing and restarting.

PR #6643, opened 2023-09-12
release 230912

sabrinatoro:


sabrinatoro:
@nicolevasilevsky and @twhetzel 
This is the first step of the release. 
I noted that there are a lot of changed definitions: 508. (see src/ontology/reports/mondo_release_diff.md)
I am reviewing them. it looks like it is mostly typos, but I don't remember updated so many typos. 
@nicolevasilevsky is it something you remember doing? 

DO NOT MERGE until we are sure that things are ok with these high number of changed definitions.

sabrinatoro:
maybe some definitions were Americanized? I see some "characterized" instead of "characterise" or "disk" instead of "disc"?

I also see ".." updated to "."

sabrinatoro:
I have convinced myself (with Nicole's confirmation) that these changes were part of the British standardization work.

PR #6648, opened 2023-09-15
Create Branch review table

matentzn:
@cmungall Here is where SPARQL really suxx. If you can provide me a SQL query to deal with this specific issue, I will start switching to OAK for table reports.

UPDATE:
---------
Using `oaklib`:

```
reports/mondo-curation-branch-review.tsv:
	python src/scripts/branch_review.py create-review-table -o $@ -f obsoletion_terms.tsv -B branch_ids.tsv 
```

OR
```
reports/mondo-curation-branch-review.tsv:
	python src/scripts/branch_review.py create-review-table -o reports.tsv -f obsoletion_terms.tsv -b MONDO:0005151
```

Pseudo-code
```
all_branches = [....list of all branches ...]
branch_dict = { branch_id: branch_id.get_descendants() }

def create_review_table(branch_id, obsoletion_candidate_ids):
    affected_mondo_ids = id.get_direct_children() for id in obsoletion_candidate_ids #these are the classes that we need to review.
    
    for affected_id in affected_mondo_ids:
        # FOR THIS ANALYSIS ONLY CONSIDER MONDO IDS, NOT BFO ETC AS PARENTS
        parents = affected_id.get_direct_parents()
        parents.remove(obsoletion_candidate_ids)
        parents_inside_branch = intersection(parents, branch_dict[branch_id])
        parents_outside_branch = parents.remove(parents_inside_branch)
        ancestors = affected_id.get_ancestors()
        ancestors_inside_branch = intersection(ancestors, branch_dict[branch_id])
        ancestors_outside_branch = parents.remove(ancestors_inside_branch)
        
        if no ancestors: affected_status = ORPHANED
        elif no ancestors_inside_branch = LEFT_THIS_BRANCH
        elif ???? = LEFT_OTHER_BRANCH <- I DONT KNOW HOW TO WRITE THIS BUT LETS DO THIS AFTER EVERYTHING ELSE. FOR NOW JUST: IF ANY(ancestor in all_branches) 
        elif parents_inside_branch = RETAINS_PARENT

        else : RETAINS_ANCESTOR
        
        data = [affected_id, labels(parents_inside_branch), labels(parents_outside_branch), labels(ancestors_inside_branch), labels(ancestors_outside_branch), affected_status] 

def labels(curies):
    return [ f"{c.label} ({c})" for c in curies ]
```


src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+                    get_children_from_relations(
+                        curie=id,
+                        relationships=obsoletion_candidates_branch_relationships,
+                    )
+                    for id in obsoletion_candidates_branch
+                )
+            )
+
+            if len(affected_mondo_ids) > 0:
+                for obsoletion_candidate in affected_mondo_ids:
```
matentzn, line 118:
```suggestion
                for affected_id in affected_mondo_ids:
```

Best rename this variable as it makes the rest of the code confusing to read.

src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+                    parent_of_obsolete_candidate = get_parent_from_relations(
```
matentzn, line 119:
```suggestion
                    parents_of_affected = get_parent_from_relations(
```

src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+                        obsoletion_candidate,
+                        obsoletion_candidates_relationships,
+                    )
+
+                    filtered_parent_of_obsolete_candidate = [
+                        x
+                        for x in parent_of_obsolete_candidate
+                        if x not in affected_mondo_ids
+                    ]
+                    filtered_parent_of_obsolete_candidate = [
```
matentzn, line 129: Same variAble name as above?

src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+                for obsoletion_candidate in affected_mondo_ids:
+                    parent_of_obsolete_candidate = get_parent_from_relations(
+                        obsoletion_candidate,
+                        obsoletion_candidates_relationships,
+                    )
+
+                    filtered_parent_of_obsolete_candidate = [
+                        x
+                        for x in parent_of_obsolete_candidate
+                        if x not in affected_mondo_ids
```
matentzn, line 127: I don't think this filter is needed

src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+                        if i not in obsoletion_candidates
+                    ]
+
+                    ancestors_of_obsolete_candidate = list(
+                        OI.ancestors([obsoletion_candidate], predicates=[IS_A, PART_OF])
+                    )
+                    filtered_ancestor_of_obsolete_candidate = [
+                        x
+                        for x in ancestors_of_obsolete_candidate
+                        if x not in affected_mondo_ids and x.startswith("MONDO:")
```
matentzn, line 141: Also not innobsoletion candidates

src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+                    ]
+
+                    parents_inside_branch = list(
+                        set(filtered_parent_of_obsolete_candidate)
+                        & set(branch_descendants_dict[branch_id])
+                    )
+                    parents_inside_branch = [
+                        i
+                        for i in parents_inside_branch
+                        if i not in filtered_parent_of_obsolete_candidate
```
matentzn, line 151: Filter Not needed

src/scripts/branch_review.py
```diff
@@ -0,0 +1,239 @@
...
+
+                    parents_outside_branch = [
+                        x
+                        for x in filtered_parent_of_obsolete_candidate
+                        if x not in parents_inside_branch
+                    ]
+                    parents_outside_branch = [
+                        i
+                        for i in parents_outside_branch
+                        if i not in filtered_parent_of_obsolete_candidate
```
matentzn, line 162: Not needed

matentzn (commented):
I made some, but not all comments, as it's hard to read the code now that the affected entity variable has been introduced. 

Can you rename all the variables to say affected_entity instead of obsoletion candidate where it makes sense? And also, instead of "filtered" say how, like "non_obsolete_parents". I will Check again tomorrow

src/scripts/branch_review.py
hrshdhgd: This is needed for lines below. I think my understanding is faulty here and may need clarification.

hrshdhgd:
I committed just refactors to variable names and some `list` <=> `set`

sabrinatoro:
What is needed: 
- report **direct children** of the obsoletion candidates
- you can focus on one branch at a time. Branches can be found [here](https://docs.google.com/spreadsheets/d/1xTfYCOCGKx_svNT8PirgGDiCINV7IDqzhBk76DNe150/edit#gid=0). We are starting with MONDO:0005151 - endocrine system disorder
- since obsoletion candidates can be at different level within a branch, we need to know whether the direct parents (and ancestors) are obsoletion candidates 
        - option 1: remove all parent/ancestor that are obsoletion candidates from the report
        - option 2: rename all obsoletion candidates to include "TO BE OBSOLETED" at the beginning of their label. (this should be done for the whole ontology, so we know that a term might actually loose more than one parent)
- Please find examples [here](https://docs.google.com/spreadsheets/d/1xTfYCOCGKx_svNT8PirgGDiCINV7IDqzhBk76DNe150/edit#gid=189872967), included column name suggestions.

sabrinatoro:
Reviewing [this spreadsheet]:

- the obsoletion candidate (colE ) and their children (colB) are not in the branch review (colA)
For example: obsoletion candidate MONDO:0013742 (colE) is not in the MONDO:0005151 (colA) branch.
BUT all the children of the obsoletion candidate (ColB) are reported (e.g. there are 2 children for obsoletion candidate MONDO:0013742)

- disregarding the "in the branch" or "out of the branch" (since this is incorrect because of what I mentioned above), some direct parents are missing. For example: 
MONDO:0008964, should have a total of 4 direct parents:  MONDO:0000824 ('congenital diarrhea'); MONDO:0000249 ('secretory diarrhea'), MONDO:0015178 (OBSOLETION CANDIDATE); MONDO:0000824 ('congenital diarrhea')

Note that the first example I look at, the obsoletion candidate MONDO:0008347 was already obsoleted...  I am assuming this is not going to pose any problem since we will run this after each obsoletion round. But I still want to mention it to make sure.

twhetzel:
@sabrinatoro in the second bullet point above:
```
- disregarding the "in the branch" or "out of the branch" (since this is incorrect because of what I mentioned above), some direct parents are missing. For example:
MONDO:0008964, should have a total of 4 direct parents: MONDO:0000824 ('congenital diarrhea'); MONDO:0000249 ('secretory diarrhea'), MONDO:0015178 (OBSOLETION CANDIDATE); MONDO:0000824 ('congenital diarrhea')
```
the term `MONDO:0000824 ('congenital diarrhea')` is duplicated

hrshdhgd:
> * MONDO:0010720
>   - direct parent in the spreadsheet: MONDO:0019052->inborn errors of metabolism

MONDO:0003847 is an ancestor of MONDO:0010720 outside the branch and denoted so in the spreadsheet. [I'm cross-validating using OLS](https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_0003847). Same for the other bullet. My guess is the OLS version is old(?)

sabrinatoro:
AffectedStatus: rules
- if there is **at least 1 (not obsoletion candidate) term in Col F** (=other direct parents in branch) (whatever parents are in Col G -”other direct parents NOT in branch”) **--> STAYS IN THE BRANCH**
- if there is **NO (not obsoletion candidate) term in Col F** (=other direct parents in branch)  **AND** **at least 1 (not obsoletion candidate) term in Col G** (=other direct parents NOT in branch)--> **LEAVES THE BRANCH**
- if there is **NO (not obsoletion candidate) term in F** **AND** **NO (not obsoletion candidate) term in G** --> **ORPHAN**

@hrshdhgd

src/ontology/mondo.Makefile
```diff
@@ -908,6 +908,22 @@ config/exclusion_reasons.tsv:
 
 all: config/exclusion_reasons.tsv
 
+##################################
+#### SPARQL Curation pipeline ####
+##################################
+
+tmp/mondo-curation-inferred.ttl:
+	$(ROBOT) convert -I $(OBOBASE)/mondo/mondo-base.owl -f ttl -o $@
```
matentzn, line 916: Comment the above two lines out to ensure they are not run.

matentzn: or better, delete them.

src/scripts/branch_review.py
```diff
@@ -0,0 +1,295 @@
...
+
+            other_parents_in_branch = add_obsoleted_label(other_parents_in_branch, obsoletion_candidates_set)
+            other_parents_not_in_branch = add_obsoleted_label(other_parents_not_in_branch, obsoletion_candidates_set)
+            all_ancestors_in_branch = add_obsoleted_label(all_ancestors_in_branch, obsoletion_candidates_set)
+            all_ancestors_outside_branch = add_obsoleted_label(all_ancestors_outside_branch, obsoletion_candidates_set)
+            relevant_obsoletion_parents_in_branch = add_obsoleted_label(relevant_obsoletion_parents_in_branch, obsoletion_candidates_set)
+
+            # Column J: Affected Status = status
+            status = ""
+            if len(other_parents_in_branch) > 0 and all(
```
matentzn, line 200:
```suggestion
            if len(other_parents_in_branch) > 0 and any(
```

This seems to be the last issue in the table!

src/ontology/mondo.Makefile
hrshdhgd: Done!

src/scripts/branch_review.py
```diff
@@ -0,0 +1,312 @@
...
+                        curie=child,
+                        relationships=children_of_obsoletion_candidate_parents_relationship,
+                    )
+                )
+                - all_obsoletes
+            )
+
+            # These are parents that need to be obsoleted and hence NOT in Branch
+            relevant_obsoletion_parents_in_branch = (
+                parents_of_child & obsoletion_candidate_child_of_branch
```
matentzn, line 164: How can this ever be true if `parents_of_child` is `- all_obsoletes`?

matentzn: (I mean, I can see that it works from the table you shared, but I don't get it from the code, sorry :P)

src/scripts/branch_review.py
```diff
@@ -0,0 +1,312 @@
...
+            other_parents_not_in_branch = parents_of_child - other_parents_in_branch
+
+            all_ancestors = (
+                set(
+                    OI.ancestors(
+                        [child], predicates=[IS_A, PART_OF], method="ENTAILMENT"
+                    )
+                )
+                # - parents_of_child
+                - {child}
```
matentzn, line 184: What does the `{` syntax mean here? Just educational question for me

src/scripts/branch_review.py
```diff
@@ -0,0 +1,312 @@
...
+            relevant_obsoletion_parents_in_branch = add_obsoleted_label(
+                relevant_obsoletion_parents_in_branch, obsoletion_candidates_set
+            )
+
+            # Column J: Affected Status = status
+            status = ""
+            if len(other_parents_in_branch) > 0 and any(
+                " - TO_BE_OBSOLETED" not in parent for parent in other_parents_in_branch
+            ):
+                status = STAYS_IN_THE_BRANCH
```
matentzn, line 222: @sabrinatoro in my opinion, this status should be renamed to: "Retains parent in branch" for clarity

matentzn: We probably also want a code that expresses if the class retains an ancestor in he branch, and check that right after this one, so maybe this is a better naming scheme?:

1. Retains parent in branch (currently `STAYS_IN_THE_BRANCH`)
2. Retains ancestor in branch (currently missing)
3. Leaves branch but retains parent in other branch (currently `LEAVES_THE_BRANCH`)
4. Orphaned (leave as is)

matentzn (commented):
Looks very nice!

src/scripts/branch_review.py
hrshdhgd: set. `child` at this point is just a CURIE. In order for the subtraction to work, I have to put it in a set.

matentzn: Oops didn't know that `{` was the set operator! Thanks!

sabrinatoro:
@hrshdhgd and I reviewed spreadsheets created from different files (included relaxed, inferred, relaxed/inferred). Here is what we found out: 
- "relaxed" does not include parents that are inferred. 
- "inferred" (alone or in combination with "relax") always removes direct parents that are also ancestors. 

Solution (that we are testing) is to create 3 spreadsheets: 
- **spreadsheet 1**: created using the "relaxed" file. --> give ALL direct parent that are asserted
- **spreadsheet 2**: created using the "inferred" file --> give all inferred parent
- **spreadsheet 3** = **FINAL/combined spreadsheet** : combination of spreadsheet 1 and 2 (with removing all redundancy) 

note: the "affected status" (orphan, leaves branch, stays in branch) should be calculated based on the information in the "combined" spreadsheet.

src/scripts/branch_review.py
hrshdhgd: All this means is remove all already obsolete terms found by oak.

twhetzel:
@hrshdhgd @matentzn - this is the PR I intended to ask when it will be merged. I think there may also be another separate PR for this Branch Review table as well.

PR #6766, opened 2023-10-16
Generate Branch stats table

matentzn:


twhetzel:
@matentzn @hrshdhgd will this PR be merged into `master`?

hrshdhgd:
Hey @twhetzel , I'll let Nico decide.

matentzn:
> Oh, this is a PR from the script I am working on not the PR for the Branch Review table script. I don't know why Nico created a PR for this.

Its good to _always_ have draft PRs! This helps with transparency!

matentzn:
@twhetzel can we close or otherwise handle this PR so it is out of the way? I lost track of where what is..

twhetzel:
I mentioned that there are a few things to sort out about this and discuss in our next 1:1 so it needs to stay for now.

branch_ids.tsv
matentzn, line 1: This file should not be here

twhetzel: This file and it's location is where it is in the Draft PR to create the branches for curation review and is needed in this code so that is why it is currently here. I agree, that the location in the repo does not seem idea.

PR #6825, opened 2023-11-01
mass obsoletion : integumentary system disorder

sabrinatoro:
addresses #6749


twhetzel (approved):
Confirmed only 1 file changed in PR and addition of manually added GH annotations.

PR #6826, opened 2023-11-01
mass obsoletion: metabolic disease

sabrinatoro:
Addresses #6750


twhetzel (approved):
Confirmed only 1 file changed and addition of GH tickets.

PR #6827, opened 2023-11-01
mass obsoletion: nervous system disorders

sabrinatoro:
Addresses #6752


twhetzel (approved):
I confirmed there is only 1 file changed in the PR and the GH annotations are added. I spot checked a few terms (acquired amyloid peripheral neuropathy, autosomal recessive optic atrophy, primary myoclonus, syndromic vitreoretinopathy, X-linked pure spastic paraplegia) and in the quick review they look as expected.

PR #6829, opened 2023-11-02
mass obsoletion bug fix

### LINKED ISSUE #6820: mass obsoletion pipeline problem
(I am happy to add this to another issue if one already exists)

It looks like the filtered_obsolete_me.txt used for the obsoletion is not the newest copy, it is the previous copy. 

Description of the issue: 
- I updated "obsolete_me.txt" with the terms to obsolete
- I ran the obsoletion pipeline using `sh run.sh make mass_obsolete2 -B GITHUB_ISSUE_URL=https://github.com/monarch-initiative/mondo/issues/6746`
- I saw that the "filtered_obsolete_me.txt" has been updated as expected: it contained only the terms that have not been obsoleted yet (and does not include the terms that have already been obsoleted). 
- After the run, the terms obsoleted were not those on the "filtered_obsolete_me.txt". The terms obsoleted were the terms that I tested yesterday (ie that were on the last "filtered_obsolete_me.txt"; and since I tested with these terms, but not officially obsolete them, they were obsoleted)

NOTE: 
- I saw the same thing yesterday, but blamed it on the fact that I couldn't be sure that I had saved  "obsolete_me.txt". And it worked after testing again. But I am seeing the issue again, so this is definitely a bug.

twhetzel:
close #6820 

This fixes the issue with needing to run the pipeline twice to see all changes applied to the ontology file. @sabrinatoro can you please check out this branch and do a functional test to run the pipeline and confirm it is working as expected?

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -495,17 +495,19 @@ tmp/mass_obsolete_warning.sparql: ../sparql/reports/mondo-obsolete-warning.sparq
 	@echo "\n** Run mondo-obsolete-warning.sparql **"
 	LISTT="$(shell paste -sd" " config/filtered_obsolete_me.txt)"; sed "s/MONDO:0000000/$$LISTT/g" $< > $@
 
-tmp/mass_obsolete.ru: ../sparql/update/mondo-obsolete-simple.ru tmp/identify_existing_obsoletes.txt config/obsolete_me.txt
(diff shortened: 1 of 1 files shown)


sabrinatoro (approved):
@twhetzel I testing this and it worked perfectly.

I tested by obsoleting different lists of terms one after another. Every time, the terms obsoleted were as expected.

PR #6831, opened 2023-11-02
mass obsoletion: focus on "endocrine disorder" branch

sabrinatoro:
Addresses #6651


twhetzel (approved):
Confirmed only 1 file changed in PR, additions of new terms have GH ticket annotation, and all previously obsoleted terms also have GH ticket annotation.

PR #6832, opened 2023-11-02
mass obsoletion: focus on "disorder of orbital region"

sabrinatoro:
Addresses #6744


twhetzel (approved):
Confirmed only file in PR, manually added GH annotations, and tickets added to terms getting new parents. Note, 4 terms (MONDO:0007072, MONDO:0010004, MONDO:0011334, MONDO:0011512) already had their new parent as part of a previous PR from other mass obsoletion spreadsheets.

PR #6833, opened 2023-11-02
mass obsoletion: focus on respiratory system disease branch

sabrinatoro:
Addresses #6754


twhetzel (approved):
Confirmed only 1 file changed, tickets added, and 6 terms obsoleted.

PR #6835, opened 2023-11-02
mass obsoletion: focus on immune system disorder

sabrinatoro:
Addresses #6747


twhetzel (approved):
Confirmed only 1 file changed, tickets added.

PR #6837, opened 2023-11-03
mass obsoletion: focus on "infectious disease" branch

sabrinatoro:
Addresses #6671


twhetzel (approved):
Confirmed only 1 file changed and tickets added for terms obsoleted and those with new parents added.

PR #6839, opened 2023-11-03
mass obsoletion: focus on "digestive system disorder" branch

sabrinatoro:
Addresses #6743


twhetzel (approved):
Confirmed only 1 file changed, tickets added for terms with new parents and obsoleted terms and previously obsoleted terms that are in this branch too.

PR #6840, opened 2023-11-03
mass obsoletion: "cardiovascular disorder" branch

sabrinatoro:
Addresses #6740


twhetzel (approved):
Confirmed only 1 file changed and ticket annotations added to relevant terms.

PR #6863, opened 2023-11-10
Remove non exact xrefs from Mondo

matentzn:
Needs careful @cmungall review.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -202,7 +202,9 @@ filtered.obo: $(SRC)
 		--update ../sparql/update/delete-axiom-annotations-by-prefix.ru \
 		--update ../sparql/update/inject-cross-species-analog.ru \
 		remove --term-file config/remove-annotations-before-release.txt \
-		convert -f obo --check false -o $@.obo && mv $@.obo $@; fi
+		convert -f obo --check false -o $@.obo \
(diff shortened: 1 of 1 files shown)


src/ontology/Makefile
```diff
@@ -202,7 +202,9 @@ filtered.obo: $(SRC)
 		--update ../sparql/update/delete-axiom-annotations-by-prefix.ru \
 		--update ../sparql/update/inject-cross-species-analog.ru \
 		remove --term-file config/remove-annotations-before-release.txt \
-		convert -f obo --check false -o $@.obo && mv $@.obo $@; fi
+		convert -f obo --check false -o $@.obo \
+	perl -ne 'print unless (m@^xref: (Orphanet|OMIM|OMIMPS|DOID|EFO|NCIT|SCTID|MESH|UMLS|ICD10CM):@ && !(m@(obsoleteEquivalent|equivalentObsolete|equivalentTo)@i))' $@.obo > $@.tmp && mv $@.tmp $@; fi
```
matentzn, line 206:
```suggestion
	perl -ne 'print unless (m@^xref: (Orphanet|OMIM|OMIMPS|DOID|EFO|NCIT):@ && !(m@(obsoleteEquivalent|equivalentObsolete|equivalentTo)@i))' $@.obo > $@.tmp && mv $@.tmp $@; fi
```

matentzn: As per @cmungall suggestion (added MESH after the same logic)

PR #6864, opened 2023-11-10
Update imports

matentzn:
@twhetzel I ran `sh run.sh make refresh-merged` and it worked - can you also test it?


src/ontology/imports/merged_import.owl
```diff
@@ -24376,8 +24397,6 @@ Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/MAXO_0000864>))
 Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/NCIT_R163>))
 Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/NCIT_R81>))
 Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/NCIT_R82>))
-Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/PATO_0000051>))
-Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/PATO_0000070>))
```
matentzn, line 24380: Seems our problem is solved! @twhetzel

twhetzel: I checked out the other branch I was trying to fix this is (`iss-6712_refresh_import`) and ran `sh run.sh make refresh-merged` and it _did not_ fix the issue for me. I can try the command in a fresh branch and will report back.

twhetzel: Even trying the command in a new branch did not work for me as I still see the ObjectProperties in `merged_import.owl` at `mondo/src/ontology/imports`:
```
Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/PATO_0000051>))
Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/PATO_0000070>))
```

Is `sh run.sh make refresh-merged` the only command you ran to refresh the imports?

matentzn: Can you provide a list of all the files updated by this command? Sounds fishy, like I have some more up to date mirrors installed? 

You agree though that the import I have updated in the PR looks ok, right?

twhetzel: Did you run any other commands before `sh run.sh make refresh-merged` as part of the effort to refresh the imports? I did not and also did not clear the mirror directory.

In `mondo/src/ontology/mirror` almost all files were updated except `ncbitaxon.owl`, `chebi.owl`,  and `chr.owl`.

twhetzel: Yes, I do agree the PR looks correct as those are the two problematic PATO IDs that needed to be removed from somewhere as object properties (since they also existed as classes).

matentzn: I think you will have to share the log file. 

Steps to debug:

1. After running `sh run.sh make refresh-merged`
2. Check if mirror/maxo.owl has one of the bad axioms in it. If so, a problem lies with the mirroring
3. If not, check if mirror/merged.owl  has one of the bad axioms in it. If so, the problem lies with the merging
4. If not, check if imports/merged_import.owl has the bad axioms in it. If so the problems lies with the module extraction.

Let me know which of these is likely at fault on your end!

twhetzel: What log file?

1 - The file `mirror/maxo.owl` only has the PATO IDs (PATO_00000051 and PATO_0000070) as classes now
2 - I don't have a `mirror/merged.owl` file
3 - As mentioned earlier, `imports/merged_import.owl` does have the PATO terms as both classes and object properties

The only files updated in `imports` when I ran `sh run.sh make refresh-merged` are "merged_import.owl" and "ncbitaxon_terms.txt". 

When I grep mirror for one of the PATO terms (`grep -r 'Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/PATO_0000051>)' mirror`) nothing is found. When I do the same in `mirror` it's found in "merged_import.owl".

matentzn: > I don't have a mirror/merged.owl file

Did you look at src/ontology/mirror/merged.owl? This file is `.PRECIOUS`, so it should never _not_ be there. 

So first step to debug is to figure out why mirror/merged.owl is not there.

matentzn:
Closed in favour of #6961

PR #6875, opened 2023-11-16
Add: SSSOM python requirements

joeflack4:
- Update: Python requirements: With missing dependency 'sssom'


requirements-unlocked.txt
```diff
@@ -1,2 +1,3 @@
 pandas
 pyyaml
+sssom
```
joeflack4, line 3: SSSOM was missing from the requirements, so I added it.

joeflack4 (commented):
@hrshdhgd @matentzn This is probably good to merge but I'll hold off until a review.

I'm surprised how many new requirements got added to `requirements.txt` simply by adding `sssom`. Our only requirements before were `pandas` and `pyyaml`, resulting in like 15 lines in `requirements.txt` after those were installed. Now after installing `sssom` as well, it's 87 lines. I have no qualms though, I'm fine to merge this and I think this Python stuff will be moving to `mondolib` eventually anyway.

@matentzn It is still requiring @nicolevasilevsky 's review for this. I brought this up in a recent PR that the [code reviewers config](https://github.com/monarch-initiative/mondo/blob/master/.github/CODEOWNERS) was perhaps requiring too much of Nicole?

https://github.com/monarch-initiative/mondo/blob/c33200382e13f410d9a6918ff72cd40b5cedf4b8/.github/CODEOWNERS#L4

@nicolevasilevsky If you want to approve this and pass the buck to others for further review I'm happy with that!

matentzn (approved):

PR #6889, opened 2023-11-17
Add preliminary check for ontobot

hrshdhgd:
This PR checks the issue body first for the phrase `Hey ontobot`. Only if present==true, will it run the rest of the workflow. If absent, the workflow immediately stops and does nothing.


matentzn (approved):
LGTM. You can assign such PRs to Trish moving forward!

PR #6938, opened 2023-12-01
Update CODEOWNERS

matentzn:


twhetzel:
I don't think this is the exact set-up we want, but also waiting until the Org meeting to make further changes on this PR.

.github/CODEOWNERS
```diff
@@ -1,9 +1,12 @@
 # Mondo code owners (Guardians of the main branch)
 
 # Removed @cmungall to stop him being added to all PRs
-*       @nicolevasilevsky 
+*       @sabrinatoro @twhetzel
 *.sparql @matentzn
 *.py	  @joeflack4
+*.md    @matentzn @twhetzel
```
sabrinatoro, line 7: If this is documentation (including "curation" related documentation), I think I should be included here.

twhetzel: Agreed and this has been updated now.

sabrinatoro (approved):

PR #6960, opened 2023-12-06
Removing some sparsely curated xrefs from Mondo

matentzn:
This is removing a few xrefs from Mondo which are to sparse, or should not be there at all.


nicolevasilevsky (approved):
Oooh, this is nice! I like it.

PR #6967, opened 2023-12-07
Removing DERMO xref

matentzn:
Part of a greater effort to clean up all the cross references


src/sparql/qc/mondo/qc-illegal-prefix-on-xref.sparql
```diff
@@ -22,7 +22,6 @@ WHERE
   }
   FILTER( STRBEFORE(str(?value),":") not in (
       "CSP",
-      "DERMO",
```
joeflack4, line 25: I see I see. First I've heard of [DermO](https://jbiomedsem.biomedcentral.com/articles/10.1186/s13326-016-0085-x).

I we removing it because we just don't want to incorporate DermO at this time, or is it just a bad xref?

matentzn: I dont know what it is exactly, but since the xref is the only one from DERMO, and this is a hyperspecific disease (Herxheimer disease), I think it does not belong into Mondo until there is a proper strategy for it in place.

matentzn: Thanks for taking a look!

twhetzel: Thanks for the explanation about this change.

twhetzel (approved):
The explanation to not include this one reference to DERMO until a strategy is in place sounds reasonable.

sabrinatoro (approved):

PR #6987, opened 2023-12-07
Just ran `black` and ruff --fix

hrshdhgd:
Just formatted the python code in the `scripts` dir.

commands: 
- `black src/scripts/`
- `ruff --fix src/scripts/`

Did not add / remove any code manually. So functionally, nothing has changed.

### DIFF
--- a/src/scripts/add_object_label.py
+++ b/src/scripts/add_object_label.py
@@ -3,7 +3,6 @@ from sssom.parsers import parse_sssom_table
 from oaklib import OntologyResource
 from oaklib.implementations import SqlImplementation
 import click
-import sys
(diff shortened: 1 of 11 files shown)


joeflack4 (commented):
@hrshdhgd I hadn't heard of `ruff` before. Do you know if there is any specific thing that it does which `black` does not do?

Are some of our other projects starting to use `ruff`?

twhetzel (approved):
LGTM

hrshdhgd:
Also, the monarch project template Cookiecutter implements it too. It offers `flake8` and `mypy` right out of the box without installing them individually. Pretty handy.

PR #6992, opened 2023-12-08
Update the GARD subset in Mondo

matentzn:
This PR updates the GARD subset in Mondo. In particular it:

1. Adds gard_rare subset declaration
2. XREFs to GARD terms (with the ORDO/OMIM terms they were sourced from)


twhetzel:
The gard_rare subset declaration is in duplicate on many if not all terms where it is added.

matentzn:
I should have deleted the import.. Now I have to wait again for the QC to pass :D

PR #7000, opened 2023-12-08
Update NORD subset

matentzn:

PR #7016, opened 2023-12-09
Add NORD subset and cross-references

matentzn:
Supercedes #7000 

The problem was that I was using the xrefs table to generate the "nord" subset, which is a different thing altogether.

- The NORD subset is a table of all Mondo terms NORD has considered rare (>10K). 
- The NORD xrefs table is a table of about 2000K cross references to NORD. 

We need to meet with NORD to discuss exactly what we are pulling in moving forward. This PR here does not change the status quo, as it is a combination of what we previously had (the "old" nord subset) and the "approved" NORD xrefs.

### DIFF
--- a/src/sparql/qc/mondo/qc-illegal-prefix-on-xref-annotation.sparql
+++ b/src/sparql/qc/mondo/qc-illegal-prefix-on-xref-annotation.sparql
@@ -61,6 +61,7 @@ WHERE
       "MESH",
       "MFOMD",
       "MONDO",
+      "NORD",
       "MONDORULE",
       "MP",
       "MPATH",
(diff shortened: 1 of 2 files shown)


matentzn:
I am merging this is as the changes are what you have approved @sabrinatoro + adding the _previous_ nord_subset declaration back (which are the correct ones).

twhetzel:
FYI - we talked Thursday about not bypassing PR review unless it's an emergency. If you continue to run into these special cases in this repo and others I've been seeing then we should sort out a better working plan for reviews.
cc: @monicacecilia @sabrinatoro

matentzn:
I made this Call because we won't be able to do all the PRs needed for the Tuesday release, and the contents of this pr were already reviewed - all of them. Just on two different PRs. I wont do it again.

PR #7017, opened 2023-12-09
Update Orphanet Rare disease subset

matentzn:
This PR is part of the series of updating all the rare-subsets (and some others, after the next release). We previously had them in an externally managed file; they are now all moved back into the edit file where they were originally.


sabrinatoro:
@matentzn 
There is 9950 terms with the Orphanet "rare disease" subset annotation. Is that the number you are expecting? 
Also, there are no "source" for these subset annotations. Should we add the Orphanet ID (similar to what we do for GARD)?

sabrinatoro (approved):
@matentzn I am approving this PR so you are not blocked with PR merging. 
Please re-assign to me for review if the number of annotations and the source needs to be updated, and you need me to take another look before merging.

matentzn:
Made a new PR with provenance #7020

PR #7020, opened 2023-12-11
Add orphanet subset including source information

matentzn:
Supercedes #7017 

@sabrinatoro this time with provenance! Thanks for the review.


sabrinatoro (approved):
@matentzn : this PR looks good. 
Note for future us: I found an x-ref which is "MONDO:equivalentObsolete", and the term is in the orphanet_rare subset. We will need to include this case when we discuss rare disease subset maintenance.

matentzn:
@sabrinatoro yes-this is true. Since subsets are externally managed, they can contain obsolete classes. I added that talking point to the  Mondo technical call agenda.

PR #7021, opened 2023-12-11
Update rare and rare inferred subsets

matentzn:
This PR updates the rare and rare inferred subsets.

Note that I didn't add provenance as we manually curate these.


sabrinatoro (approved):
@matentzn 
It looks good to me. 
I see: 
4330 terms with rare-inferred subset
19184 terms with rare subset. 

If these numbers look ok, it is ready to merge after QC have passed.
Thank you @matentzn !

matentzn:
There are 16209 "rare" terms (which are way too many, but that is a problem for the next iteration (this is in line with what we had before)

PR #7023, opened 2023-12-11
Add "Obsoletion candidate" tag for Dec 2023 release

sabrinatoro:
Based on this obsoletion candidate list
https://docs.google.com/spreadsheets/d/1tt1Wk70j9XiHLV1vKQyNiHhaazh286pobpJk1ecSCCg/edit#gid=434975340

### DIFF
--- a/src/sparql/reports/existingcomment.sparql
+++ b/src/sparql/reports/existingcomment.sparql
@@ -1,0 +1,20 @@
+prefix owl: <http://www.w3.org/2002/07/owl#>
+prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix MONDO: <http://purl.obolibrary.org/obo/MONDO_>
+
+#description: terms with comments
+
+SELECT DISTINCT ?entity ?comment WHERE {
+  	VALUES ?entity {
+    MONDO:0011469
+    MONDO:0011876
+  }
+  
+    ?entity rdfs:comment ?comment .
+        
+   FILTER (isIRI(?entity) && STRSTARTS(str(?entity), "http://purl.obolibrary.org/obo/MONDO_"))
+}
+
+ 
+


src/sparql/reports/existingcomment.sparql
```diff
@@ -0,0 +1,20 @@
+prefix owl: <http://www.w3.org/2002/07/owl#>
+prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix MONDO: <http://purl.obolibrary.org/obo/MONDO_>
+
+#description: terms with comments
+
+SELECT DISTINCT ?entity ?comment WHERE {
```
joeflack4, line 8: Maybe it's too early for me to review; I see that this is just in draft!

I'm assuming that this will change to a sparql update, and will be run on the 10 or so terms in the spreadsheet!

sabrinatoro: yes, it is still early.
I think the reason you were added is because this PR will include a Sparql query for report. 
I'll change this PR into "ready to review" when I have added tags

sabrinatoro:
Assigning to Trish and Nicole (whoever is reviewing this PR should untag the other person as the assignee.

src/ontology/mondo-edit.obo
```diff
@@ -31495,8 +31495,10 @@ is_a: MONDO:0019052 {source="DOID:14749/inferred", source="MONDO:Redundant", sou
 id: MONDO:0002013
 name: lymphangioma
 def: "A benign lesion composed of dilated lymphatic channels. Painless swelling is the usual clinical manifestation." [NCIT:C8965]
+comment: Reason of obsoletion: out of scope - MONDO:excludeHistoricalDisease. Term to consider: lymphatic malformation-
```
twhetzel, line 31498: Is this missing part of the term name, `lymphatic malformation-`?

twhetzel: Actually, just looks like `-` is extra.

twhetzel:
I am reviewing this now.

src/sparql/reports/existingcomment.sparql
twhetzel: @sabrinatoro is there an updated query to commit?

twhetzel (approved):
lgtm

src/sparql/reports/existingcomment.sparql
twhetzel: Or was this file not intended to be committed?

joeflack4: @sabrinatoro I know this is is merged now, but just wanted to follow up. Was this file intentionally added? Is it related to the main work in this PR or is it of a different purpose?

I just mention that because I expected maybe this file was boilerplate and it was going to change to an update query that updated the ontology to add the obsoletion candidate related edits, but that doesn't seem to be teh case.

twhetzel: @joeflack4 great question! I had a chance to chat with Sabrina earlier today and this is an example query so curators can run this in general, modulo adding in new curies, when this needs to be done in the future.

PR #7054, opened 2023-12-18
Remove @sabrinatoro as assignee for GH workflows

hrshdhgd:
Remove @sabrinatoro as assignee for GH workflows


hrshdhgd:
@sabrinatoro , there are also markdown tables e.g. : https://github.com/monarch-initiative/mondo/blob/master/.github/ISSUE_TEMPLATE/add-gene-related-syndrome.md that have you listed as an assignee in the table. Is this okay to be unchanged?

sabrinatoro (approved):
Thank you @hrshdhgd

PR #7091, opened 2024-01-03
release 2024-01-03

sabrinatoro:


sabrinatoro:
@matentzn I am seeing something strange. 
When I look at mondo.obo (or mondo.owl), there are some axioms that I don't see, though these are in the mondo-edit. For example: 
- MONDO:0700227: has an equivalent axiom which is not in mondo.obo
- MONDO:0010819: is missing the axiom: 'has material basis in germline mutation in' some ELOVL4
- MONDO:1010683: is missing the equivalent  'non-human animal disease' and ('cross-species analog' some 'AA amyloidosis') and ('in taxon' some 'Gallus gallus'.     It is also missing the individual axioms
- same for all new non-human diseases.

matentzn:
I have provided my analysis in slack; I think we can proceed without fixing anything. But please make sure you are convinced of my analysis!

twhetzel (approved):
Approving based on earlier review by Nico and Sabrina.

PR #7132, opened 2024-01-15
Add Mapping from renal leiomyoma to kidney leiomyoma in NCIT

Silvia113-SM:
In this PR we mapped renal leiomyoma in MONDO (MONDO:0006053) to kidney leiomyoma in NCIT. We decided that it is exactMatch because renal = "belonging to kidney"


sabrinatoro (approved):
The change makes sense and the x-ref was added correctly. Thank you @Silvia113-SM

PR #7199, opened 2024-01-25
Add EPM to `mondo.sssom.config.yml`

joeflack4:
## Changes
- Update: mondo.sssom.config.yml: Added EPM (extended prefix map)

---

@souzadevinicius FYI

### DIFF
--- a/src/ontology/metadata/mondo.sssom.config.yml
+++ b/src/ontology/metadata/mondo.sssom.config.yml
@@ -76,6 +76,274 @@ curie_map:
   # SNOMEDCT_2010_1_31: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/SNOMEDCT_2010_1_31/"
   OMIA: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/OMIA/"
 
+extended_prefix_map:
+  - prefix: MONDO
+    prefix_synonyms: []
+    uri_prefix: http://purl.obolibrary.org/obo/MONDO_
+    uri_prefix_synonyms: []
+  - prefix: UMLS
+    prefix_synonyms: []
+    uri_prefix: http://linkedlifedata.com/resource/umls/id/
+    uri_prefix_synonyms: []
+  - prefix: NCIT
+    prefix_synonyms: []
+    uri_prefix: http://purl.obolibrary.org/obo/NCIT_
+    uri_prefix_synonyms: []
+  - prefix: DOID
+    prefix_synonyms: []
+    uri_prefix: http://purl.obolibrary.org/obo/DOID_
+    uri_prefix_synonyms: []
+  - prefix: EFO
+    prefix_synonyms: []
+    uri_prefix: http://www.ebi.ac.uk/efo/EFO_
+    uri_prefix_synonyms: []
+  - prefix: HGNC
+    prefix_synonyms: []
+    uri_prefix: http://identifiers.org/hgnc/
+    uri_prefix_synonyms:
+      - 'https://identifiers.org/hgnc:'
+      - 'https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:'
+  - prefix: HGNC_symbol
+    prefix_synonyms: []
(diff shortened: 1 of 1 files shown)


src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -76,6 +76,274 @@ curie_map:
   # SNOMEDCT_2010_1_31: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/SNOMEDCT_2010_1_31/"
   OMIA: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/OMIA/"
 
+extended_prefix_map:
```
joeflack4, line 79: ## Add EPM
I made a mistake when recently working on #394. I added in this `extended_prefix_map` into the copy of `mondo.sssom.config.yml` in the `mondo-ingest` repo. However, the source of this file actually comes from `mondo.  So really the workflow needs to be:

Update the file in `mondo` -> pull update into `mondo-ingest`

Future considerations:
- I think that `mondo-ingest` should add this file to the `.gitignore` so that this doesn't automatically happen.
- We need to figure out how to make sure that when files like this get updated in `mondo`, `mondo-ingest` can automatically update. This happens automatically in fresh builds, but not in development.

joeflack4, line 79: ## Check / update for correctness @joeflack4 
- [ ] Check / update for correctness

<details><summary>Details</summary>
<p>

This `extended_prefix_map` was created by transforming the `curie_map` in `mondo.sssom.config.yml` in `mondo-ingest`. However, that `curie_map` is not the same as the one here. I need to recreate this `extended_prefix_map` so that it is the most up-to-date. I don't know if (a) there were things about `mondo-ingest`'s version that were better than `mondo`, or (b) there were things about the `mondo` version that were better than `mondo-ingest`, or (c) both. I suspect 'b', but I'd have to look into it. I'm also not sure I'm the best person to make that determination.

To recreate the EPM, I could use the new `curies` function that Charlie and I created to generate an EPM based on an input bimap (e.g. `curie_map` in `mondo.sssom.config.yml`). Or I could just create it manually.

</p>
</details>

joeflack4 (commented):
.

src/ontology/metadata/mondo.sssom.config.yml
matentzn: I think these two files (mondo and mondo ingest mapping configs) may have to diverge (separately managed). They must be very different from one another by now.. How different are they? They serve also quite different purposes. I would tend to remove the update mechanism and curate them both separately.

matentzn: The EPM looks great but as I said above, we probably need it only in mondo-ingest if we sever the connection between mondo.sssom.config in Mondo and mondo ingest repo due to their very different use cases (mondo ingest: ETL, lexmatch; mondo repo: publication of SSSOM files to community.

joeflack4: Ah, I see. Well then, let's keep them separate. Are you OK with this plan:
- [ ] 1. I use me and Charlie's new function to convert the `mondo` `curie_map` into an `extended_prefix_map` and update this PR so that `mondo.sssom.config.yml` here is internally consistent.
- [ ] 2. I update `mondo-ingest` and remove the goal for `mondo.sssom.config.yml`, and just keep it committed as a static file that needs to be maintained there.

joeflack4: Ahh, OK. I'm resolving this thread now then. See action items here: https://github.com/monarch-initiative/mondo/pull/7199#discussion_r1467092075

matentzn: 2 for sure. 

1 you can try, but I am not sure what the consequences of "making it consistent" will be! We shall see I guess.

joeflack4: Ok, I'm going to close this, do (2) and create an issue for (1), I guess low priority.

Issue #7232, opened 2024-02-03
Design of NORD endpoint for mappings 


matentzn:
Sean has implemented an amazing endpoint for us to obtain the NORD data for Mondo.

The current NORD API endpoint returns the following information

| mondo_id      | report_ref | report_ref_source | preferred_name                                                                | preferred_name_source | synonym_type                                        | subset                                              | subset_source | subset_soource2 |
|---------------|------------|-------------------|-------------------------------------------------------------------------------|-----------------------|-----------------------------------------------------|-----------------------------------------------------|---------------|-----------------|
| MONDO:8000024 |            | MONDO:NORD        | autoimmune lymphoproliferative syndrome, type III caused by mutation in PRKCD |                       | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED |               | MONDO:NORD      |
| MONDO:8000018 | NORD:850   | MONDO:NORD        | benign paroxysmal positional vertigo                                          | NORD:850              | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED | NORD:850      | MONDO:NORD      |
| MONDO:8000015 |            | MONDO:NORD        | 46,XY sex reversal 11                                                         |                       | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED |               | MONDO:NORD      |
| MONDO:8000012 |            | MONDO:NORD        | neurologic, endocrine, and pancreatic disease, multisystem, infantile-onset 1 |                       | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED |               | MONDO:NORD      |
| MONDO:8000011 |            | MONDO:NORD        | visceral neuropathy, familial, 1, autosomal recessive                         |                       | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED | http://purl.obolibrary.org/obo/mondo#NORD_PREFERRED |               | MONDO:NORD      |
[...]

PR #7233, opened 2024-02-03
Create QC check to notify curators if an animal disease ends up in the rare disease subset

### LINKED ISSUE #7030: Mondo rare subset: remove non-human disease animal
The current rare disease subset includes non-human disease animal diseases. 
The rare disease subset should only include human diseases. 

It looks that the only culprits are 2 non-human disease terms that have the NORD_rare subset annotation: 
- tuberculosis, avian (MONDO:0025135)
- anaplasmosis in cattle (MONDO:0044748)

(the other terms are classification terms which are included for classification reasons)

matentzn:
Fixes #7030


matentzn:
@twhetzel I assign this PR to you as its failure is prove that it works. To proceed you can either:

- Wait until NORD subset is updated #7232
- Add QC exceptions to the two classes that are currently rare
- Manually drop the "nord_rare" subsets from both of these as Gioconda says in https://github.com/monarch-initiative/mondo/issues/7030#issuecomment-1881710637

I kinda prefer the third as it

1. Can be done right now, quickly
2. Doesnt do any damage

src/sparql/qc/mondo/qc-animal-disease-rare.sparql
```diff
@@ -0,0 +1,25 @@
...
+prefix mondoSparqlQcMondo: <http://purl.obolibrary.org/obo/mondo/sparql/qc/mondo/>
+prefix mondo: <http://purl.obolibrary.org/obo/mondo#>
+
+# Tests if an animal disease made it into the rare disease subset
+
+SELECT DISTINCT ?entity ?property ?value WHERE
+{
+  VALUES ?property { <http://purl.obolibrary.org/obo/mondo#rare> }
+  ?entity <http://www.geneontology.org/formats/oboInOwl#inSubset> ?property .
+  ?entity rdfs:subClassOf* <http://purl.obolibrary.org/obo/MONDO_0005583>
```
joeflack4, line 16: Just providing my code reviewer face check. Syntax and everything looks OK to me. Verified that `MONDO_0005583` is for animal disease, and looks like correctly selects that which is in the rare disease subset.

I actually don't 100% know how filter this works:
`?entity mondo:excluded_from_qc_check mondoSparqlQcMondo:qc-animal-disease-rare.sparql .`
I didn't know we had this fine-grained exclusion of things from QC check. Looks like not only is there a predicate for excluding from QC check, but there is this weird file name as property value that I've never seen before. I'm going to guess perhaps that there are currently no such exclusions, but that this is a standard thing we put in our QC checks so that we can easily add exclusions in the future.

Only a minor cosmetic suggestion:
It looks like it returns distinct `?entity ?property ?value`: `CURIE <http://purl.obolibrary.org/obo/mondo#rare> "Animal disease in Rare subset"`
If I'm correct about that this change is more readable`?entity ?value`: `CURIE "Animal disease in Rare subset"`

matentzn: > I actually don't 100% know how filter this works:
?entity mondo:excluded_from_qc_check mondoSparqlQcMondo:qc-animal-disease-rare.sparql .

Yes, the main problem is that there are a number of very useful QC checks that do not _always_ apply, but most of the time. So we invented this mechanism to allow curators to explicitly bypass a QC check. `mondo:excluded_from_qc_check` is a relation in Mondo (annotation property) and `mondoSparqlQcMondo:qc-animal-disease-rare.sparql` is the PURL of the QC check called `animal-disease-rare`. 

If you look at mondo-edit.obo, you will find many such exclusions on other tests.

 > Only a minor cosmetic suggestion:

All QC checks return the `?entity ?property ?value` in that order, which is a prerequisite of using them in the context of `robot report`. Now it returns

```
http://purl.obolibrary.org/obo/MONDO_0025135,http://purl.obolibrary.org/obo/mondo#rare,Animal disease in Rare subset
```

Can you say again what you would think is better to bind to each of the three variables?

joeflack4: Thanks for taking time to get me up to speed!

As `?entity ?property ?value` is the standard for all QC returns (I didn't know), then I suggest we change nothing.

All I was saying was that "mondo#rare" is implicit in "Animal disease in Rare subset".

twhetzel:
> Manually drop the "nord_rare" subsets from both of these as Gioconda says in https://github.com/monarch-initiative/mondo/issues/7030#issuecomment-1881710637 
@matentzn are you requesting only "nord_rare" be removed?

Should "rare" also be removed?

matentzn:
> Should "rare" also be removed?

`rare` is updated automatically, you can remove it as well if you like, but next time I update the subsets, which is currently only in https://github.com/monarch-initiative/mondo/pull/6903 (not master), it will remove itself as well.

twhetzel:
Ok, I didn't know that `rare` is updated automatically so I will not remove it in the PR.

matentzn:
> Ok, I didn't know that rare is updated automatically so I will not remove it in the PR.

Argh I didnt think this through, sorry - the PR will not pass if we dont make that change manually once since the QC check is looking for the "rare" subset specifically; or else I need to update the whole subset, which adds a lot of spurious information to that PR. Sorry. Brain malfunction.

twhetzel:
No worries, that makes more sense to me that both are removed as a "hotfix" for now. The [PR](https://github.com/monarch-initiative/mondo/pull/7254) is updated with this change.

twhetzel:
Merging now that the QC checks pass based on https://github.com/monarch-initiative/mondo/pull/7254#issuecomment-1935381262

PR #7270, opened 2024-02-13
non-human disease by gene - dosdp

sabrinatoro:
This PR create a new pattern for non-human diseases coming from omia and defined by a variation in a specific gene.

Note that this pattern is the same as the human disease pattern, but the class for disease is specific to non-human diseases.

### DIFF
--- a/src/patterns/dosdp-patterns/nonhuman_disease_by_gene.yaml
+++ b/src/patterns/dosdp-patterns/nonhuman_disease_by_gene.yaml
@@ -1,0 +1,50 @@
+pattern_name: nonhuman_disease_by_gene
+
+pattern_iri: http://purl.obolibrary.org/obo/mondo/patterns/nonhuman_disease_by_gene.yaml
+
+description: '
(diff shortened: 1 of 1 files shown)


src/patterns/dosdp-patterns/nonhuman_disease_by_gene.yaml
matentzn, line 1: Before we get to the details of the pattern, we need to get a sense of the practicality of creating a new pattern for every single non-human disease class which does not have a "corresponding human term".

There are some other options, like 

1. Making the `disease` slot configurable and using an enum validation check that it is either one of `human` or `non-human animal disease`. 

Also we need to be careful if the corresponding human pattern is using "human disease" (not "disease") as the disease slot, else all the animal diseases will end up as children of human diseases due to inference.

matentzn (changes requested):

sabrinatoro:
@matentzn and I discussed this on our 1:1 on Feb 14th, 2024. 
The pattern [OMIM_disease_series_by_gene](https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml) refers to the variable "disease" as "MONDO:0000001", which is the parent term of "human disease" and "non-human disease". Therefore, it would make sense to keep the original for both human and non-human diseases. (note: I have confirmed that "human disease" and "non-human disease" are disjointed). 
I will close this PR, and create a new one where the description of the original pattern is updated to reflect this.

Issue #7271, opened 2024-02-13
Add non-human animal genetic diseases


katiermullen:
Add OMIA species-specific genetic diseases using NCBIGene ids.

See spreadsheet: https://docs.google.com/spreadsheets/d/1aKuKc2yKy4Z1_yzntsOSi-Z-2V_AMKdmcwOZsVpUwtM/edit#gid=1157876657 'Species specific genetic diseases.'

To do:

- [x] Create patterns
- [x] Add NCBIGene into Mondo. (Checked with Monarch - NCBIGene is fine for this)
- [ ] Add the terms from OMIA.

katiermullen:
Discussion with @kevinschaper and @sabrinatoro on using NCBIGene for non-human animals in Monarch:

Kevin Schaper
Jan 26, 2024, 8:24 PM
to Sabrina, me

Sorry for not replying earlier. 

I'm not 100% sure what it will look like for a gene to disease association to come in from Mondo (via Phenio, and the kgx translation into kg-phenio), but I think I should be able to pick up the association, and leave out the gene node coming in from the ontology in favor of getting it from NCBIGene. 

I'm not sure yet about adding genes for 130 species - I might bring up that question in the huddle on Tuesday. It seems like an obvious yes that we should bring in any gene from any species if we have phenotype data for, but I don't know whether that also means that we should include every gene from each species, and then get interactions+expression etc for all of those genes, etc etc. 



On Wed, Jan 24, 2024 at 5:09 PM Sabrina Toro <sabrina@tislab.org> wrote:
Thank you Katie, 

Kevin, please be aware that we are only looking at OMIA for Mondo (ie, we are only adding the OMIA diseases in Mondo). 
Monarch (in general) could get a lot of information for OMIA, but I think that it will be much easier to wait for OMIA to update their system to use ontology terms such as Mondo (for disease) and uPheno (or MP) for phenotypes. At this point, I think it would be way too much work for Monarch to ingest their data as they are.
We have been helping them with the transition to using ontology terms, and have been collaborating with them for a while now (on VBO and Mondo). We therefore have a very good interaction with the OMIA team (and they are extremely nice and approachable and will be very excited for Monarch to ingest their data). So whenever you are ready for it, we can facilitate the conversation (whenever you are ready). 



On Wed, Jan 24, 2024 at 4:51 PM Katie Mullen <katie@tislab.org> wrote:
HI Kevin,

[...]

PR #7277, opened 2024-02-14
Update disease_by_gene pattern

sabrinatoro:
Address #7271

The pattern [OMIM_disease_series_by_gene](https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml) refers to the variable "disease" as "MONDO:0000001", which is the parent term of "human disease" and "non-human disease". 
Therefore, this pattern could be used for both genetic diseases for human (from OMIM) and non-human animal (from OMIA). 

I updated the existing description to reflect the above. 
Note that I did not rename the pattern because it has been used "everywhere" in the ontology. 

@matentzn @cmungall @katiermullen

### DIFF
--- a/src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml
+++ b/src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml
@@ -4,10 +4,10 @@ pattern_iri: http://purl.obolibrary.org/obo/mondo/patterns/disease_series_by_gen
 
 description: '
 
-  This pattern is meant to be used for OMIM Mendelian diseases (ie unitary genetic diseases, as described in
+  This pattern is meant to be used for (1) OMIM Mendelian diseases (ie unitary genetic diseases, as described in
   [PMID:33417889](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7820621/)), including children of OMIM phenotypic
-  series (OMIMPS), which are represented as grouping classes in Mondo. Notes about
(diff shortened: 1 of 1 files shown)


src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml
```diff
@@ -4,10 +4,10 @@ pattern_iri: http://purl.obolibrary.org/obo/mondo/patterns/disease_series_by_gen
 
 description: '
 
-  This pattern is meant to be used for OMIM Mendelian diseases (ie unitary genetic diseases, as described in
+  This pattern is meant to be used for (1) OMIM Mendelian diseases (ie unitary genetic diseases, as described in
   [PMID:33417889](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7820621/)), including children of OMIM phenotypic
-  series (OMIMPS), which are represented as grouping classes in Mondo. Notes about
-  the OMIMPS (see also OMIM_phenotypic_series.yaml):  - every instance of the OMIMPS
+  series (OMIMPS), which are represented as grouping classes in Mondo, and (2) OMIA Mendelian diseases (https://omia.org/) which are defined by a variation in a gene.
+  Notes about the OMIMPS (see also OMIM_phenotypic_series.yaml):  - every instance of the OMIMPS
```
matentzn, line 10: Missing new line before the "every instance"?

matentzn (commented):
I dont feel empowered to review this by FWIW LGTM.

PR #7436, opened 2024-03-19
March Orphanet alignment

sabrinatoro:
robot template and excluded terms [here](https://docs.google.com/spreadsheets/d/134EwohbnPms3qVN3eTlo3cnZ_Bu1BvXmcs1uhOLc0fs/edit#gid=1213304026)


nicolevasilevsky (approved):
there is not much in here this month! That's good, we're staying on top of it. :)

sabrinatoro:
I agree! We are getting close to being in a "maintenance" phase.

PR #7441, opened 2024-03-20
DO-alignment - part1: x-ref

sabrinatoro:
added x-ref based on exact lex and lex match: https://docs.google.com/spreadsheets/d/1dZt0Xz0KpZ4738w00P_Tcz0ur_fGF1OuHXWYxBEKNqo/edit#gid=279804497

Note that 2 new terms were created, a term was renamed to mirror the situation in OMIM.


nicolevasilevsky (approved):
looks good, thanks!

PR #7449, opened 2024-03-20
update import to add 2 new NCBITaxon

sabrinatoro:


twhetzel:
@sabrinatoro do you also plan to create a PR to add `NCBITaxon:87882` to https://github.com/obophenotype/ncbitaxon/blob/master/subsets/taxon-subset-ids.txt? 

I think this additional step was added during a time you were OOO, https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/#detailed-instructions-for-adding-a-new-ncbi-taxon-identifier

sabrinatoro:
> @sabrinatoro do you also plan to create a PR to add NCBITaxon:87882 to https://github.com/obophenotype/ncbitaxon/blob/master/subsets/taxon-subset-ids.txt?

I did not, and didn't realize there was a specific extra step for NCBITaxon. I will try this and reach out if there are still problems. Thanks!

twhetzel:
This extra step will not fix the other problem of the taxon not being added, but also hopefully will not create any new problems :)

PR #7464, opened 2024-03-22
obsolete obsolete alpha-thalassemia-abnormal morphogenesis

### LINKED ISSUE #7196: MONDO:0022425 alpha-thalassemia-abnormal morphogenesis[Obsolete]
**Mondo term (ID and Label)**
MONDO:0022425 alpha-thalassemia-abnormal morphogenesis

**Reason for deprecation**
Only xref on this is a GARD link that is no longer active in GARD. I cannot find any entries in UMLS based on the names provided in mondo for this entity. 
I do find an article from 1998 describing the case (by authors D N Abuelo , E N Forman, L P Rubin ) https://pubmed.ncbi.nlm.nih.gov/9028450/ and see that it is cited in an article for Hb Barts hydrop fetalis syndrome survivors registry https://pubmed.ncbi.nlm.nih.gov/28057638/ suggesting that perhaps this condition was actually BHFS? 

None or MONDO:0015579 Hb Bart's hydrops fetalis ?

twhetzel:
close #7196


sabrinatoro (commented):
@twhetzel 
There is no comment in the issue about whether you (or someone else) agreed with the obsoletion and reviewed the suggested term to add as "consider". 

After a quick review, I think it would be appropriate to add the annotation "consider" MONDO:0015579 Hb Bart's hydrops fetalis. Though it is definitely an edge case. 

Please take a look (or comment if you have already looked) and either add this "consider" annotation, or send this PR back to me if you think it should not be added.

twhetzel:
@sabrinatoro - since there was not a note to confirm this should be deleted I did not previously confirm. I have only seen this term listed on the [GARD page](https://rarediseases.org/gard-rare-disease/alpha-thalassemia-abnormal-morphogenesis/) and the  "D N Abuelo , E N Forman, L P Rubin" reference mentioned in the ticket when searching PubMed (found using one of the synonyms of this term). Are you still ok with obsoleting this term?

I did not add the suggested "consider" term (MONDO:0015579 Hb Bart's hydrops fetalis) since it was not listed here `comment: Reason of obsoletion: out of scope - MONDO:excludeHistoricalDisease. Term to consider: -`. 

After further reading, this (MONDO:0015579 Hb Bart's hydrops fetalis) seems a viable option for `consider` since "Hb Bart's hydrops fetalis" does involve a form of abnormal morphogenesis alpha-thalassemia. 

Another `consider` option is: 
- hemoglobin H disease MONDO:0013512 - https://www.ncbi.nlm.nih.gov/books/NBK441826/#:~:text=HbH%20disease%20is%20classified%20as,fatal%20hydrops%20fetalis%20in%20utero.

Do you want to obsolete the term and have me add these two `consider` options?

sabrinatoro:
@twhetzel we mentioned several times on different calls and in the meeting minutes that all obsoletion/merge issues should be reviewed before proceeding. Because we do not want to "lose 2 months", we add the obsoletion/merge tags when the issue is received, BUT we should always review the issue and make sure that we agree with the obsoletion/merge (and start a discussion if needed) before proceeding. 

For this specific issue, based on Megan's description and your analysis above, I think that it makes sense to obsolete and add the 2 "consider" options.

twhetzel:
Gotcha, since some of the other obsolete only tickets had explicit comments to review them I didn't realize you always want obsolete (without merging) tickets and merge(which includes an obsoletion) tickets to both be reviewed for agreement before proceeding. I thought you were saying as you write obsoletion/merge to mean tickets tagged with [Merge].

Consider annotations added and ready for review.

sabrinatoro (approved):

PR #7499, opened 2024-03-28
update name

### LINKED ISSUE #7433: SPOAN syndrome
http://purl.obolibrary.org/obo/MONDO_0012297
SPOAN Syndrome 

**Suggested new label:**
Spastic paraplegia, optic atrophy, and neuropathy

https://www.clinicalgenome.org/affiliation/40113/ 

The longer name is already listed as both a synonym and the preferred term to use by OMIM: https://www.omim.org/entry/609541?search=spoan&highlight=spoan.

sabrinatoro:
Closes #7433
Note I kept the original name with the MONDO id to indicate that this name was the original label


twhetzel (approved):
lgtm

twhetzel:
Question: Is keeping the MONDO id as the source for the original label that is changed to an exact synonym the SOP and can be added [here](https://mondo.readthedocs.io/en/latest/editors-guide/f-entities/#disease-naming)?

PR #7507, opened 2024-03-29
Update NORD subset March 2024

matentzn:
- [ ] I would suggest a 20 minute review at least, as the system stabilizes.

I am noticing:

- A number of "exact" synonyms where we would ordinarily say "related". This suggests that there are cases here NORD (cc @galyea123) assigns a Mondo ID even in cases where the NORD report is a not exactly identical. 
- A number of newly added synonyms from "NORD", but no "NORD_LABEL" synonym type.

Thats about it.


sabrinatoro (approved):

twhetzel:
>> The QC should be fixed though.
@sabrinatoro what QC should be fixed?

sabrinatoro:
The QC should be fixed with the synonyms review I have done. (this was written before the QC finished running)
And the QC was indeed fixed.

galyea123:
[like]  Gioconda Alyea reacted to your message:
________________________________

PR #7509, opened 2024-03-29
Update construct-orphanet-rare-subset.sparql - make sure the subset does not include obsolete mappings

matentzn:


sabrinatoro (approved):
looks good to me.

PR #7512, opened 2024-03-29
New updates from NORD

matentzn:
I dont know why they are that many..


sabrinatoro:
I think that the second update had a lot of new annotations because the previous one had a lot of lost one. So I think with both PR, we didn't win/lost much.

sabrinatoro (approved):

PR #7516, opened 2024-04-01
ran NORM + add obsolete tags

sabrinatoro:
Ran NORM, which updated all the gene names.


twhetzel (approved):
lgtm - confirmed all terms in the obsolete only and merge gSheets are marked in the PR

PR #7578, opened 2024-04-10
Dont look - Parking NANDO alignment pipeline here until we reach an agreement

matentzn:


matentzn:
Replaced by general EMC pipeline.

PR #7519, opened 2024-04-02
Add OTAR subset and update EFO mappings

matentzn:
This PR:

- adds the open target subset to Mondo
- Updates the Mondo-EFO mappings according to the current EFO
- @zoependlington I would appreciate a brief review (just sampling some content) to make sure this all looks right

PR #7594, opened 2024-04-12
Update OTAR subset Attempt 2

matentzn:
Replaces #7519 

- adds the open target subset to Mondo
- Updates the Mondo-EFO mappings according to the current EFO
- DOID evidence is removed from xrefs, as xrefs are now maintained by the source (EFO)
- Non-exact mappings to EFO are dropped

- @zoependlington I would appreciate a brief review (just sampling some content) to make sure this all looks right


twhetzel (approved):
should be good after changes following [PR attempt 1](https://github.com/monarch-initiative/mondo/pull/7519)

PR #7602, opened 2024-04-18
added "icd11.foundation" x-ref

sabrinatoro:
NOTE: I don't know which issue to refer to.

In this PR:
**1.  icd11.foundation X-ref based on orphanet/icd11foundation mappings**
   - curators reviewed the mappings for which the Mondo and icd11 labels were not the same
   - spreadsheet with review is [here](https://docs.google.com/spreadsheets/d/1npe1PdZrb0YTtAbPqgu7OZ961g7KwC0_eZLAlK7RCEc/edit#gid=1469844691)
   - we added here the (1) x-ref for which were the same in Mondo and ICD11, (2) x-ref which curators agreed with the mapping
   - sources for these x-ref: 
              - MONDO:equivalentTo or MONDO:obsoleteEquivalent 
              - orphanet ID that mapped to icd11 (indicating that this x-ref was supported by this mapping) 
              - orcid of curator who reviewed this mapping

NOTE: review indicated that
- some orphanet x-ref in Mondo needed to be updated
- we question some orphanet/icd11 mappings

**2.  Update of QC check to allow icd11.foundation as prefix for xref**
Based on discussion on Slack, icd11.foundation is the accepted prefix. However I am not sure about the capitalization.

### DIFF
--- a/src/sparql/qc/mondo/qc-illegal-prefix-on-xref.sparql
+++ b/src/sparql/qc/mondo/qc-illegal-prefix-on-xref.sparql
@@ -33,6 +33,7 @@ WHERE
       "ICD10EXP",
       "ICD10WHO",
       "ICD11",
+      "icd11.foundation",
       "ICD9",
       "ICD9CM",
       "ICDO",
--- a/src/ontology/profile.txt
+++ b/src/ontology/profile.txt
@@ -36,7 +36,7 @@ ERROR	file:../sparql/qc/mondo/qc-proxy-merges.sparql
 ERROR	file:../sparql/qc/mondo/qc-related-exact-synonym-omim.sparql
 ERROR	illegal_use_of_built_in_vocabulary
 ERROR	invalid_entity_uri
-ERROR	invalid_xref
+WARN	invalid_xref
 ERROR	label_formatting
 ERROR	label_whitespace
 ERROR	lowercase_definition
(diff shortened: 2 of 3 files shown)


src/sparql/qc/mondo/qc-illegal-prefix-on-xref.sparql
```diff
@@ -33,6 +33,7 @@ WHERE
       "ICD10EXP",
       "ICD10WHO",
       "ICD11",
+      "icd11.foundation",
```
joeflack4, line 36: ## `icd11.foundation` prefix
Yep, this is the prefix that we are using for this source!

src/ontology/mondo-edit.obo
```diff
@@ -61627,6 +61656,7 @@ xref: DOID:5683 {source="MONDO:equivalentTo"}
 xref: GARD:15010 {source="Orphanet:145"}
 xref: ICD10CM:C50 {source="MONDO:relatedTo", source="Orphanet:145/nd", source="Orphanet:145/attributed", source="Orphanet:145"}
 xref: ICD10CM:C56 {source="Orphanet:145/nd", source="Orphanet:145/attributed", source="Orphanet:145"}
+xref: icd11.foundation:1258896144 {source="MONDO:equivalentTo", source="Orphanet:145", source="https://orcid.org/0000-0002-4142-7153"}
```
joeflack4, line 61659: Not an expert on this part of things, but these kinda refs look like what I expected to see!

twhetzel (approved):
LGTM - all 1,255 items marked with 'OK' as the value in the "lex match" column or 'ok to add' or 'add with Mondo:obsolete" in the "manual review" column are included in the diff.

PR #7625, opened 2024-04-22
add PITX2 atrial fibrillation

lindenb:
Hi this is my first PR for this project, I manually edited mondo-edit.obo to add the relationship PITX2 <-> atrial fibrillation.  I hope everything is ok.

(later) and while I was there, I also added the relation HEY2<->Brugada https://pubmed.ncbi.nlm.nih.gov/28637782/


sabrinatoro:
@lindenb Thank you for creating this PR. 
I have carefully reviewed the annotations you provided. However, upon evaluation, it appears that they are not suitable for inclusion in Mondo.
Every annotations in Mondo need to be true in 100% of the cases. Therefore, to associate a gene with a disease, the disease must be caused by a variation in this gene in ALL cases. 

Upon examination, it looks like 'Brugada syndrome' and 'atrial fibrillation' can be caused by genetic mutations in multiple genes. Therefore, the gene annotations you proposed for inclusion in the ontology cannot be accepted.

However, these gene-to-disease annotations are extremely useful and would be more appropriately incorporated into the Monarch Knowledge Graph. We are actively exploring avenues to facilitate direct submissions; however, no such mechanism is currently available.
 
Please let me know if you have any questions, and please do not hesitate to suggest other changes. Your contributions are appreciated.

PR #7626, opened 2024-04-22
Update qc-xref-syntax.sparql

matentzn:
See https://github.com/ontodev/robot/pull/1186

This replaces the old permissive xref check by the new ROBOT invalid xref which is not yet released.


twhetzel (approved):
lgtm

PR #7629, opened 2024-04-23
ICD11foundation x-ref

sabrinatoro:
Added all the ICD11foundation x-ref that were reported in the exact lexical matching file.
File is here: https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/main/src/ontology/lexmatch/unmapped_icd11foundation_lex_exact.tsv


sabrinatoro:
FYI: @matentzn @joeflack4 @twhetzel

joeflack4 (commented):
Interesting.. not sure how to account for illegal proxy merges coming in through lexmatch. Forgive my ignorance on this part of the pipeline, but aren't the lexmatch matches just suggestions? If what they suggest is an illegal proxy merge, then would would not take a look at them and probably discard those as invalid, non-matches?

I see you've uploaded the [lexmatch file as a google sheet](https://docs.google.com/spreadsheets/d/1-mFEuwnm-WIbCJDUJF85F69UHRxdpqeRuMvgK7jo43c/edit#gid=553130211), and have a tab for "proxymerge review".

Just adding some additional investigation into this. I took a look at the original MONDO->ORDO->ICD11 SSSOM that I created, and I did find some illegal proxy merges there as well... see [MONDO->ORDO->ICD11 SSSOM illegal proxies google sheet](https://docs.google.com/spreadsheets/d/1npe1PdZrb0YTtAbPqgu7OZ961g7KwC0_eZLAlK7RCEc/edit#gid=507491316). There are 3 MONDO IDs there which have duplicates / illegal proxy merges: MONDO:0005823, MONDO:0013626, and MONDO:0019171. Of these, I only see that MONDO:0019171 appears in Sabrina's lexmatch google sheet tab 'proxymerge review'. But when I look at the tab with the actual lexmatch results, I don't see that ID listed.

sabrinatoro:
@joeflack4 I think it is part of the review. The lex matching is a suggestion, and we rarely realize that there is a proxy merge until the QC checks tell us. 
As we move forward, this will happen less often because the ICD11 code will not be suggested anymore (since it is already in Mondo). 
The very good thing (even though it is annoying to review-but it is my job :-) ) is that it unveil issues about the orphanet-ICD11 mappings. So it is all fine.

cmungall:
Use OAK to validate mappings rather than rely on waiting for a GH action

https://github.com/INCATools/ontology-access-kit/blob/main/notebooks/Commands/ValidateMappings.ipynb

joeflack4:
@cmungall @matentzn @twhetzel If OAK has means to validate illegal proxy matches, shouldn't `sssom-py` also? I just checked and it did not detect.

I guess we could validate lexmatch files in `mondo-ingest`, but we'd have to do some refactoring. They're like half robot template / half SSSOM.

@cmungall I finagled the lexmatch file until it passed `sssom-py` validation, but when I ran `runoak validate-mappings` on it, I got a not-validation related error:
- https://github.com/INCATools/ontology-access-kit/issues/742

twhetzel:
@sabrinatoro I am looking over the mondo-edit.obo file, one question is why does `MONDO:0009255` have both of these lines:
```
xref: icd11.foundation:1173858031 {source="MONDO:equivalentTo"}
xref: icd11.foundation:1173858031 {source="Orphanet:79237", source="MONDO:equivalentTo"}
```

sabrinatoro:
This is correct, I think these lines are duplicated because they were added via the orphanet-icd11 mappings and via the exact lexical matching. 
I agree, these lines should be merged. I wonder if a normalization step would do the trick...

twhetzel:
I don't know if `NORM` will do that, but worth a try. If not, my next try would be a SPARQL query to find these and manually remove.

matentzn:
Definitely normalise this it will do the trick

twhetzel (changes requested):
I addition to needing to run `NORM` to collapse the cases where the same Mondo ID has xrefs to the same source and source id, e.g.  `xref: icd11.foundation:1173858031 {source="MONDO:equivalentTo"} and xref: icd11.foundation:1173858031 {source="Orphanet:79237", source="MONDO:equivalentTo"}` there are 85 cases where the same Mondo ID is mapped to two _different_ ICD11 terms. I was not expecting terms to be mapped to  >1 ICD11 term given the workflow to add the ICD11 mappings gathered from Orphanet first and then do the lexical alignment.

I reviewed the cases where the Mondo term was mapped to >1 ICD11 term and added comments [here](https://docs.google.com/spreadsheets/d/1hcgoF7GHUTkRDz-fktsEvVZR-iIHMTOfqsBE9tgNUs4/edit?usp=sharing) in the Sheet "Curated Dups". 

The general cases are:
- one of the ICD11 xRefs is in the Extension Codes branch --> I'm not sure if both xRefs to ICD11 terms in the Extension Code branch should be included. These are marked in the column "Correct xRef" as `Both match` and the "Notes" column indicates which xRef is in the Extension Codes branch
- there seem to be a few xRefs that are mis-mapped based on the ICD11 structure --> These are marked with what should be the correct mapping in the "Correct xRef" column
- a few that need further review due to how the Mondo term is labeled, defined, and/or the synonyms it has --> These are marked in the "Correct xRef" column as `Needs Review`

I used the [WHO-FIC Maintenance Platform](https://icd.who.int/dev11#/http://id.who.int/icd/entity/162810284) to look up these ICD11 codes. However since this site includes unreleased, work in progress development and not all ICD11 codes mapped to Mondo in this PR can be found in the [ICD11 browser](https://icd.who.int/en), e.g. `249843059` which is mapped to MONDO:0016691, pilocytic astrocytoma along with `424726772` also being mapped to this Mondo term, should mappings to ICD11 that are not found in the main released version be included, will these ICD11 codes eventually be in the released version, should we have used a different source file for ICD11?

matentzn:
+1 on most definitely adding 1 or more orcids

> there seem to be a few xRefs that are mis-mapped based on the ICD11 structure

This is a very complex issue. We treat structure merely as evidence for the correctness of a mapping. The goal of determining exactness is to determine "intent" (whether or not explicitly reflected in the structure of the ontology) and approach mapping from a "practical POV" (nothing is truly an exact match, but if it "appears the same" and pretending it is the same does impact diagnostic tools adversely, we assert it).

I would break out the 85, make a new PR for these and otherwise merge it, but I understand from slack that you prefer to be more careful.

src/ontology/mondo-edit.obo
matentzn, line 1: I would advice to try to avoid commenting on Pull requests directly, instead, comment on _files or lines_ in the pull request _Changes_, because its very difficult otherwise to have isolated discussions (this PR alone has 5 things we would have to discuss seperately like: how to use oak for automated mapping review? What do to about proxy merges? etc etc). And of course: this makes it easer to resolve them.

> Just adding some additional investigation into this. I took a look at the original MONDO->ORDO->ICD11 SSSOM that I created, and I did find some illegal proxy merges there as well... see [MONDO->ORDO->ICD11 SSSOM illegal proxies google sheet](https://docs.google.com/spreadsheets/d/1npe1PdZrb0YTtAbPqgu7OZ961g7KwC0_eZLAlK7RCEc/edit#gid=507491316). There are 3 MONDO IDs there which have duplicates / illegal proxy merges: MONDO:0005823, MONDO:0013626, and MONDO:0019171. Of these, I only see that MONDO:0019171 appears in Sabrina's lexmatch google sheet tab 'proxymerge review'. But when I look at the tab with the actual lexmatch results, I don't see that ID listed.

"Illegal" is where one external id (foundation, ORDO) maps to two distinct Mondo ids. The way I see your spreadsheet examples, it is the other way around, i.e. Two external, one Mondo? That is ok (if it is correct of course), as it just means that we consider two external ids the same.

Generally, as you are correctly stating, lexmatch generates _canidate matches_ - not mappings. It is the job of the reviewer to sift through these as there is no way to tell from a proxy merge, with 100% certainty, which of the the candidates should be dropped. A combination of `oak review-mappings` and `boomer` would be able to resolve these with 90% certainty maybe, and we may decide that this is enough. But as long as our quality threshold for a resource is "human review", in my view, we should not use tools to drop candidates before they even get here (into the PR). We can, however, decide that for some sources, we dispense with human review altogether and let it all be done by LLM and boomer. This as the advantage of much greater coverage.

joeflack4: > I would advice to try to avoid commenting on Pull requests directly, instead, comment on files or lines in the pull request Changes

I agree. But how did you do that? When I go to "Files changed", there is only `mondo-edit.obo`, which has too many changes to display the diff. So there is no line that I can click on to create a thread. How did you create a thread here? Can you screenshot what you're doing?

joeflack4: You're right that these are legal; not proxy merges. Ugh, I swear, I like double or triple checked these before commenting. It's like how people get confused by their left versus their right or something. Sorry for wasting your time on this.

I am glad to learn about how the lexmatch candidate matches have been set up to be handled, though. That makes sense; thanks for explaining for me.

matentzn: You can comment of the file: 

image

Re rest, I will say something on slack.

Issue #5549, opened 2022-11-01
OBO Academy exercise: add axioms


sabrinatoro:
- add an Equivalent To 
- add SubClass of

**NOT a real example!!!!!!!** 
a variation in the gene KCNJ10 was found to be responsible for cone-rod dystrophy 8 (MONDO:0011564). 

**!!!! This issue is for ontology training only !!!!!!**

sabrinatoro:
Remember to 
- create a new branch
- create a PR (after you made changes)
- refer to this issue in your PR
- tag Sabrina (or someone else) to review your work

sabrinatoro:
For demo/teaching only. Closing.

PR #7635, opened 2024-04-25
Added subclass of to 'avian leukosis'

ehurwitz:
Address #5549


ehurwitz:
Just for training

sabrinatoro:
It looks good @ehurwitz. Good job!

PR #7677, opened 2024-05-03
Remove unnecessary check from Mondo

matentzn:
This check does not do any thing, i.e. the functional part is commented out (it is shaving off maybe 30 seconds from the overall runtime)


twhetzel (approved):
ok

PR #7683, opened 2024-05-05
Remove UMLS CN xrefs

### LINKED ISSUE #7035: Remove all UMLS:CN* cross references from Mondo
I advocate to remove all UMLS:CN* cross references from Mondo as they falsely suggest that they are UMLS cross references - they are not! Neither will they ever resolve anywhere. UMLS:CN* cross-references, correct me if I am wrong @kanems, are temporary UMLS CUIs created by the MedGen team to facilitate curation efforts in cases where UMLS is missing a disease. These terms may later be replaced by proper UMLS CUIs.  

See question by @twhetzel in https://github.com/monarch-initiative/mondo/issues/6560#issuecomment-1856065251_

What I would like to do instead is:

- [ ] Properly reference all medgen cross-references from Mondo (for linkouts, i.e. MEDGEN:123)
- [X] Discuss if there is value at all in curating MEDGEN_UMLS_CUI:CN* references from Medgen in Mondo. @kanems, what do you think? Is there a reason why the general public may like `MEDGEN_UMLS_CUI:CN*` cross references?

Related

- [ ] https://github.com/monarch-initiative/medgen/issues/15
- [x] https://github.com/monarch-initiative/mondo-ingest/issues/366

matentzn:
fixes #7035

In this PR, I simply removed all `UMLS:CN*` references in mondo-edit.obo. As a reminder:

1. These are confusing, as they have nothing to do with UMLS and are indeed temporary identifiers created by MedGen
2. These are redundant, as we also have the MedGen UID references


twhetzel (approved):
lgtm, thanks for the context for this PR

Issue #7682, opened 2024-05-04
Review all subsets in Mondo


matentzn:
I have to sort out the subsets pipelines in Q2. This led me to do a quick review on all of them:

https://docs.google.com/spreadsheets/d/1EbZEqRxtUcKD8rM6T_m10HGjB1DVXAw2NoWj7fEtDH8/edit#gid=0

I would suggest 
- @twhetzel and @sabrinatoro to look through these and add their own comments
- discuss any disagreements and uncertainties in a Tech call
- Execute the recommendations

PR #7678, opened 2024-05-03
Update ORDO subsets

matentzn:
Based on @joeflack4 and @twhetzel work in https://github.com/monarch-initiative/mondo-ingest/pull/510

This commit updates the ORDO subsets in Mondo which were outdated for years. They are needed for the new ORDO RD distiniction.

PR #7695, opened 2024-05-07
Update rare disease subset

matentzn:
Changes:

- 23 added and 17 removed GARD rare terms  and mappings (not sure where they come from)
- 617 added and 398 lost inferred terms in RD subset
- 2 added NORD rare terms, 4 moved, 10 updated preferred labels from NORD
- 337 added and 948 removed terms to RD subset, 
- 1 removed orphanet RD term


sabrinatoro:
@matentzn 
- Can you confirm that you used the "old" file that gard gave us some time ago? (ie not a new file)
- Most GARD x-ref in Mondo are not "equivalentTo". I think we need to update the ontology accordingly using the file from GARD (let's wait for a more recent one).
- all the "lost" GARD rare terms go hand in hand with "new" one. These are from term merged. Weirdly enough, the lost one is associated with the term that we kept, and the "new" on is associated with the obsoleted one (ie the term merged into the other one). I think this has to do with the fact that the file from GARD doesn't take into consideration the updates in the ontology (e.g. term merged).
- the failed QC: qc-animal-disease-rare.sparql is also because the files from NORD/GARD did not get updated (we could fix manually for this release)
- the failed QC: FAIL Rule ../sparql/qc/general/qc-related-exact-synonym.sparql are all coming from NORD. I think (based on my recollection from previous review) that it comes from Nord saying it should be an exact synonym, and me having changed these to "related" manually.

Generally, we need to put some workflow in place to make sure that the rare authorities take Mondo changes into account in their lists (and/or that we put something in place to check on these when we update the subsets). This should be for the next release (or the one after).

sabrinatoro:
NORD:
was this update based on a new file from NORD?
There are a lot of terms that Gioconda requested to be in the nord-rare subset and I don't see them there. This is probably an error on their part.

twhetzel:
I see `subset: ordo_disease {source="Orphanet:377788"}` and `subset: orphanet_rare {source="Orphanet:377788"}` on Mondo disease (MONDO:0000001) were removed since the `ordo_rare` PR, which fits with the comment in the [Create ORDO subset ROBOT template](https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1592954704). 

The counts still match since the [Update ORDO subset PR](https://github.com/monarch-initiative/mondo/pull/7678) for `ordo_group_of_disorders` and `ordo_clinical_subtype`, but not for `ordo_disease`, which is now at 6248 (MONDO:0000001 did not have subsets in that PR).

I don't know how the other subsets were generated so can't comment on those. @sabrinatoro can you also review?

matentzn:
> Can you confirm that you used the "old" file that gard gave us some time ago? (ie not a new file)

Yes the old file - but combined with the "new mappings". 

> Most GARD x-ref in Mondo are not "equivalentTo". I think we need to update the ontology accordingly using the file from GARD (let's wait for a more recent one).

This is unlikely as GARD only gives us OMIM and ORDO ids - which we map with their "exact" counterparts in Mondo?

> all the "lost" GARD rare terms go hand in hand with "new" one. These are from term merged. Weirdly enough, the lost one is associated with the term that we kept, and the "new" on is associated with the obsoleted one (ie the term merged into the other one). I think this has to do with the fact that the file from GARD doesn't take into consideration the updates in the ontology (e.g. term merged).

> the failed QC: qc-animal-disease-rare.sparql is also because the files from NORD/GARD did not get updated (we could fix manually for this release)

> the failed QC: FAIL Rule ../sparql/qc/general/qc-related-exact-synonym.sparql are all coming from NORD. I think (based on my recollection from previous review) that it comes from Nord saying it should be an exact synonym, and me having changed these to "related" manually.

I will re-run this without updating GARD and NORD now - these are not so critical.

> Generally, we need to put some workflow in place to make sure that the rare authorities take Mondo changes into account in their lists (and/or that we put something in place to check on these when we update the subsets). This should be for the next release (or the one after).

Lets do it after the release! Thank you for the review!

> I see subset: ordo_disease {source="Orphanet:377788"} and subset: orphanet_rare {source="Orphanet:377788"} on Mondo disease (MONDO:0000001) were removed since the ordo_rare PR, which fits with the comment in the https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1592954704.

Yes, exactly!

> The counts still match since the https://github.com/monarch-initiative/mondo/pull/7678 for ordo_group_of_disorders and ordo_clinical_subtype, but not for ordo_disease, which is now at 6248 (MONDO:0000001 did not have subsets in that PR).

This is expected. I want to have a serious discussion about #7682 and add QC for absolutely everything moving forward!


Thank you for the review! I will create a new PR, omitting updates to GARD and NORD now to only have the bare minimum needed for the release.

matentzn:
Closed in favour of #7696

PR #7696, opened 2024-05-08
Update RD subset in Mondo (Attempt 2)

matentzn:
Second attempt supplanting #7695.

This PR omits updating GARD and NORD which need to be coordinated with the teams.

Changes in this PR:

+subset: inferred_rare: 612
+subset: rare: 335
-subset: inferred_rare: 398
-subset: ordo_disease {source="Orphanet:377788"}
-subset: orphanet_rare {source="Orphanet:377788"}
-subset: rare: 943


sabrinatoro:
There are 117 mondo term in the ordo-disease subset that are not in the ordo-rare subset. All these are obsolete in Mondo, and therefore this is expected that they are not in the rare subset. 
I think it looks good to me. 

@twhetzel do you want to also review?

sabrinatoro (approved):

twhetzel:
I'm most familiar with the ordo subsets so can only comment on those. There does seem to be a change between PRs from [Update rare disease subset](https://github.com/monarch-initiative/mondo/pull/7695), the counts still match since the https://github.com/monarch-initiative/mondo/pull/7678 for `ordo_group_of_disorders` and `ordo_clinical_subtype`, but not for `ordo_disease`, which is now at 6248 (MONDO:0000001 did not have subsets in that PR). I think this was 6251 in [Update rare disease subset](https://github.com/monarch-initiative/mondo/pull/7695).

sabrinatoro:
@twhetzel Thank you for checking. 
Since the discrepancy is only 3 diseases, I suggest that we merge and go ahead with the release. 
It would be helpful to know what these 3 diseases are though, so we can investigate what is going on.

PR #7699, opened 2024-05-08
May release

sabrinatoro:


imports/merged_import.owl
matentzn, line 1: Generally I would favour decoupling imports from releases.

src/ontology/mappings/mondo.sssom.tsv
matentzn, line 1: Most of these changes due to UMLS:CN* removal, as expected.

src/ontology/mappings/mondo_closematch_doid.sssom.tsv
matentzn, line 1: The deletion of these files is expected because we have decided to not publish non-exact mappings to DOID, ordo and OMIM.

src/ontology/mappings/mondo_hasdbxref_efo.sssom.tsv
matentzn, line 1: lots of changes: expected due to updated in EFO subset import

matentzn (approved):
I would have liked to have mondo.obo in this diff.. This is the most useful file to review!

For the rest, looks good!

imports/merged_import.owl
sabrinatoro: I ran `IMAGE=odkfull:dev ./run.sh make IMP=false all -B`
therefore the imports should have been decoupled... ??

Issue #7579, opened 2024-04-10
Possible issue | non-MONDO term in is_a clause


Odessit007:
I see in the latest mondo.obo PATO:0000001 being is_a term for MONDO:0021125. Is this intended or a bug because several tools that I use t parse OBO files fail on this so I decided to check. Thank you in advance for clarification.

```
[Term]
id: MONDO:0021125
name: disease characteristic
def: "An attribute of a disease." [https://orcid.org/0000-0002-6601-2165]
synonym: "disease qualifier" EXACT []
synonym: "modifier" EXACT [NCIT:C41009]
synonym: "qualifier" EXACT [NCIT:C41009]
xref: NCIT:C41009 {source="MONDO:equivalentTo"}
is_a: PATO:0000001
property_value: exactMatch NCIT:C41009
```

It's something new in the latest release. The previous one looked like this:
```
[Term]
id: MONDO:0021125
name: disease characteristic
def: "An attribute of a disease." [MONDO:cjm]
synonym: "disease qualifier" EXACT []
synonym: "modifier" EXACT [NCIT:C41009]
synonym: "qualifier" EXACT [NCIT:C41009]
xref: NCIT:C41009 {source="MONDO:equivalentTo"}
property_value: exactMatch NCIT:C41009
```

sagehrke:
Dear @Odessit007, 
Thank you for reaching out to the Mondo Disease Ontology, part of the [Monarch Initiative](https://monarchinitiative.org/)!

We have received your ticket and will review it as soon as possible. 

If you have any additional information or updates relevant to your inquiry, please add them in a comment on this ticket. Your input is valuable in helping us understand and resolve the issue effectively.

Thank you for your patience and for your contributions to Mondo! 

Sincerely,
The Mondo Team

sabrinatoro:
@matentzn can you please look into this? 
In our newest release, we also have that "disease characteristic" (MONDO:0021125) is a "quality" (PATO:0000001).
Thank you!

matentzn:
Thank you @Odessit007 for reaching out!

This is what happens:

"disease charactertistic" is, indeed, a subclass of "pato quality" (so this is not a "bug" per se). Here is a snippet from the editors file:

```
[Term]
id: MONDO:0021125
name: disease characteristic
def: "An attribute of a disease." [https://orcid.org/0000-0002-6601-2165]
synonym: "disease qualifier" EXACT []
synonym: "modifier" EXACT [NCIT:C41009]
[...]

PR #7711, opened 2024-05-14
Add test for OAK parsing and OBO validation of release product

matentzn:
Precludes problems like #7579 in the future.

This PR:

1. Runs a OBO validator over the _release product_ mondo.obo rather than just over the edit file (which is more important for users!)
2. Run a test parse with OAK to ensure that the primary OBO release is OAK pronto parseable,.


twhetzel:
This does not pass the QC checks

PR #7716, opened 2024-05-16
refresh imports

twhetzel:
Followed the new (almost ready to be merged refresh import docs, https://github.com/monarch-initiative/mondo/pull/7492/files) to refresh the imports to add:
```
http://identifiers.org/hgnc/17985
http://purl.obolibrary.org/obo/NCBITaxon_3052342
http://purl.obolibrary.org/obo/NCBITaxon_3052343
```

### DIFF
--- a/src/ontology/imports/hgnc_terms.txt
+++ b/src/ontology/imports/hgnc_terms.txt
@@ -1115,6 +1115,7 @@ http://identifiers.org/hgnc/17966
 http://identifiers.org/hgnc/17967
 http://identifiers.org/hgnc/17975
 http://identifiers.org/hgnc/17978
+http://identifiers.org/hgnc/17985
 http://identifiers.org/hgnc/17989
 http://identifiers.org/hgnc/17993
 http://identifiers.org/hgnc/17994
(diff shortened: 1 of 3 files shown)


sabrinatoro (approved):
@twhetzel 
I think you didn't do the following step:
 Open `mondo-edit.obo` in Protege and save the ontology (as .obo).
 
 I just did and now the QC are running.
I am approving this PR because the rest is as expected. We can merge it as soon as the QC passed.

twhetzel:
@sabrinatoro I did that step, `Open mondo-edit.obo in Protege and save the ontology (as .obo).` and verified the gene I added was found in Protege.

sabrinatoro:
@twhetzel 
weird... because when I did it, there were some changes in the mondo-edit.obo (as shown in the diff).
We should keep an eye on that!

twhetzel:
@katiermullen I already created a PR to update the taxon subset file.

@sabrinatoro I think that step should just be part of the responsibility of whoever refreshes the import. Do you agree?

sabrinatoro:
@twhetzel 
Are you talking about the step of adding the NCBITaxon to the obo file ?
I don't have a strong opinion on that (I never thought about it until now). 
But yes, we could make it that it is the responsibility of whoever refreshes the import.

PR #7724, opened 2024-05-20
Update Protégé instructions.

gouttegd:
Current instructions about how to configure Protégé’s memory settings are out of date:

* in most cases editors using Protégé 5.6 should _not_ need to tweak its memory settings;
* if/when they need to do so, the recommended way with Protégé 5.6 is to use the `~/.Protege/conf/jvm.conf` file (which works across all platforms), rather than editing the (MacOS-specific) `Info.plist` file.


twhetzel (approved):
Thanks @gouttegd - this looks good to me.

PR #7731, opened 2024-05-22
Update NORD subset

matentzn:
The source files to this change can be found in this Draft PR:
https://github.com/monarch-initiative/mondo-ingest/pull/535

I didnt want to wait to merge it (as this file should automatically update when the mondo ingest pipeline is run).

Anyways, small changes, and good to move forward on this!


sabrinatoro:
@matentzn The QC fails are because NORD says a synonym is exact and we say a synonym is not exact. 
I can see that I have updated these manually previously. Should I update the mondo-edit file manually for now (until we set up something else)?

matentzn:
You can fix it manually for now - this will happen once per month, so it is not thaaaat bad, but annoying. Ideally we report the conflict to NORD

sabrinatoro (approved):
Approved. 
Waiting on the QC checks to pass after review of the synonym scopes

Issue #7033, opened 2023-12-13
Add KGX TSV as an official release format for the MONDO release pipeline


bryanlaraway:
As I'm working on rare disease projects within N3C, I'm looking to get MONDO ingested into that platform as part of an automated method for identifying rare disease patients based on the MONDO links to various diagnostic codes (ICD9/ICD10/SNOMED/etc). However, N3C needed the MONDO data in CSV/TSV format. The KGX TSV format found [here](https://kg-hub.berkeleybop.io/kg-obo/mondo/) will work, but it may not be up to date with the latest release of MONDO.

Could we add a KGX TSV format as part of the official MONDO release pipeline, to stay up to date with MONDO releases?

PR #7736, opened 2024-05-23
Add KGX pipeline and simple release to Mondo release pipeline

### LINKED ISSUE #7730: Add kgx as new release artefact to Mondo
@matentzn - Please provide rare subset releases (edges and nodes) as a KGX dumps in TSV format if possible. For example - https://www.dropbox.com/scl/fi/pkjp8k77pwt9m6z4evdcc/mondo-rare.kgx_edges.tsv?rlkey=5kv8cq99njza7917ecvvqr5px&e=1&dl=0

matentzn:
Fixes #7730 #7033 

We only add KGX for the subsets and main release pipeline here, not to the base or minimal releases. This directly addresses the above issues, but maybe more is needed in the future.

- [ ] A good idea would be to run a test release


twhetzel:
@matentzn - did you run a test release?

src/ontology/Makefile
```diff
@@ -244,8 +244,7 @@ reasoned-plus-equivalents.owl: reasoned.owl imports/equivalencies.owl
 $(ONT).owl: reasoned.owl
 	$(ROBOT) merge -i $< annotate -V $(ONTURI)/releases/`date +%Y-%m-%d`/$(ONT).owl -o $@
 
-# obo is self-contained
-$(ONT).obo: $(ONT)-base.owl
+$(ONT).obo: $(ONT).owl
```
twhetzel, line 247: What is `$(ONT).owl` vs. `$(ONT)-base.owl`?

src/ontology/Makefile
```diff
@@ -256,6 +255,9 @@ $(ONT).json: $(ONT).owl
 	$(ROBOT) convert --input $< --check false -f json -o $@.tmp.json &&\
 		mv $@.tmp.json $@
 
+$(ONT)_nodes.tsv $(ONT)_edges.tsv: $(ONT).json
+	kgx transform --input-format obojson --output-format tsv --output $(ONT) $< && test -f $(ONT)_nodes.tsv
```
twhetzel, line 259: Is a check that `$(ONT)_edges.tsv` exists also needed?

src/ontology/Makefile
```diff
@@ -264,6 +266,43 @@ $(ONT)-with-equivalents.obo: $(ONT)-with-equivalents.owl
...
+# drop every axiom: filter --term-file keep_terms.txt --trim true
+#	remove --select imports --trim false
+# NOTE FOR FUTURE: this is slightly different from the usual simple as it completely removes all
+# Object properties from the ontology (remove --select object-properties).
+$(ONT)-simple.owl: reasoned.owl $(OTHER_SRC) $(SIMPLESEED) $(IMPORT_FILES)
+	$(ROBOT) merge --input $< \
+		reason --equivalent-classes-allowed none --exclude-tautologies true --annotate-inferred-axioms true \
+		relax \
+		remove --axioms equivalent \
+		remove --select object-properties \
```
twhetzel, line 292: why are object properties being removed for "simple"?

twhetzel (approved):
looks good to me, a few comments just to help me understand things better

@matentzn - did you run a test release with this change?

src/ontology/Makefile
matentzn: $(ONT).owl (mondo.owl) corresponds to the full ontology you see in Protege - with all the imports. "base" has the imports stripped out.

src/ontology/Makefile
matentzn: I dont think so, if `kgx transform` produces the nodes file, it will produce the edges file.

src/ontology/Makefile
matentzn: This was the decision we made: the simple release product should only be the class hierarchy, nothing else. In particular: no relationships, such as disease characteristics and similar.

matentzn:
Yes! I did run a test release! Thank you for checking!

PR #7752, opened 2024-05-29
Update GARD subset

matentzn:
@sabrinatoro for context, see email Eric sent today!


sabrinatoro (approved):
this looks good. 
I noticed that some Mondo terms were on the rare-gard subset, but do not have an x-ref. This is because GARD told us these terms should be in the rare subset, and they will (at one point) create a new GARD entry.

PR #7760, opened 2024-05-30
Add NANDO cross references

matentzn:
Instead of making them exact, as discussed, just add them as xrefs for this version.


matentzn:
Mistake in QC, fixed in https://github.com/monarch-initiative/mondo/pull/7761

PR #7767, opened 2024-05-31
Update the external content subset

matentzn:
Updating a number of externally managed content pieces in Mondo, including the rare subset and otar managed content and more.

The large number of removed "inferred_rare" cases are due to the updated gard subset.


twhetzel:
Without the input data used for the changes, I don't see how this can be checked in any detail. 
@sabrinatoro do you have insight into these content changes and can review?

sabrinatoro (approved):
It looks good to me. 
For future update, could we maybe add some statistics (e.g. how many terms were added/removed from each subset) to support the review? I will create an issue.

PR #7781, opened 2024-06-05
Update ORDO subsets in Mondo edit

matentzn:
Based on the new Mondo Ingest


sabrinatoro (approved):
ordo_disease and ordo_clinical_subtype subsets were removed as discussed

PR #7788, opened 2024-06-06
Update the Orphanet, Rare and Rare inferred subsets

matentzn:
Automatic update of Orphanet, Rare Inferred and Rare subset, changes to which are based on the the updated ordo_disorder subset.


sabrinatoro (approved):
This looks good to me.

PR #7789, opened 2024-06-06
updated definition for "adult Fanconi syndrome"

### LINKED ISSUE #7323: Adult Fanconi syndrome should be merged with "acquired" FS
From @pnrobinson 

> adult Fanconi syndrome --- needs to be merged with "acquired" FS!!!

sabrinatoro:
Closes #7323


twhetzel (commented):
Based on the [ticket comment](https://github.com/monarch-initiative/mondo/issues/7323#issuecomment-2153468475) I thought the `xref` to "NCIT:C4377" would be removed also.

There are also synonyms that use NCIT:C4377 as the source so that seems awkward.

There was an additional [user comment](https://github.com/monarch-initiative/mondo/issues/7323#issuecomment-21537403410 in the ticket I'm not sure if you want to address here in this PR.

sabrinatoro:
@twhetzel 
When the definition from NCIT:C4377 is misleading, it still refers to the concept of "adult Fanconi Syndrome". Therefore I kept it as an x-ref.

While the comment from the other user is interesting, it is not relevant to this issue: the concept's definition is that "adult fanconi syndrome is a fanconi syndrome that occurs in adult". It is most often acquired, but can sometimes be hereditary. The omim record (if it is confirmed to be for adult, which I haven't investigated) would be for "hereditary".

twhetzel (approved):

Issue #7071, opened 2023-12-21
How to represent linkouts vs cross-references vs mappings in Mondo


matentzn:
In the huddle this week we talked a bit about the differences between mappings, cross-references and linkouts.

<img width="1246" alt="image" src="https://github.com/monarch-initiative/mondo/assets/7070631/4994475f-26f4-4035-b154-9b9d0c7ee1eb">

The time has come to develop an air-tight datamodel for this, and align it with biolink, cc @sierra-moxon.

Right now, we have roughly these structures:

1. Mappings:

Mappings are shipped using skos:exactMatch.

```
property_value: exactMatch DOID:4
```

Things to consider:

1. We do not currently publish the mapping inside the ontology with SSSOM metadata, but we probably could
2. We redundantly publish "mappings" using xrefs (see "Cross-references" below) and "skos:exactMatch") which may be good (German saying: [Doppelt haelt besser](https://de.wiktionary.org/wiki/doppelt_h%C3%A4lt_besser), but could result in confusions "which to use". 

2. Cross-references

Cross-references use the xref property in OBO format, and then a list of "source" properties to denote "confirmatory evidence" (who said this?) and custom provenance like "what semantic precision is this?". The fact that both of these use the same property ("source") is a bit unfortunate tbh and will make it harder for UIs to display the "right provenance information". 

```
xref: UMLS:C4310752 {source="MONDO:equivalentTo", source="MONDO:ncbi_mim2gene_medline"}
```

3. Linkouts

Linkouts use "see the see also" as a property value, with xsd:anyURI annotation and a "source" property to denote the source.

```
property_value: seeAlso "https://rarediseases.info.nih.gov/diseases/10691/duane-syndrome-type-3" xsd:anyURI {source="GARD:0010691"}
```

@sabrinatoro was floating the idea of using a new property that is more specific, which is an option, but could prove difficult to align with other stacks.

This issue is purely about the representation of these three datatypes. For more on how to display these links see: https://github.com/monarch-initiative/monarch-app/issues/521.

cmungall:
I think your 3 is just axiom annotations? Aka provenance. Not sure we need
a new overloaded term for this

Linkouts already has a meaning eg ncbi linkouts. And do we have so many
seealsos?

matentzn:
[...]

PR #7792, opened 2024-06-11
Add curated content resource property

matentzn:
Right now, we cannot agree community wide on a standard way to represent linkouts, not even whether or not they should be curated in the ontology at all. This is breaking a bit more with Mondo being a pure ontology vs a bit databasey, but right now, we just need a convenient way to link to curated content for our stakeholders.

Here, we add the property so we can start using it, even if we might change the ID to an OMO id in the future.

Context:

- https://github.com/information-artifact-ontology/ontology-metadata/issues/165 
- https://github.com/monarch-initiative/mondo/issues/7071

### DIFF
--- a/src/sparql/qc/general/qc-permitted-properties.sparql
+++ b/src/sparql/qc/general/qc-permitted-properties.sparql
@@ -25,7 +25,7 @@ SELECT DISTINCT ?term ?property WHERE
 		dc:conformsTo,
 		dc:creator,
 		dce:date,
-		MONDO:0700275,
+		mondo:0700275,
 		mondo:excluded_from_qc_check,
 		mondo:excluded_subClassOf,
 		mondo:excluded_synonym,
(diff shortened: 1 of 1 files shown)


sabrinatoro (approved):

src/ontology/mondo-edit.obo
```diff
@@ -547007,6 +547007,14 @@ property_value: IAO:0000233 https://github.com/monarch-initiative/mondo/issues/5
 property_value: IAO:0000233 https://github.com/obophenotype/upheno/issues/871
 is_symmetric: true
 
+[Typedef]
+id: curated_content_resource
+name: curated content resource
+def: "A curated content resource is a web-resource that contains curated content about the ontology term." []
+comment: This property is likely to change the ID. Please do not use in production systems.
```
twhetzel, line 547014: What does `This property is likely to change the ID` mean? Change which ID?

matentzn: id: curated_content_resource

is likely to be swapped to an official OBO id, like 

```
OMO:0002005
```

Its just to painful to add these relations to OMO right now, and we want to move on a bit!

Issue #6085, opened 2023-03-17
create DOS-DP patterns for non-human diseases


sabrinatoro:
we need templates for non-human diseases. 
These templates would probably be the same as for the human counterpart.

sabrinatoro:
@cmungall and @matentzn I have some questions. Could you please advise? Thank you!
(If this is too long to discuss here, I could bring this up on a call.)

**Question 1:** 
- pattern 1: species-specific disease (e.g. dog diseases)
- pattern 2: infectious disease (e.g 'disease has primary infectious agent' some 'infectious agent')

Do we need a 3rd pattern for "infectious disease in a specific species"? or can we mix the 2 patterns? 


**Question 2:**
We already follow a pattern to relate non-human diseases to their human analogs ("cross-species analog" some 'human disease').
Should a non-human disease have 2 equivalents? for example: 
'infectious disease, non-human animal'
= 'non-human animal disease' and ('cross-species analog' some 'infectious disease')
= 'non-human animal disease' and ('disease has infectious agent' some root)

Note: I am not 100% sure that the "cross-species analog" relation is correct since I don't think it is 100% correct to be passed to the children terms, but this is a different conversation for another issue/time.

matentzn:
> Note: I am not 100% sure that the "cross-species analog" relation is correct since I don't think it is 100% correct to be passed to the children terms, but this is a different conversation for another issue/time.

I think you are right, and I would like to suggest to delete this relationship right after the initial classification is computed. We should not expose the "cross-species analog" relationship to our users, instead, we can use the relationship to classify the diseases and then use sparql to simply add regular mapping predicates to the release files (annotation level) which do not inherit down. If you raise this issue we can solve it in about 20 minutes.

For question 1, the lazy main answer is: generally we only need one pattern (pattern 1) for all diseases. This assumes that there is always an infectious disease counterpart in the human branch that can be used. My assumption is that for your use case, there is not; so you want to add a non-human disease which does not have a human counterpart?

This is best discussed in a call.

sabrinatoro:
From my 1:1 with Nico: 
- we would use the "cross-species analog" relation to get the classification, but we would remove it as part of the release process (ie we would keep the classification but not the relation which would (and should not) get passed to the children)

- RE: DOS-DP pattern: 
    - for all the **non-human diseases that have a "cross-species analog"**: 
          - the pattern should be to use the "cross-species analog" axiom
[...]

PR #7794, opened 2024-06-11
nonhuman genetic disease pattern

sabrinatoro:
Addresses #6085

I also updated the existing patterns for genetic diseases in human to point to the specific "human disease" and not general "disease"

### DIFF
--- a/src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
+++ b/src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
@@ -1,20 +1,15 @@
-pattern_name: disease_series_by_gene
+pattern_name: nonhuman_disease_series_by_gene
 
-pattern_iri: http://purl.obolibrary.org/obo/mondo/patterns/disease_series_by_gene.yaml
+pattern_iri: http://purl.obolibrary.org/obo/mondo/patterns/nonhuman_disease_series_by_gene.yaml
 
-description: 'This pattern is for diseases that are caused by a single variation or mutation in
-  a single gene, that have gene-based names, such as new disease terms that are requested
-  by ClinGen, like MED12-related intellectual disability syndrome. See more details about [disease naming in Mondo here](https://mondo.monarchinitiative.org/pages/disease-naming/).  Examples: [MED12-related
-  intellectual disability syndrome](http://purl.obolibrary.org/obo/MONDO_0100000),
-  [TTN-related myopathy](http://purl.obolibrary.org/obo/MONDO_0100175), [MYPN-related
-  myopathy](http://purl.obolibrary.org/obo/MONDO_0015023)'
+description: 'This pattern is for non-human diseases that are defined based on a variation in one gene, such as many genetic diseases from the Online Mendelian Inheritance in Animals (OMIA).  Examples: TBD '
 
 contributors:
-- https://orcid.org/0000-0002-6601-2165
-- https://orcid.org/0000-0001-5208-3432
+- https://orcid.org/0000-0002-4142-7153
+- https://orcid.org/0000-0002-5002-8648
 
 classes:
-  disease: MONDO:0700096
+  disease: MONDO:0005583
   gene: SO:0001217
 
 relations:
--- a/src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml
+++ b/src/patterns/dosdp-patterns/OMIM_disease_series_by_gene.yaml
@@ -26,7 +26,7 @@ contributors:
 - https://orcid.org/0000-0002-4142-7153
 
 classes:
-  disease: MONDO:0000001
+  disease: MONDO:0700096
   gene: SO:0000704
 
 relations:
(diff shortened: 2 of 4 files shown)


src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - gene
+
+def:
+  text: Any %s in which the cause of the disease is a variation in the %s gene.
+  vars:
+  - disease
+  - gene
+
+equivalentTo:
+  text: '%s and ''has material basis in germline mutation in'' some %s'
```
sabrinatoro, line 56: @cmungall and @matentzn 
I did not add the taxon here because the taxon is included in the gene record. 
In your opinion, do you think we should add taxon information here? Is there a reason to add this redundancy? 
Thank you!

matentzn:
Best add to Tech call agenda

twhetzel:
Added, but the next Tech call is not until Friday, June 21

src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
cmungall: You need either

1. a property chain `'has material basis in germline mutation in' o 'in taxon' -> 'in taxon'`, in RO
2. the ensure the genus is already species-specific. Presumably this is not always the case as some PS analogs will be cross-mammal, cross-vertebrate, etc
3. just make a var for taxon even if it is duplicative with either/both genus/differentia

But going back to your Q from slack about infectious disease classification:

1. I would in general try and keep the ladders parallel. If some order/class/etc level virus is a useful grouping for H, it should also be for NH? But I appreciate this may not always be the case, esp for agents that never infect human, in which case
2. create a `non_human_disease_by_infectious_agent` DP
     - this would only use NH terms as genus
     - the taxon of the host would need to go in as a variable
     - classification would follow NCBITaxonomy
     - we can worry about making the xp-analog relations complete via rules later

If you do go with 2 it points to doing the same this for this PR, e.g. including a taxon var

sabrinatoro:
@katiermullen @matentzn @cmungall 
I updated the pattern to include the taxon information (decided to go the redundancy route).
Could one of you please review and approve? Thank you!

src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
```diff
@@ -22,38 +24,45 @@ annotationProperties:
 vars:
   disease: '''disease'''
   gene: '''gene'''
+  taxon: "'taxon'"
```
katiermullen, line 27: For my education - is this supposed to be '''taxon''' and not "'taxon'"?

katiermullen (approved):
@sabrinatoro I have one tiny comment. Can you please take a look? Otherwise, it looks great to me and can be merged.

src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+def:
+  text: Any %s, occuring in %s, in which the cause of the disease is a variation in the %s gene.
+  vars:
+  - disease
+  - taxon
+  - gene
+
+equivalentTo:
+  text: '%s and (''has material basis in germline mutation in'' some %s) and (''in taxon'' some %s)'
```
matentzn, line 64: Note here that there are two ways you can model this:

1. you can push the taxon into the gene. assuming you always use hgnc identifiers for humans and ncbigenes for animals, and you always use the correct gene id pertaining to a taxon specific gene, this is enough, no need for recording the taxon specifically to distinguish from human disease
2. however, you likely want the taxon to separate grouping classes "dog disease", which means having the taxon can be helpul (unless the genes in our ingest have them already!). You might also need this for non-gene non-human-animal diseases, so maybe just curate redundantly. If you do, however leave it like you have it, not that _technically_, you may want to add the human taxon to the human diseases. 

Long story short: its probably best to leave it has you have it, just adding some considerations in for future decision making.

matentzn (approved):
LGTM!

src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+description: 'This pattern is for non-human diseases that are defined based on a variation in one gene, such as many genetic diseases from the Online Mendelian Inheritance in Animals (OMIA).  Examples: TBD '
+
+contributors:
+- https://orcid.org/0000-0002-4142-7153
+- https://orcid.org/0000-0002-5002-8648
+
+classes:
+  disease: MONDO:0005583
+  gene: SO:0001217
```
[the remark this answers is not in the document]
sabrinatoro: Good call. I will also update the human pattern

src/patterns/dosdp-patterns/nonhuman_disease_series_by_gene.yaml
sabrinatoro: After a lot of discussion, we decided to keep the redundancy. But I agree, we might decide to remove this redundancy at a later time

PR #7795, opened 2024-06-11
Update medgen cross references

matentzn:
This updates
- all medgen references in Mondo
- all UMLS references in Mondo according to Medgen

@sabrinatoro this does not have to be done super fast. I can easily recreate it. But it is a beast - not sure how you want to approach it..


sabrinatoro:
@matentzn do we have a link to the file shared by MedGen, by chance? 
That would be helpful for review.

matentzn:
Yep, this is the one: https://ftp.ncbi.nlm.nih.gov/pub/medgen/MedGenIDMappings.txt.gz

twhetzel:
Adding my Slack comment here as well:
Were these all added as `source="MONDO:equivalentTo"`? We discussed changing one of these, but need to check notes which one. I think it was Mondo to UMLS. I'll have time later today to check my notes.

matentzn:
> Were these all added as source="MONDO:equivalentTo"? We discussed changing one of these, but need to check notes which one. I think it was Mondo to UMLS. I'll have time later today to check my notes.

Yes, we removed the equivalentTo from the mesh mappings, but the MONDO-MEDGEN-UMLS mappings are 1:1 according to MedGen, so it should be fine!

sabrinatoro:
All the QC errors are because 1MedGen/CUI = 2 Mondo term:
- Mondo term 1 MONDO:EquivalentTo
- Mondo term 2 = obsolete term that was merged to Mondo term 2 ; this term has already the source MONDO:obsoleteEquivalent

I removed the x-ref on "Mondo term 2" because that is the regular SOP when merging terms. We however need to review how to handle x-ref to obsoleted terms.

matentzn:
> I removed the x-ref on "Mondo term 2" because that is the regular SOP when merging terms. We however need to review how to handle x-ref to obsoleted terms. 

IMO we need to tell our collaborators to not map to obsolete classes, and internally filter our references to obsoleted classes..

sabrinatoro:
> IMO we need to tell our collaborators to not map to obsolete classes, and internally filter our references to obsoleted classes..

I agree, but we need to let them know. I created [this issue](https://github.com/monarch-initiative/mondo-ingest/issues/563) to make sure that we handle these cases appropriately and we get a report to share with external sources for their update of their mappings.

sabrinatoro (approved):

PR #7800, opened 2024-06-13
Update RD subset

matentzn:
This updates RD subset, we can still see a few issues, like

synonym: "Acanthocheilonemiasis" EXACT [NORD:705]
synonym: "Acanthocheilonemiasis" RELATED [GARD:0000004, NORD:705]

I guess we need to report any resulting problems.

@sabrinatoro good opportunity to further nail our SOP when we have constraint violations from externally managed content!


sabrinatoro (approved):
I took care of the QC errors (mostly related to synonym scope).
As soon as this has passed QC check, we can merge.

twhetzel:
How is the SOP when we have constraint violations from externally managed content coming along? Asking for my future self.

PR #7801, opened 2024-06-14
Draft 1: Update confirmed subclass of evidence in Mondo

matentzn:
This PR 

- Adds a workflow for updating confirmed subclass of evidence
- Updates mondo-edit.obo accordingly

### Issues of dispute

Against our previous agreement, the current version of the "confirmed" subclassOf axioms covers "Case 2 (SCR is direct in source, but indirect Mondo)". So, in other words, we include subclassOf axioms that are implied by Mondo (not asserted in the edit file!) and confirmed by an external source. 

Example:

```
[Term]
id: MONDO:0000266
name: pulmonary aspergilloma
def: "A aspergillosis that involves the lung." [MONDO:patterns/location]
...
intersection_of: MONDO:0005657 ! aspergillosis
intersection_of: disease_has_location UBERON:0002048 ! lung
```

results in this:

```
[Term]
id: MONDO:0000266
name: pulmonary aspergilloma
def: "A aspergillosis that involves the lung." [MONDO:patterns/location]
...
is_a: MONDO:0005657 {source="DOID:0050153"} ! aspergillosis
intersection_of: MONDO:0005657 ! aspergillosis
intersection_of: disease_has_location UBERON:0002048 ! lung
```

https://docs.google.com/document/d/1H8fJKiKD-L1tfS-2LJu8t0_2YXJ1PQJ_7Zkoj7Vf7xA/edit#heading=h.9hixairfgxa1

Chris would probably be very against this, and I must say, I am also against it, but weakly. Indeed, I am like 55% against it, and 45% think this _could_ be a good idea. The advantage is that we see _all the evidence_ as Mondo curators - every positive subsumption. When we make classification decisions in the future, we can see them all in the edit file (not just the non-redundant ones). Note - all of these are _implied_, so they are all correct. Today. 

The clear disadvantage is though that changing our logical definitions / patterns (equivalent class statements) will no longer be sufficient to change classification - if you really want to get rid of some subsumption moving forward, you must physically delete both the equivalence class axioms _and_ the asserted subclass axiom. Here is a particularly bad scenario: We decide a pattern no longer works and change it, which leads to changes in classification. The old classification will _not_ be removed, as the subclass of axioms have now been baked into the class hierarchy. I dont know though how much of a problem this is in reality.

I would suggest you 

1. Consult Chris on slack about his opinion on this. Its a big, PI-level, change.
2. If he says "no", we need to change the source for the "confirmed subclass pipeline" to "mondo-edit". Its not a huge deal, but may require some thinking.
3. If he says "yes", we still need to determine if there is a bug in this PR or not - I spot-checked 10 random examples and they are all correct (correct in the sense, mondo entails the subclass axiom).


sabrinatoro:
I don't mind the redundancy. In fact, I think it is helpful as it adds a level of confidence to the term classification (which is always good since Mondo integrates multiple sources).

We could consider removing this redundancy in the released version of the ontology (actually we might already do it), but I personally would keep it.

sabrinatoro (approved):
I reviewed a few changes in the mondo-edit.obo file, and these seems as expected. 
I found new source evidence on existing SC annotations, and new SC annotations+source.

@twhetzel, please take a look, especially for the changes in the subclass.yaml and mondo.Makefile files. Please merge if you approve. Thank you!

src/ontology/mondo.Makefile
```diff
@@ -963,6 +970,25 @@ update-%-mappings: $(TMPDIR)/new-exact-matches-%.owl
 		make NORM
 		mv NORM $(SRC)
 
+tmp/subclass-axioms.owl: $(SRC)
+	$(ROBOT) filter --input $(SRC) --axioms SubClassOf --preserve-structure false --trim false \
+		--drop-axiom-annotations "oboInOwl:source=~'(DOID|ICD10CM|NCIT|OMIM|OMIMPS|Orphanet):.*'" \
```
twhetzel, line 975: Is this the full list of components, e.g. does ICD11Foundation need to be added here?

matentzn: Whatever the correct list is, the question is good - it makes me wonder if we will ever remember to update this command here when new sources are added to Mondo ingest.

matentzn: For now it is correct though (I checked the [source](https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/main/src/ontology/reports/sync-subClassOf.confirmed.tsv), no icd11 in there at the moment), but we need to create another issue to ensure that we will update this list when new sources are added!

matentzn: Made an issue here: https://github.com/monarch-initiative/mondo-ingest/issues/615, assigning this back to you.

twhetzel: Why is ICD11Foundation not included in the subclass sync and therefore would be added here?

.github/workflows/subclass.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+#  schedule:
+#    - cron: '0 14 * * 1'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.5.1
```
twhetzel, line 11: @matentzn need to update week of 22-Jul-2024 when ODK 1.5.2 is released.

src/ontology/mondo.Makefile
matentzn: I suspect that the we do not have right now any ICD 11 mappings in mondo, and therefore nothing in the subclass sync? Otherwise I don’t know. But feel free to just add the icd11 foundation prefix here!

.github/workflows/subclass.yaml
matentzn: Already done #7985

matentzn: https://github.com/INCATools/ontology-development-kit/releases/tag/v1.5.2

twhetzel: Do you not want to update the ODK version [here](https://github.com/monarch-initiative/mondo/pull/7801/files#diff-b785e7cd995fae5debc10cfbcfa1e93673fcaaf27871d9e2fe5c3dd662c2a58dR11)?

src/ontology/mondo.Makefile
twhetzel, line 975:
```suggestion
		--drop-axiom-annotations "oboInOwl:source=~'(DOID|ICD10CM|icd11.foundation|NCIT|OMIM|OMIMPS|Orphanet):.*'" \
```

.github/workflows/subclass.yaml
twhetzel, line 11:
```suggestion
    container: obolibrary/odkfull:v1.5.2
```

twhetzel (commented):
Two suggested changes/commits

.github/workflows/subclass.yaml
matentzn: good point thx for catching

matentzn:
thx @twhetzel, i will wait for your approval to merge!

twhetzel (approved):
lgtm

PR #7802, opened 2024-06-14
NEVER MERGE: Show how mondo-edit based subclass sync would look like

matentzn:
Shows how mondo-edit based subclass sync would look like. The subclass-sync pipeline is run using the updated mondo-ingest confirmations which _only_ use mondo-edit, not mondo.owl.


matentzn:
Closed in favour of #7801

PR #7805, opened 2024-06-14
Draft 2: Update all confirmed subclassOf evidence using just mondo-edit.

matentzn:
Updates all subclassOf evidence to the latest versions of the source ontologies.

@cmungall @twhetzel @sabrinatoro this is the same as #7801, using only mondo edit rather than mondo.owl which contains "Case 2" from our subclass SOP, see #7801 OP.

Note: given that this is a subset of #7801 its probably less controversial to merge as a first pass, but yea, we need to think hard which is better.


sabrinatoro:
I don't see any problem with this PR. It looks great! 
Also, having more evidence means that there is more support for the subclass annotation (which helps when reviewing)

PR #7818, opened 2024-06-18
DO NOT MERGE!!! Update definition for MONDO:0100016

### LINKED ISSUE #7611: MONDO:0100016 early-onset generalized dystonia [Revise text def]
**Mondo term (ID and Label)**
MONDO:0100016 early-onset generalized dystonia

There are currently synonyms that seem more like definitions on this term:
"A generalis[z]ed isolated dystonia characteris[z]ed by early-onset, which may be clinically indistinguishable from DYT-TOR1A and may be the most common cause of early-onset generalis[z]ed dystonia, at least outside the Askenazi Jewish population."
 
Note that there is a UMLS concept that is 'similar' to this preferred name, but I am not sure that 'early' and 'childhood' should be equated in this situation: "Childhood Onset Dystonias" (UMLS CUI: C0752202) includes one of the many different concepts on https://id.nlm.nih.gov/mesh/D020821.html . Because of the lack of clarity in the scope of this MeSH term, I am not advocating for it as an X-ref, just trying to connect the dots on the various sources of information we're trying to reconcile in MedGen.  I cannot find a better match in UMLS for this "early-onset generalized dystonia."

sabrinatoro:
Closes #7611

Exercice!!!


sabrinatoro:
This PR was for tutorial purposes. Closing.

Issue #7640, opened 2024-04-25
[Revise subclass] MONDO:0016033 Cornelia de Lange syndrome


ValWood:
**Mondo term (ID and Label)**

Cornelia de Lange syndrome  has superclasses "congenital limb malformation" or a "dyostosis"
but it does not appear to be either of these?

**Suggested revision and reasons**

**Your nano-attribution (ORCID)**
If you don't have an ORCID, you can sign up for one [here](https://orcid.org/)

sagehrke:
Dear @ValWood,
Thank you for reaching out to the Mondo Disease Ontology, part of the [Monarch Initiative](https://monarchinitiative.org/)!

We have received your ticket and will review it as soon as possible. 

If you have any additional information or updates relevant to your inquiry, please add them in a comment on this ticket. Your input is valuable in helping us understand and resolve the issue effectively.

Thank you for your patience and for your contributions to Mondo! 

Sincerely,
The Mondo Team

yshwetar:
Issue has been resolved by the following pull-request: https://github.com/monarch-initiative/mondo/pull/7823

MONDO:0016033, Cornelia De Lange Syndrome (CdLS)

Requested removing parent classes "Congenital Limb Malformation" and "Dysostosis".

According to PMID: 20301283, CdLS is characterized with "Limb abnormalities (either symmetric or asymmetric) ranging from severe reduction defects with complete absence of the forearms, to various forms of oligodactyly (missing digits) involving primarily the upper extremities, to small hands (micromelia) and phalangeal abnormalities (5th digit clinodactyly and short first metacarpal resulting in a proximally placed thumb) at the mild end". I have annotated this PMID to "Congenital Limb Malformation" and recommend it remains as a parent class.

Alternatively, CdLS appears to present as a Dysostoses by definition, however is not labeled that anywhere in my literature search. Dysostosis seems to be used to describe disorders with specific patterns of bone development. Considering the lack of usage of such term in the literature, as well as various limb malformation presentations seen in CdLS, I agree with the collaborator to remove Dysostosis as a parent class of CdLS.

- Retain "Congenital Limb Malformation" as a parent class
- Remove Dysostosis as a parent class

PR #7823, opened 2024-06-18
#7640-Revision of Cornelia De Lange Syndrome (Real)

yshwetar:
MONDO:0016033, Cornelia De Lange Syndrome (CdLS)

Requested removing parent classes "Congenital Limb Malformation" and "Dysostosis".

According to PMID: 20301283, CdLS is characterized with "Limb abnormalities (either symmetric or asymmetric) ranging from severe reduction defects with complete absence of the forearms, to various forms of oligodactyly (missing digits) involving primarily the upper extremities, to small hands (micromelia) and phalangeal abnormalities (5th digit clinodactyly and short first metacarpal resulting in a proximally placed thumb) at the mild end". I have annotated this PMID to "Congenital Limb Malformation" and recommend it remains as a parent class.

Alternatively, CdLS appears to present as a Dysostoses by definition, however is not labeled that anywhere in my literature search. Dysostosis seems to be used to describe disorders with specific patterns of bone development. Considering the lack of usage of such term in the literature, as well as various limb malformation presentations seen in CdLS, I agree with the collaborator to remove Dysostosis as a parent class of CdLS.


sabrinatoro (approved):
It looks great! Great work @yshwetar !! 
You can now merge the PR!

PR #7830, opened 2024-06-20
Fix mismatched parentheses in QC check

### LINKED ISSUE #7824: Issues with qc-xref-without-source.sparql
It looks like this QC (sparql/qc/general/qc-xref-without-source.sparql) has been implemented for all sources (OMIM, OMIMPS, Orphanet, DOID, NCIT).

1) what happened to x-refs that did not have a "MONDO:" source?  

2) it looks like this QC does not report all errors. In the next comment, I added examples of terms for which there are xrefs without "MONDO:" source. We should also look at examples where Orphanet said that 1orphanet maps to multiple OMIM terms (we used to have a lot of these examples, but I cannot find any). 

Notes: 
- we do remove all the x-refs that are not equivalent to the released version of Mondo. Mentioning this as it might have some sort of effect.
- I noticed this QC when I curated an OMIM id without a "MONDO:" source in mondo-edit. This reported an error for this omim id

matentzn:
The src/sparql/qc/general/qc-xref-without-source.sparql check had mismatched parentheses which explained why almost no errors where caught correctly. Unfortunately, due to the enormous amount of NCIT and Orphanet errors, I had to drop these from the check..

Fixes #7824 

Needs curation now, as more than 60 errors still present. the default solution is to simply delete all of these xrefs from Mondo!

### DIFF
--- a/src/sparql/qc/general/qc-xref-without-source.sparql
+++ b/src/sparql/qc/general/qc-xref-without-source.sparql
@@ -2,21 +2,21 @@ prefix owl: <http://www.w3.org/2002/07/owl#>
 prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
 prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 
+# description: This QC check ensures that, for our core namespaces, we always have a source
+
 SELECT DISTINCT ?entity ?xref WHERE {
     ?entity oboInOwl:hasDbXref ?xref .
-    OPTIONAL {
-      ?entity owl:deprecated ?obsolete .
-    }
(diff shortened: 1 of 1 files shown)


sabrinatoro:
@matentzn could we remove all the NCIT and Orphanet computationally? 
If you share a list, I can review a few of these to get a pattern of what we see (e.g. DOID were either without source or source from EFO), and if nothing looks suspicious, we could just remove these too.

sabrinatoro (approved):
I reviewed the QC fails, and I approve this PR. 
The PR can be merged once the QC have passed. 
@matentzn Is there something else needed in this PR? or can we merge it?

PR #7874, opened 2024-06-27
Correct synonym as definition

yshwetar:
For issue #7611 

**Previously listed synonyms were in fact the disease definition.** Two synonyms were in place (US english and UK english). **Both were removed and a definition using the same contents was formed.**


sabrinatoro (approved):
This looks good. Thank you @yshwetar.

PR #7882, opened 2024-06-28
Adding obsoletion tag

sabrinatoro:


src/sparql/reports/obsoletiontagcheck.sparql
```diff
@@ -0,0 +1,382 @@
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
```
twhetzel, line 1: @sabrinatoro did you intend to commit this SPARQL query?

src/sparql/reports/obsoletiontagcheck.sparql
```diff
@@ -0,0 +1,382 @@
...
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX replaced_by: <http://purl.obolibrary.org/obo/IAO_0100001>
+PREFIX MONDO: <http://purl.obolibrary.org/obo/MONDO_>
+prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+
+#this query reports the label and the comment (when available) corresponding to specific Mondo IDs
+
+SELECT ?mondo_id ?label ?comment ?subset ?deprecated  WHERE 
+  {
+ values ?mondo_id {
```
twhetzel, line 13: If this query should be in the PR, where did these Mondo IDs come from?

twhetzel (changes requested):
I usually check that the terms in the candiatesForObsoletion [gSheet](https://docs.google.com/spreadsheets/d/1tt1Wk70j9XiHLV1vKQyNiHhaazh286pobpJk1ecSCCg/edit?gid=1661957154#gid=1661957154) match-up with the changes in `mondo-edit.obo`. Depending on what the crossed out lines in the gSheet mean, there are at most ~34~ 32 new obsolete candidates, but many more in the `mondo-edit.obo` file.

src/sparql/reports/obsoletiontagcheck.sparql
sabrinatoro: Yes, I committed this sparql query because it is helpful to have when having a lot of obsoletion candidates and we want to check them before merging the ROBOT template (otherwise the ROBOT merging fail, one term at a time).

src/sparql/reports/obsoletiontagcheck.sparql
sabrinatoro: My goal was to have the query in the repo so we can use it later. If we want to use it again, we would need to replace the MONDO ids with the new obsoletion candidates. Because of this, I didn't think it was necessary to keep the 300+ terms in the query, and I removed them.

sabrinatoro:
The number of obsoletion candidates is more than that this time. If you look at the [googlesheet](https://docs.google.com/spreadsheets/d/1tt1Wk70j9XiHLV1vKQyNiHhaazh286pobpJk1ecSCCg/edit?gid=434975340#gid=434975340): 
- Tab ROBOT_AddObsSubsetTag to add in release 2024-07 : 346 new obsoletion candidates
- Tab ROBOT_AddObsTagForMerge to add in release 2024-07 : 5 new obsoletion candidates
- Tab ROBOT_NewTermSplit-ObsTag to add in release 202407 : 1 new obsoletion candidate

sabrinatoro:
WRT the sparql query: 
When adding a lot of obsoletion candidate tag, we run into failing on the ROBOT template merging because a comment already exist (for example). To facilitate this, I (with Nico's help) created a sparql query to identify the candidates which will make the ROBOT merging fail. 
I thought it would be a good tool to have in the repo for future use. 
I removed most of the MONDO:ID present in the original query because it doesn't make sense to keep them (the important and stable part of the query is everything else). 
I was in a hurry and did not explain everything related around this sparql query. Sorry. I hope it makes more sense now.

twhetzel:
For "Tab ROBOT_AddObsSubsetTag to add in release 2024-07", should there be comments for all rows in the spreadsheet and therefore a `comment: Reason of obsoletion...` property on the Mondo class?

There are only 339 classes with `comment: Reason of obsoletion...`, but 352 classes with `subset: obsoletion_candidate`.

twhetzel (approved):
Thanks for pointing me to the other tabs. 

I added a comment about not all classes having a reason for obsoletion comment, but will approve in the meantime so this can be merged if it those missing properties were intended.

sabrinatoro:
13 candidates already had a comment, so I had to add the obsoletion-related comment manually (otherwise the ROBOT merge would fail because terms have more than one comment). 
This is why you see 339 classes with comment: Reason of obsoletion..., but 352 classes with subset: obsoletion_candidate.

PR #7922, opened 2024-07-05
update xref annotations for OMIM Phenotypic series

### LINKED ISSUE #7766: OMIM Phenotypic series updates
OMIM Phenotypic series not active in OMIM but active in Mondo x-refs (per May update)
(DELETE)
mondo_id	ref_type	ref_target
MONDO_0017138	OMIMPS	300000
MONDO_0000049	OMIMPS	610799
MONDO_0000169	OMIMPS	156850
MONDO_0000218	OMIMPS	616814
MONDO_0030923	OMIMPS	105500
MONDO_0957319	OMIMPS	177735

OMIM phenotypic series that are active/present in OMIM but were missing from Mondo X-refs in May 2024 update
(ADD)
PS619539
PS604571
PS226400
PS209920
PS177735
PS161950
PS151623
PS105210

**Other comments:**
This is not comprehensive, there may be other PS that need to be added; I caught these in a semi-regular check for our processing of OMIM terms that have been renamed to '___ type 1' and realized there was a gap in our OMIMPS data vs. Mondo's x-refs.

twhetzel:
closes #7766 

FYI - For PS151623 Li Fraumeni syndrome, while this is in the [OMIM exclusion list](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/config/omim_exclusions.tsv#L17309) the Mondo term also has the OMIMPS xref.


src/ontology/mondo-edit.obo
```diff
@@ -388432,7 +388435,7 @@ synonym: "hypospadias-hypertelorism syndrome" EXACT [Orphanet:2745]
 synonym: "Opitz BBBG syndrome" RELATED [GARD:0000193]
 synonym: "Opitz G syndrome" EXACT [NCIT:C125487]
 synonym: "Opitz G/BBB syndrome" EXACT [DOID:0050780]
-synonym: "Opitz GBBB syndrome" EXACT [OMIMPS:300000]
+synonym: "Opitz GBBB syndrome" EXACT [OMIM:300000]
```
sabrinatoro, line 388438: @twhetzel I don't understand why this x-ref was updated to OMIM:300000, since OMIM:300000 refers to another Mondo ID. 
I think this OMIMPS x-ref should be kept or another x-ref should be used for this synonym (but not OMIM:300000), but I need to know more before saying that this is what should be done.

sabrinatoro (changes requested):
@twhetzel I have one comment about a x-ref that was updated to using OMIM instead of OMIMPS

src/ontology/mondo-edit.obo
twhetzel: https://omim.org/phenotypicSeries/PS300000 is no longer active. what xref for this synonym should be used now?

sabrinatoro: the xref should be any source(s) that use this synonym. It could be DO, Orphanet,etc., a PMID, and/or your orcid.

twhetzel: Ah, ok I think I understand now what you are asking so I've updated the synonym xref to sources that use that term.

sabrinatoro (approved):
Looks good! Thank you! 
Approving and merging.

PR #7930, opened 2024-07-08
rename term and add definition

### LINKED ISSUE #7604: MONDO:0042605
MONDO:0042605 Y chromosome infertility due to DAZ1 deletion
**Suggested new label:**
Y chromosome infertility 

https://rarediseases.info.nih.gov/diseases/185/y-chromosome-infertility
The syns on this mondo term cite this GARD ID as a source, but those are not current/correct names, per GARD.
Likewise GARD points to the GeneReviews article https://www.ncbi.nlm.nih.gov/books/NBK1339/  which provides a lot of information including links back to OMIM gene entries (See Table B in the text). And the GeneReviews text implies that multiple DAZ gene copies must be impacted in one of the partial deletion alleles for the phenotype to present.

twhetzel:
closes #7604 

Rename term based on information in GARD. I did not find an OMIM entry other than a gene page. This [review article](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3329454/) was helpful.


src/ontology/mondo-edit.obo
```diff
@@ -517780,11 +517780,12 @@ is_a: MONDO:0002254 ! syndromic disease
 
 [Term]
 id: MONDO:0042605
-name: Y chromosome infertility due to DAZ1 deletion
+name: Y chromosome infertility
+def: "A condition that affects the production of sperm, making it difficult or impossible for affected men to father children." [GARD:0000185]
```
sabrinatoro, line 517784: this definition is too broad for the meaning of "Y chromosome infertility": other conditions can affect production of sperm

sabrinatoro (changes requested):
I am having trouble understanding what was asked, but also what these changes here mean.
- There is no term that is MONDO:equivalentTo for GARD:0000185. For instance, MONDO:0042605 show GARD:0000185 as x-ref for some synonym, it is not an equivalent x-ref for this term.
- looking at the GeneReview and the GARD record, it looks like the way to address this issue is not a term relabeling, but more of a review of the existing terms, and possibly reclassification. 
- What is the difference between GARD:0000185 and GARD:16574?
- Note that we update GARD x-ref automatically based on mappings given by GARD, and I cannot find GARD:0000185 (or GARD:185) as an exact match. 

We need to look into this issue deeper to understand the situation and what needs to be done, but I don't think that the changes in this PR are quite correct based on my very quick analysis.

sabrinatoro:
Maybe we need to discuss this issue on a call?

src/ontology/mondo-edit.obo
twhetzel: ok, that was the description from GARD

twhetzel:
Sure, let's discuss  on a call.

twhetzel:
I think this section (subclasses of MONDO:0005372 'male infertility') of Mondo needs to be reviewed in general and suggest any further changes for MONDO:0042605 'Y chromosome infertility due to DAZ1 deletion' be included in this future review. If that's ok with you @sabrinatoro let's close this PR.

sabrinatoro:
@twhetzel I am ok with closing this. 
Please, update the original issue with the summary of all our discussions so we remember the discussions we had, why we didn't update the term label, and where the pain points are. Also, please create a new issue to review MONDO:0005372 'male infertility', referring to this issue. Thank you!

PR #7948, opened 2024-07-11
Add new terms and OMIA xrefs for non-human animal diseases

### LINKED ISSUE #7225: OMIA xrefs for non-human animal diseases
Several issues to address regarding OMIA terms:

1. Design pattern for non-human animal diseases due to genetic mutation. Patterns should be similar to human pattern (e.g. disease based on anatomical entity). The terms with a “cross-species analog” pattern should inherit the human disease pattern (ie we should not need to add another pattern). Need to bring NCBIGene into Mondo for this step.

2. We are adding more OMIA terms to Mondo to improve our coverage of non-human animal diseases. In order to do some testing, I have compared the OMIA SQL dump from 2023-02-09 (our original source of OMIA information) to 2023-12-15. I note that there are several OMIA terms where the name has changed. Please see [spreadsheet](https://docs.google.com/spreadsheets/d/1BKfnDKSn7t8kEEC1ulijGG2FPYWUAi_Nb6FdJp79mlA/edit#gid=0) which is sorted for OMIA terms that are to be included in Mondo and have had name changes. 

To avoid confusion for users who use these Mondo-OMIA mappings and ensure that these mappings are stable and can be trusted, we would like to bring this issue to your attention and suggest solutions. 

Option 1: OMIA does not reuse IDs to refer to different concepts. If a concept changes, the OMIA ID would be obsoleted, and a new ID would be given (similar to SOP for ontologies). This will ensure that our Mondo-OMIA mappings will always be correct since they will always be in sync. However, we recognize that a proliferation of IDs might be problematic, and OMIA existing records need to be updated frequently as more specific gene-disease linkages are discovered in non-human animals.

Option 2: OMIA creates a file that includes these "renamings" (OMIA ID, the previous name, and the new name), and publishes this file with every release. This will help us with keeping the Mondo-OMIA mappings up to date during synchronization.

3. We were wondering whether you would consider implementing some kind of persistent URL scheme for OMIA. Right now, the URL for ‘hypertrophic cardiomyopathy, domestic cat’ OMIA:000515-9685 is https://omia.org/OMIA000515/9685/ . The best practice is to not append characters onto the local ID in the URL (e.g the final ‘/’ character). This will aid in linking OMIA database xref terms in Mondo to the correct OMIA page. Currently, if you view the Mondo term for  ‘hypertrophic cardiomyopathy, domestic cat’ [here](https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_1010459) and then click on the link for the OMIA db-xref, you will get a page not found error due to the final ‘/’. 

Ideally, 
-The URL should be permanent (persistent). When the website changes domain, the identifiers/URIs of the OMIA term stay the same, and still resolve. 

-The suffix, or remainder, of the PURL should be the ID. Would it be possible to make sure that https://omia.org/OMIA000515/9685/ resolves to the same page as https://omia.org/OMIA000515/9685 (without the final ‘/’)?

katiermullen:
Add species-specific non-human animal diseases from OMIA. These new terms are largely acquired diseases without cross-species analogs.

Closes #7225


sabrinatoro (approved):
looks good to me. 
Thank you @katiermullen. approving and merging

PR #7949, opened 2024-07-11
Add a few non-human animal genetic diseases

katiermullen:
Addresses #7271

@sabrinatoro please review this draft PR. 

Note: In the ROBOT template for these terms, in order to avoid a ROBOT MANCHESTER PARSE ERROR, I had to use the gene symbol (e.g. CHAT) rather than the gene ID (e.g. NCBIGene:486775).

In the gene list on Mondo, there are two CHAT entries - http://identifiers.org/hgnc/1912 and http://identifiers.org/ncbigene/486775. Is this a problem? They are orthologs, but should the non-human animal terms only point to the ncbigene entries? 

Thank you!


sabrinatoro:
> In the gene list on Mondo, there are two CHAT entries - http://identifiers.org/hgnc/1912 and http://identifiers.org/ncbigene/486775. Is this a problem? They are orthologs, but should the non-human animal terms only point to the ncbigene entries?

We need to make sure that the gene points to the non-human animal term. As long as we use the NCBIgene ID, it is ok (though the names are the same .

PR #7966, opened 2024-07-16
Add first draft of Matrix disease list

matentzn:


src/sparql/reports/matrix-disease-list.sparql
```diff
@@ -0,0 +1,149 @@
...
+    }
+    ?grouping_class_subset oio:inSubset ?_subset .
+    BIND("TRUE" as ?filter_grouping_subset_ancestor)
+  }
+  
+  OPTIONAL {
+    VALUES ?_subset { 
+      <http://purl.obolibrary.org/obo/mondo#ordo_subtype_of_a_disorder>
+      <http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype>
+    }
```
sabrinatoro, line 65: @matentzn do we have both here because we still have both subsets in Mondo? 
should we only keep the one that is most up-to-date?

src/sparql/reports/matrix-disease-list.sparql
```diff
@@ -0,0 +1,149 @@
...
+    ?category_class oio:inSubset ?_subset .
+    BIND("TRUE" as ?filter_orphanet_disorder)
+  }
+  
+  OPTIONAL {
+    ?category_class rdfs:subClassOf ?subtype_subset .
+    VALUES ?_subset { 
+      <http://purl.obolibrary.org/obo/mondo#ordo_subtype_of_a_disorder>
+      <http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype>
```
sabrinatoro, line 82: @matentzn same question as above

src/sparql/reports/matrix-disease-list.sparql
matentzn: I totally agree, we should delete it..

src/sparql/reports/matrix-disease-list.sparql
```diff
@@ -0,0 +1,149 @@
...
+      <http://purl.obolibrary.org/obo/mondo#rare_grouping>
+    }
+    ?grouping_class_subset oio:inSubset ?_subset .
+    BIND("TRUE" as ?filter_grouping_subset_ancestor)
+  }
+  
+  OPTIONAL {
+    VALUES ?_subset { 
+      <http://purl.obolibrary.org/obo/mondo#ordo_subtype_of_a_disorder>
+      <http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype>
```
matentzn, line 64:
```suggestion
```

src/sparql/reports/matrix-disease-list.sparql
matentzn:
```suggestion
```

src/sparql/reports/matrix-disease-list.sparql
matentzn: Done.

src/sparql/reports/matrix-disease-list.sparql
```diff
@@ -0,0 +1,149 @@
...
+    ?category_class rdfs:subClassOf ?omim .
+    ?omim <http://www.w3.org/2004/02/skos/core#exactMatch> ?match .
+    FILTER(STRSTARTS(STR(?match), "https://omim.org/entry/"))
+    BIND("TRUE" as ?filter_omim)
+  }
+  
+  OPTIONAL {
+    ?category_class rdf:type owl:Class .
+    FILTER NOT EXISTS {
+      ?leaf_x rdfs:subClassOf ?category_class
```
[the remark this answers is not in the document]
matentzn: Brain teaser.

The 🥬 (leaf) is actually the `?category_class` in this case. So we say, 

1. Give me all classes `?category_class rdf:type owl:Class .`
2. Where there is no other class "leaf_x" which is a subclass of it.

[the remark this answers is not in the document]
sabrinatoro: AHH!!!! I fell through the trap! :-D 
I always say "do not rely on a term name"... and here I did fall for it! hahaha! 

Thank you, it makes sense!

PR #7974, opened 2024-07-18
Updates the obsolesence / deprecation status of DOID OMIM NCIT ORDO xrefs

matentzn:
This PR updates the obsolesence status for a number of classes. To review this I would:

1. Do an eyeball review (looking for changes that "stand out", like previously obsolete classes now not being obsolete anymore
2. Checking for 5-10 random classes (2 from each ontology) of they are truly deprecated (sanity check to feel better).

## Comments for Tech team for future reference

This was done by first running

sh run.sh make update_deprecated_mappings -B

To add the updated obsolesence status in, then

sh run.sh make deprecated_annotation_merging -B

To do some, but not all, automatic replacements using hacky sed

Then manually searching for the remaining cases (you are prompted to do that when the previous command is finished).


sabrinatoro:
Since this is the first time we have had this PR, I wonder if we should iron out the issues first before merging. 

For OMIM: the x-refs that are now obsolete have been "moved to" another omim id, for example:
Screenshot 2024-07-18 at 3 54 33 PM

Similarly, the DO id (at least some of them) that are now obsolete are from term merges, and are alternative id to other term. For example: 
Screenshot 2024-07-18 at 3 56 03 PM

(I haven't looked deep down at NCIT to see if there is something similar). 

I think we need to have a plan for these cases as it might indicate that we should review the Mondo term and possibly merge terms in Mondo and/or add new x-ref.

sabrinatoro:
QC fails are because the Mondo term was already obsoleted, and we updated the equivalent x-ref to Mondo:equivalentobsolete, instead of Mondo:obsoleteEquivalentObsolete. 
Maybe we need to resolve this too.

matentzn:
> I think we need to have a plan for these cases as it might indicate that we should review the Mondo term and possibly merge terms in Mondo and/or add new x-ref.

Yes, I agree; It is very difficult to automate this.. Maybe its best if we

1. Dont review these cases
2. Make it so that during the "mapping" pipeline, "moved to" relations in the source are considered as "exact mappings" and added to the new mappings table.

So OMIM:123--moved to->OMIM:124 in the source, MONDO:123--exact-->OMIM:123. OMIM:123 gets obsoleted in Mondo as we do her in this PR. When OMIM:124 is considered for matching, we add MONDO:123--exact-->OMIM:124 to the matches. 

> QC fails are because the Mondo term was already obsoleted, and we updated the equivalent x-ref to Mondo:equivalentobsolete, instead of Mondo:obsoleteEquivalentObsolete.

Must have been a copy paste mistake, as the pipeline takes care of these cases; sorry about that.

twhetzel:
@sabrinatoro following the Tech call, you said it is ok to merge this. Do you plan to fix the qc errors or should someone else?

sabrinatoro (approved):
My concerns (in the comments) have been discussed on the technical call of July 19th 2024. I have resolved the QC failure, and I am approving this PR. 
@matentzn and @twhetzel, this is ready to be merged if there is nothing else that you want to do in this PR.

matentzn (commented):
LGTM, merging!

PR #7982, opened 2024-07-20
refresh imports for August release

twhetzel:
Mondo imports refreshed following these steps: https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/

Since this month has 5 weeks, if you prefer this run on 28-Jul-2024 instead let me know.


sabrinatoro:
@twhetzel did you open mondo in protege and "save as"? (Step 7 [here](https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/#:~:text=Close%20Protege%20and%20open%20mondo%2Dedit.obo%20in%20Protege%20again%20and%20save%20the%20ontology%20(as%20.obo))

I see a diff when I do it (only one updated CL term name). I can merge this diff if you want, but I want to make sure that this comes from a "forgotten" step and not the indication of another issue.

sabrinatoro (approved):
@twhetzel I am approving this PR, but we should make sure to open and save mondo-edit.obo for the updates to be visible in this file before we merge the PR.

twhetzel:
@sabrinatoro I did "Save" the file in Protege, but after I re-opened the ontology again in Protege today using "Save" was not enough to change the file, I had to use "Save as...". The update to `mondo-edit.obo` for the re-named CL term is also now in the PR list of files.

sabrinatoro (approved):

PR #7985, opened 2024-07-21
Fixing ODK version in Mondo to 1.5.2

### LINKED ISSUE #7981: Update ODK to v1.5.2
Update ODK to use version 1.5.2 when this ODK version is released next week.

NOTE: My understanding is that this type of change will require us(@sabrinatoro) to run a mock release to make sure everything is working as expected to avoid disappointment (using my best British phrasing) on the day of the Mondo release.

matentzn:
Fixes #7981 

As per discussions in the tech call, this PR fixes the ODK version to 1.5.2. This way, we can upgrade ODK versions in tandem with mondo ingest, and avoid errors that are caused by variability in the ODK development snapshot.


twhetzel (approved):
lgtm

PR #7993, opened 2024-07-23
Adding HGNC import goal to Makefile

matentzn:
The mondo-rare subset needs this goal:

https://github.com/monarch-initiative/mondo/blob/97be19bdd0b1304fdb40965ee060d3712cb429c2/src/ontology/Makefile#L406

This PR fixes this by making the hgnc import for the rare subset a temporary goal and making sure it is actually run during the release process.


twhetzel (changes requested):
I checked out this branch `hgnc-import` and tested as: `sh run.sh make subsets/mondo-rare.owl IMP=false MIR=false`

The process did not complete successfully:
```
	annotate --ontology-iri http://purl.obolibrary.org/obo/mondo/tmp/hgnc_import.owl annotate -V http://purl.obolibrary.org/obo/mondo/releases/2024-07-23/tmp/hgnc_import.owl --annotation owl:versionInfo 2024-07-23 convert -f owl --output tmp/hgnc_import.owl.tmp.owl && mv tmp/hgnc_import.owl.tmp.owl tmp/hgnc_import.owl
File 'imports/hgnc_terms_combined.txt' does not exist
```

matentzn:
@twhetzel I have simplified the goal now, so it does not need the hgnc_combined.txt file anymore. Check again?

twhetzel:
I re-ran as: `sh run.sh make subsets/mondo-rare.owl IMP=false MIR=false` and it completes successfully and creates (at least) `subsets/mondo-rare.owl` and `(TMPDIR)/hgnc_import.owl`.

As for files under git version control, it also modified `imports/ncbitaxon_terms.txt`. I'm not sure if that was intended and have not had time to dig through to see where this is being changed yet. Approving in the meantime.

twhetzel (approved):

Issue #7983, opened 2024-07-21
Remove OMIM gene mentions from Mondo xrefs / provenance


matentzn:
I jused the following to search for OMIM gene ids in Mondo (`tmp/omim-genes.tsv` is a list of all OMIM genes I got from our OMIM ingest):

```
grep -Ff tmp/omim-genes.tsv mondo-edit.obo > omim_matches.txt
```

I got these matches:

- Synonyms provenance:
  - `synonym: "ENO4" RELATED ABBREVIATION [MONDO:Lexical, OMIM:131375]`: OMIM:131375 [is a gene](https://omim.org/entry/131375)
  - `synonym: "enolase, sperm specific" EXACT [MONDO:Lexical, OMIM:131375]`: OMIM:131375 [is a gene](https://omim.org/entry/131375)
  - `synonym: "HEREDITARY PERSISTENCE of FETAL HEMOGLOBIN" RELATED [OMIM:141900, OMIM:142200, OMIM:142250]`: At least OMIM:141900 [is a gene](https://omim.org/entry/141900)
- Xrefs:
  - `xref: OMIM:131375 {source="MONDO:obsoleteEquivalent"}`
  - `xref: OMIM:141900 {source="MONDO:relatedTo", source="UMLS:C0019025"}`
  - `xref: OMIM:142200 {source="MONDO:relatedTo", source="UMLS:C0019025"}`
  - `xref: OMIM:142250 {source="MONDO:relatedTo", source="UMLS:C0019025"}`
- Definitions:
  - `def: "OBSOLETE. The Duffy blood group system, which consis[.. CUT FOR BREVITY]." [OMIM:110700]`
  - `def: "OBSOLETE. The sense of bitter taste is mediated by [.. CUT FOR BREVITY]." [OMIM:171200]`
  - `def: "OBSOLETE. Multiple genes influence normal human skin, ha..[CUT FOR BREVITY]." [OMIM:227220]`
  - `def: "OBSOLETE. Two types of melanin, the red pheomelanin and t..[CUT FOR BREVITY]." [OMIM:266300]`

### Suggested action items:

- [ ] Remove all xrefs (this I am very sure of)
- [ ] Leave definitions as is, no risk (use of OMIM symbols in text only)
- [ ] Leave synonym prevenance as is as the synonym sync will take care of this (`HEREDITARY PERSISTENCE of FETAL HEMOGLOBIN` is a silly synonym from the perspective of capitalisation

[...]

PR #8011, opened 2024-07-28
Remove references to OMIM entries that are for genes 

twhetzel:
Closes https://github.com/monarch-initiative/mondo/issues/7983


matentzn (approved):
Looks great! Thank you! Agree with everything.

PR #8025, opened 2024-08-02
Add clingen linkouts to Mondo, one off

matentzn:
This adds all ClinGen linkouts to Mondo as an experimental feature using a Mondo internal relationship.

Looks good in Protege:

image

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -85,7 +85,7 @@ all_imports: $(IMPORT_FILES)
 COMPONENTS = mondo-tags
 COMPONENT_FILES = $(foreach n, $(COMPONENTS), components/$(n).owl)
 
-SUBSETS = mondo-minimal mondo-rare
+SUBSETS = mondo-rare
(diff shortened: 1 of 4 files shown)


matentzn:
Note to reviewer:the 6 "removed lines" are the consequences of normalisation:

image

twhetzel:
@matentzn is this intended for the August release?

matentzn:
No need to merge this, I can update this PR again right after the release and we have it for the next one. Up to you. The PR does not affect anything in Mondo (only adds some new relations) so I would personally wave it through, but I see the value to sticking to process.

sabrinatoro (approved):
This is fantastic! Thank you @matentzn. 

I am ok with merging today, though it is not crucial. 
@twhetzel, I am letting you decide how you want to proceed (I can deal with conflict if we want to go ahead and merge for August release). 
If we don't merge today, this update will be part of the update of the ClinGen update, which is Week 3 of the release cycle (is that correct, @matentzn ?), and available in the September release.

twhetzel (approved):

PR #8030, opened 2024-08-05
Add pipeline to automate external content

matentzn:
This PR is about the full automation of the external content pipeline.

It should not be merged before:

1. Testing the associated github action
2. Making sure that bad content is excluded and reported in mondo ingest.

Basically, QC should _not fail_ after a run of the external ingest pipeline.


matentzn:
Superceded now, already in master branch.

PR #8031, opened 2024-08-05
obsolete "rare"

### LINKED ISSUE #6448: Review "bearer of some rare" axioms

sabrinatoro:
Closes #6448

@matentzn I obsoleted the classes "rare or common", "rare" and "not rare.

Please note that
1) only Mondo-edit.obo was changed (and not 2 files as you mentioned in the issue) 2) all the axioms were changed to use "obsolete" instead of being completely removed.

Is this because I "obsoleted" instead of "deleted" the term? How should I proceed from here?


sabrinatoro:
This PR is replaced by https://github.com/monarch-initiative/mondo/pull/8048

PR #8035, opened 2024-08-06
release August

sabrinatoro:


sabrinatoro:
@matentzn and @twhetzel 
I am seeing something very weird... 
Canine sarcoma (MONDO:0700149) in mondo-edit.obo has a SubClassOf annotation (asserted) to "canine neoplasm" (MONDO:0700139) with a source (NCIT:C128195)

This annotation is not in the mondo.owl
I don't know why. I cannot think of what is redundant, and also, this subclass of is therefore missing in mondo.owl

sabrinatoro:
UPDATE: 
I did not see "canine sarcoma" in the "canine neoplasm" branch, but it is in it! 
Therefore all is well. 

We should talk about the new report `reports/difference_release_base.md` on a call because the "Relationships removed" section is confusing.

twhetzel (approved):
lgtm

PR #8038, opened 2024-08-07
Updating synonyms rectal neuroendocrine tumor/neoplasm

### LINKED ISSUE #7618: Overlapping synonyms and incorrect sourcing: MONDO:0015068 & MONDO:0003646
**Mondo term (ID Label)** 
MONDO:0003646 rectum neuroendocrine neoplasm

neuroendocrine tumor of rectum, well differentiated, low or intermediate grade is included as a syn, but this is already in use on  MONDO:0015068 neuroendocrine tumor of rectum, well differentiated, low or intermediate grade
And the synonym(s) overlap between these 2 as well, but I think NCI differentiates between tumor and neoplasm in this case:
neuroendocrine tumor/tumour(s) of rectum

http://purl.obolibrary.org/obo/NCIT_C5698 no longer uses 'tumor' as a syn but is cited as the name source for some of the syns on MONDO:0003646, so the names from NCI may need to be reviewed for these (and possibly more?) 

yshwetar:
Closes #7618

For MONDO:0015068
- Changed label
- Added two synonyms based on NCIT:C

For MONDO:0003646
- Added Orphanet label to already present synonym


src/ontology/mondo-edit.obo
```diff
@@ -66808,7 +66808,7 @@ subset: rare
 synonym: "NET of the rectum" EXACT [Orphanet:100081]
 synonym: "neuroendocrine neoplasm of rectum" EXACT [MONDO:patterns/neuroendocrine_neoplasm, NCIT:C5698]
 synonym: "neuroendocrine neoplasm of the rectum" EXACT [NCIT:C5698]
-synonym: "neuroendocrine tumor of rectum" EXACT [DOID:5777, NCIT:C5698]
+synonym: "neuroendocrine tumor of rectum" EXACT [DOID:5777, NCIT:C5698, Orphanet:100081]
```
sabrinatoro, line 66811: I don't think this synonym should be exact for this term. 
1) NCIT:C5698 does not include it as a synonym (https://www.ebi.ac.uk/ols4/ontologies/ncit/classes?short_form=NCIT_C5698)
2) [orphanet:100081)[https://www.orpha.net/en/disease/detail/100081?name=100081&mode=orpha] this mapping should be reviewed as it might refer to "tumor" and not the more general "neoplasm"
3) this term is very similar to MONDO:0015068-rectal neuroendocrine tumor

sabrinatoro (changes requested):
@yshwetar I think this issue is a little more complicated than just the synonyms.
There are 2 terms in Mondo: 
MONDO:0003646 - 'rectum neuroendocrine neoplasm'
MONDO:0015068 - 'rectal neuroendocrine tumor'

- What is the main difference between these terms? Is it that one is a neoplasm (could be cancerous, benign, etc) and the other one is a "well differentiated neoplasm"? Or is that "rectal neuroendocrine tumor" is a very specific type of tumor (based on the definition)

- Taking this into consideration, and assuming that we trust the mondo to NCIT mappings,  are the mappings (eg to orphanet and DO correct)?

src/ontology/mondo-edit.obo
yshwetar: I agree with your comments. What I think is best we do is

- Go with what NCIT has given its a cancer focused resource
- Orphanet:100081 is to be associated with MONDO:0015068
- All x-refs with Orphanet:100081 on MONDO:0003646 should be removed, as I believe it is referring to MONDO:0015068

yshwetar:
Closing this PR as we are now addressing the issue with #8065

PR #8040, opened 2024-08-07
Extend self loop check to all properties

### LINKED ISSUE #8039: Self Edge Loop MONDO:0008945
Hi Mondo Team,

We continue to see self-edge loop complexities in the base file. Is this expected? It is causing issues in our load. I expect that there may be more lurking about.

Screenshot 2024-08-07 at 10 01 34 AM

matentzn:
Fixes #8039 

Not sure if there will be some legitimate self loops, so lets see what the test will complain about.

### DIFF
--- a/src/sparql/qc/general/qc-cycles.sparql
+++ b/src/sparql/qc/general/qc-cycles.sparql
@@ -14,7 +14,7 @@ SELECT ?entity ?property ?value
 
 WHERE
 {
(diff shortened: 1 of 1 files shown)


twhetzel (approved):
The query looks good to me. I removed cyclic props.

twhetzel:
@sabrinatoro do you want to review this?

sabrinatoro (approved):
this looks good to me.

PR #8041, opened 2024-08-07
Rearranging logic for  hearing loss terms

### LINKED ISSUE #7621: [Revise subclass] autosomal dominant nonsyndromic hearing loss (MONDO:0019587)
**Mondo term (ID and Label)**

I noticed that
autosomal dominant nonsyndromic hearing loss (MONDO:0019587)
seems to have a logical inconsistency .

Screenshot 2024-04-21 at 15 51 40

It is broader than its 2 parents
prelingual non-syndromic genetic hearing loss
postlingual non-syndromic genetic hearing loss
(if these are both autosomal, they should be subclasses, not superclasses)

The logical conclusion of this organization is that  we have exact overlap between annotations:

Screenshot 2024-04-21 at 15 53 14

where I would expect them to be mutually exclusive?

yshwetar:
Closes #7621

The following was done for 'hearing loss, autosomal recessive' MONDO: 0019588, and 'autosomal dominant nonsyndromic hearing loss'

- Added term tracker
- excludedsubclass of for prelingual and postlingual non-syndromic genetic hearing loss
- Placed these terms as children of autosomal dominant nonsyndromic hearing loss


yshwetar:
Closing this PR as we are now addressing the issue with #8073

Issue #7912, opened 2024-07-03
remove "mondo-with-equivalents" and "mondo-minimal" from release


sabrinatoro:
We announced the following:
We plan to remove the following ontology variants from future releases (starting with the July release): mondo-with-equivalents ; mondo-minimal. If you use these ontology variants, please let us know as soon as possible!

We have not heard that these ontology variants are in used. We can go ahead and remove them.

plweller3:
Here at ClinGenm we've been using the mondo-with-equivs for years.  We use it in several projects, you could say it is a critical source of data for us.

I saw no such announcement that this file was going away.  What is it being replaced with?  Is the json schema unchanged in the new file?

matentzn:
@plweller3 first of all, big apologies for the issues this has caused you. We announced it on the mondo-users mailing list 

https://groups.google.com/g/mondo-users/c/Q-xuJEvjPxI/m/bwmnd0dOAAAJ

and in our [Outreach call](https://docs.google.com/presentation/d/1TNhKY3rGoiDU-HpsbL28J7tz_FmC2cFbYvEq1C6DXuc/edit#slide=id.g23eb74d161a_0_0)

But we should have checked more specifically with you, @courtneythaxton and @larrybabb, sorry about that!

I suggest we follow a three part approach

1. I will add the mondo-with-equivalents to the release now so they are there (for now).
2. You and I set up a call and I will tell you the replacement options and how they differ (the schema is the same, one relation is changed).
3. We see that we move ClinGen to the more modern file (it is really not much different, just removing a relationship we now deem obsolete).

What do you think?

matentzn:
I have added the mondo-to-equivalents variants now back to the release as a one off, but it would be good to proceed to steps 2 and 3 as well.. :)

matentzn:
After reviewing @plweller3 pipeline, I suggest we do this:

1. We will wait for @plweller3 to confirm that everything still works as expected using mondo.json.
1. Add a redirect for the mondo-with-equivalents file to the normal mondo (https://github.com/OBOFoundry/purl.obolibrary.org/pull/1002/files). 
1. Not reinstate the release artefact as proposed in #8051, and close that PR

JohnLevander:
A colleague of mine was using mondo-minimal in a GitHub action to create an ontology.  He has since left the project, and I am now maintaining that code.  I tried switching to using mondo-import, but I'm running out of memory on the GitHub Action VM when running the following robot.jar command:

[...]

PR #8051, opened 2024-08-09
DO NOT MERGE: Adding mondo-with-equivalents back

matentzn:
In case we cannot resolve https://github.com/monarch-initiative/mondo/issues/7912 we should consider adding it back..


twhetzel:
Leaving here through the September release to see if this will be needed.

PR #8054, opened 2024-08-10
updates to have ICD11 and EFO entries in the mondo.sssom.tsv file

twhetzel:
In reviewing data for the Synonym Sync I saw that icd11.foundation was not included as a source even though Mondo now includes xrefs to this source. After further review and chat with Nico on Slack, these are the files I updated so that EFO and icd11.foundation are included in the `mondo.sssom.tsv` file. 

To confirm these changes work, I ran `sh run.sh make mappings/mondo.sssom.tsv -B` and then reviewed the `mondo.sssom.tsv` to confirm the updated content.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -198,7 +198,7 @@ ANN = annotate -V $(ONTURI)/releases/`date +%Y-%m-%d`/$@.owl
 GITHUB_ACTION = false
 
 filtered.obo: $(SRC)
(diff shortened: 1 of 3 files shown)


matentzn (approved):
Everything looks great! If you wouldnt have checked the pipeline, I'd be worried about using the `.` character in a regex, but tbh, if it works, it works, and this looks otherwise wonderful.

Issue #8055, opened 2024-08-11
Property used to represent provenance for excluded_synonym is not consistent


twhetzel:
The property used to annotate the provenance of an `excluded_synonym` is not consistent in mondo-edit.obo.

```
$ grep 'excluded_synonym' mondo-edit.obo 
property_value: excluded_synonym "adult Fanconi Anemia" xsd:string {source="DOID:1062"}
property_value: excluded_synonym "Lateral Wall of bladder" xsd:string {source="DOID:11593"}
property_value: excluded_synonym "supraglottis" xsd:string {source="DOID:13476"}
property_value: excluded_synonym "animal viral hepatitis" xsd:string {source="DOID:1884"}
property_value: excluded_synonym "Contractural arachnodactyly" xsd:string {source="GARD:0006975"}
property_value: excluded_synonym "familial dysautonomia, type II" xsd:string {source="DOID:0050548"}
property_value: excluded_synonym "Cutis hyperelastica" xsd:string {xref="DOID:13359", xref="SCTID:238848002", xref="SCTID:398114001", xref="UMLS:C0013720", xref="icd11.foundation:1122707206"}
property_value: excluded_synonym "India rubber skin" xsd:string {xref="SCTID:398114001"}
```

This was found by Joe during work for the Synonym Sync project in order to _not_ suggest adding back an excluded synonym.

Which property, e.g. `xref` or `source`, should be used for the provenance annotations for an excluded_synonym?

matentzn:
I have added comprehensive QC for this issue (and many other related issues) in #8056. The gist of it is:

By default we should only be using `oio:source` for provenance. Exceptions:
    
1. Synonyms use XREF for metadata
2. Definitions use XREF for metadata

Adding this QC revealed 50 or so issues in Mondo that I fixed, so it was great fun to do so!

twhetzel:
Nice to have the rules succinctly stated.

PR #8057, opened 2024-08-11
Fixing axiom annotation properties in mondo-edit

matentzn:
@twhetzel and @joeflack4 have identified a few cases in #8055 where provenance was annotated inconsistently. All of these cases, and more, are fixed in this PR.

check the details in https://github.com/monarch-initiative/mondo/commit/42b13b3318d7cae3e0d39187c9ca5039ff228b85 but the gist of it is: there are only very select specific cases where a property other than oio:source should be used.

### DIFF
--- a/src/sparql/qc/general/qc-axiom-provenance-source.sparql
+++ b/src/sparql/qc/general/qc-axiom-provenance-source.sparql
@@ -14,11 +14,9 @@ SELECT DISTINCT ?entity ?property ?value
 
 WHERE
 {
-    ?entity a owl:Class;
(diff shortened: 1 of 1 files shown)


twhetzel (approved):

twhetzel:
@sabrinatoro do you also want to review this?

sabrinatoro (approved):
I made a few changes (for example, we do not have a "x-ref" or "source" for comments, we add them in the comments themselves). 
I approved and will merged after QC tests have passed.

PR #8073, opened 2024-08-17
Review of pre and post lingual AD hearing loss logic

### LINKED ISSUE #7621: [Revise subclass] autosomal dominant nonsyndromic hearing loss (MONDO:0019587)

yshwetar:
Closes #7621

- Added disjoint logic
- Excluded subclass of for all terms with both pre and post lingual subclass definition
- Added Orphanet sources for excluded subclass of
- Added excluded from QC check for pre and post lingual as they have only one child term each


sabrinatoro (approved):
Perfect! Great job!
Approving and merging.

PR #8079, opened 2024-08-20
Update mondo-idranges.owl

paulacalipho:
id range for Paula Duek Roggli

### DIFF
--- a/src/ontology/mondo-idranges.owl
+++ b/src/ontology/mondo-idranges.owl
@@ -185,5 +185,14 @@ Datatype: idrange:18
 Datatype: xsd:integer
 Datatype: rdf:PlainLiteral
 
+Datatype: idrange:19
 
+        Annotations: 
+            allocatedto: "PaulaDuekRoggli"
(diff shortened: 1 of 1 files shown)


src/ontology/mondo-idranges.owl
```diff
@@ -185,5 +185,14 @@ Datatype: idrange:18
 Datatype: xsd:integer
 Datatype: rdf:PlainLiteral
```
sabrinatoro, line 186: @paulacalipho could you please remove these 2 lines?
Datatype: xsd:integer
Datatype: rdf:PlainLiteral

These should only be at the very end of the file. (there was no way for you to know)

sabrinatoro (changes requested):
@paulacalipho please see my comment about removing 2 lines in the file.

sabrinatoro (approved):
Perfect! 
I approve and merge this PR. 

@paulacalipho, you are all set! Please creating new Mondo terms!

PR #8084, opened 2024-08-21
Run make update external content - update mondo-edit.obo

twhetzel:
New manual run of make command `sh run.sh make update-external-content`


matentzn (approved):
LGTM! Looks like the Makefile changes also ended up in this commit but why not. Lets just push things forward!

sabrinatoro (approved):
It looks good to me. Approving and merging

PR #8091, opened 2024-08-21
Fix qc-proxy-merges query for ROBOT report

matentzn:


twhetzel (approved):
I tested this change locally in a debug branch in mondo-ingest and the make goal (tmp/mondo-incl-robot-report.tsv) that will use this new query ran completely, which is why this query needs to be updated.

Issue #7704, opened 2024-05-11
Exclude `Orphanet:377788` from rare subset pipeline


joeflack4:
## Overview
Exclude `Orphanet:377788` "Disease" from the rare disease subset generation pipeline.

## Context
See this discussion thread:
- https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1592954704

Nico:
> Please exclude
> MONDO:0000001 | http://purl.obolibrary.org/obo/mondo#ordo_disease | Orphanet:377788 | disease | disease
> -- | -- | -- | -- | --
> 
> From the ORDO subset - which is the "ORDO" disorder class itself. Adding this makes the entire ontology become "rare"!

Joe:
> Orphanet:377788 "Disease" is a direct subclass of Orphanet:557493 "Disorder", which is one of the subsets that we've selected. 

Nico:
> I didn't realise "disease" was a subclass of "disorder"... create an issue to... exclude the term from the rare disease subset generation pipeline (with the logic that, while the term is, indeed in the ordo_disorder subset, it should not be added to the rare disease subset, which you are not concerned with here).

matentzn:
To be clear, `Orphanet:377788` is already excluded from the rare disease subset - this issue is about explicitly excluding it from the rare subset pipeline. Thanks @joeflack4, will take care of it!

joeflack4:
Thanks for clarification! I'm completely in the dark on this side of things; those two things sound almost identical. I updated the title~

PR #8095, opened 2024-08-21
Create RD subset pipeline

matentzn:
also closes https://github.com/monarch-initiative/mondo/issues/7704

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -282,30 +282,30 @@ MONDO_INGEST_EXTERNAL_LOCATION=https://raw.githubusercontent.com/monarch-initiat
 ######################################################
 
 update-rare-disease-subset:
-	$(MAKE) subset-metrics -B && cp tmp/subset-metrics.tsv tmp/subset-metrics-before.tsv
+	$(MAKE) subset-metrics -B && cp $(TMPDIR)/subset-metrics.tsv $(TMPDIR)/subset-metrics-before.tsv
 	$(MAKE) update-orphanet-rare -B
 	$(MAKE) update-gard -B
 	$(MAKE) update-nord -B
 	$(MAKE) update-inferred-subset -B
 	$(MAKE) update-rare-subset -B
-	$(MAKE) subset-metrics -B && cp tmp/subset-metrics.tsv tmp/subset-metrics-after.tsv
+	$(MAKE) subset-metrics -B && cp $(TMPDIR)/subset-metrics.tsv $(TMPDIR)/subset-metrics-after.tsv
 	@echo "Subset metrics before..."
-	cat tmp/subset-metrics-before.tsv
+	cat $(TMPDIR)/subset-metrics-before.tsv
 	@echo "Subset metrics after..."
-	cat tmp/subset-metrics-after.tsv
+	cat $(TMPDIR)/subset-metrics-after.tsv
 
 ##### Orphanet Rare ################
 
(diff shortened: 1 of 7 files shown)


src/ontology/mondo.Makefile
```diff
@@ -266,19 +266,142 @@ $(TEMPLATES_DIR)/ROBOT_addMedGen_fromConflictResolution.tsv: tmp/July2023_CUIRep
 $(TEMPLATES_DIR)/ROBOT_addMedGen_fromIngest.tsv:
 	wget "https://github.com/monarch-initiative/medgen/releases/latest/download/medgen-xrefs.robot.template.tsv" -O $@
 
+## FIX THIS BEFORE MERGING
```
twhetzel, line 269: @matentzn what is supposed to be fixed here?

matentzn: Oh yes. I doubled checked it, this has already been fixed on master branch, I deleted the goal (and comment)

twhetzel:
This still fails despite being in the :
```
FAIL Rule ../sparql/qc/mondo/qc-animal-disease-rare.sparql: 1 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0005676,http://purl.obolibrary.org/obo/mondo#rare,Animal disease in Rare subset
``

**Comment by matentzn** [comment]
@twhetzel checking out for today - I _think_ this PR is ready to go! We will discuss the more complicated part on the Mondo Ingest side more carefully.

File: src/ontology/mondo.Makefile [hunk 2]
```diff
@@ -266,79 +266,197 @@ $(TEMPLATES_DIR)/ROBOT_addMedGen_fromConflictResolution.tsv: tmp/July2023_CUIRep
...
+		query --update ../sparql/update/insert-gard-obsoletion-status.ru \
+		convert -f obo --check false -o $(SRC).obo
+	mv $(SRC).obo $(SRC) && make NORM && mv NORM $(SRC) && make deprecated_annotation_merging && make NORM && mv NORM $(SRC)
+
+##### NORD #########################
+
+.PHONY: update-nord
+update-nord:
+	make $(TMPDIR)/external/processed-nord.robot.owl -B
+	grep -vE '^(xref: NORD:|subset: nord_rare)' $(SRC) > $(TMPDIR)/mondo-edit.tmp || true
```
twhetzel, line 372: Since NORD also includes xrefs, does this also need a query similar to `insert-gard-obsoletion-status.ru`?

twhetzel (approved):
@matentzn I have 1 question inline. If no further action is needed based on the question, please merge.

matentzn:
> @matentzn I have 1 question inline. If no further action is needed based on the question, please merge.

I answered this on slack, adding here for the record only:

> Very important question. The more general version of this question is this:
Should we in all cases reject erroneous content from upstream sources, or sometimes attempt to repair it? In this case the question is: If the source provides an xref to an obsolete term, should we
Remove that xref during preprocessing and add it to the source report so they can fix it or
Fix that xref to mean: “this xref is to an obsolete term but no big deal” -
I personally thinks we should go for 1. I left the GARD as a single exception to that rule, as we dont have a very good update circle with them yet (at the moment they sporadically email updates to us), so I opted for repairing, but for all other sources I opted for (1). Let me know what you think!

src/ontology/mondo.Makefile
matentzn: See slack, for now, I think its good like this!

PR #8098, opened 2024-08-21
Run update rare disease subset

twhetzel:
Results of running rare disease pipeline as: `sh run.sh make update-rare-disease-subset -B`.

I separated this out to a different branch since it looks like there might be some qc failures.

Edit: Now that I've pushed this PR I see it would have been fine to have the data and code changes together in https://github.com/monarch-initiative/mondo/pull/8095

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -282,30 +282,30 @@ MONDO_INGEST_EXTERNAL_LOCATION=https://raw.githubusercontent.com/monarch-initiat
 ######################################################
 
 update-rare-disease-subset:
-	$(MAKE) subset-metrics -B && cp tmp/subset-metrics.tsv tmp/subset-metrics-before.tsv
+	$(MAKE) subset-metrics -B && cp $(TMPDIR)/subset-metrics.tsv $(TMPDIR)/subset-metrics-before.tsv
 	$(MAKE) update-orphanet-rare -B
 	$(MAKE) update-gard -B
 	$(MAKE) update-nord -B
 	$(MAKE) update-inferred-subset -B
 	$(MAKE) update-rare-subset -B
-	$(MAKE) subset-metrics -B && cp tmp/subset-metrics.tsv tmp/subset-metrics-after.tsv
+	$(MAKE) subset-metrics -B && cp $(TMPDIR)/subset-metrics.tsv $(TMPDIR)/subset-metrics-after.tsv
 	@echo "Subset metrics before..."
-	cat tmp/subset-metrics-before.tsv
+	cat $(TMPDIR)/subset-metrics-before.tsv
 	@echo "Subset metrics after..."
-	cat tmp/subset-metrics-after.tsv
+	cat $(TMPDIR)/subset-metrics-after.tsv
 
 ##### Orphanet Rare ################
 
-tmp/orphanet-rare-subset.owl: $(SRC)
+$(TMPDIR)/orphanet-rare-subset.owl: $(SRC)
 	$(ROBOT) merge -i $(SRC) reason \
 		query --format ttl --query ../sparql/construct/construct-orphanet-rare-subset.sparql $@
 
 .PHONY: update-orphanet-rare
 update-orphanet-rare:
-	$(MAKE) tmp/orphanet-rare-subset.owl
-	grep -vE '^(subset: orphanet_rare)' $(SRC) > tmp/mondo-edit.tmp || true
-	mv tmp/mondo-edit.tmp mondo-edit.obo
-	$(ROBOT) merge -i $(SRC) -i tmp/orphanet-rare-subset.owl --collapse-import-closure false convert -f obo --check false -o $(SRC).obo
+	$(MAKE) $(TMPDIR)/orphanet-rare-subset.owl
+	grep -vE '^(subset: orphanet_rare)' $(SRC) > $(TMPDIR)/mondo-edit.tmp || true
+	mv $(TMPDIR)/mondo-edit.tmp mondo-edit.obo
+	$(ROBOT) merge -i $(SRC) -i $(TMPDIR)/orphanet-rare-subset.owl --collapse-import-closure false convert -f obo --check false -o $(SRC).obo
 	mv $(SRC).obo $(SRC) && make NORM && mv NORM $(SRC)
 
 ##### GARD #########################
@@ -321,10 +321,10 @@ tmp/gard-rare-subset.owl: $(SRC)
 
 .PHONY: update-gard
 update-gard:
-	$(MAKE) tmp/gard-rare-subset.owl
-	grep -vE '^(xref: GARD:|subset: gard_rare)' $(SRC) > tmp/mondo-edit.tmp || true
-	mv tmp/mondo-edit.tmp mondo-edit.obo
-	$(ROBOT) merge -i $(SRC) -i tmp/gard-rare-subset.owl --collapse-import-closure false convert -f obo --check false -o $(SRC).obo
+	$(MAKE) $(TMPDIR)/gard-rare-subset.owl
+	grep -vE '^(xref: GARD:|subset: gard_rare)' $(SRC) > $(TMPDIR)/mondo-edit.tmp || true
+	mv $(TMPDIR)/mondo-edit.tmp mondo-edit.obo
+	$(ROBOT) merge -i $(SRC) -i $(TMPDIR)/gard-rare-subset.owl --collapse-import-closure false convert -f obo --check false -o $(SRC).obo
 	mv $(SRC).obo $(SRC) && make deprecated_annotation_merging && make NORM && mv NORM $(SRC)
 
 ##### NORD #########################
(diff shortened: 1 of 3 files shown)


src/ontology/mondo-edit.obo
```diff
@@ -110614,7 +110618,9 @@ id: MONDO:0005676
 name: borna disease
 def: "An encephalomyelitis of horses, sheep and cattle caused by borna disease virus." [MESH:D001890]
 comment: This disease is considered zoonotic, i.e. able to be transmitted between human and non-human animals (PMID:34376142). In Mondo, we distinguish between human and non-human animal diseases, and we currently do not have a general term to represent a species-agnostic version of the disease. This record is for the non-human version of “borna disease”.
+subset: nord_rare {source="MONDO:NORD"}
 subset: otar {source="MONDO:OTAR"}
+subset: rare
```
twhetzel, line 110623: Should this be in the `rare` subset? If Yes, then it needs the annotation: `excluded_from_qc_check` and value  `mondoSparqlQcMondo:qc-animal-disease-rare.sparql`

src/ontology/mondo-edit.obo
```diff
@@ -19704,6 +19704,7 @@ subset: otar {source="MONDO:OTAR"}
 subset: rare
 synonym: "Choroiditides" EXACT [MESH:D002833]
 synonym: "choroiditis" EXACT [NCIT:C35111, Orphanet:280892]
+synonym: "Posterior Uveitis" EXACT [NORD:1601]
 synonym: "Posterior Uveitis" RELATED [NORD:1601]
```
twhetzel, line 19708: The new NORD content has "Posterior Uveitis" as an exact synonym so the `RELATED` will be removed.
Screenshot 2024-08-22 at 2 35 59 AM

src/ontology/mondo-edit.obo
```diff
@@ -111948,6 +111954,7 @@ subset: gard_rare {source="GARD:4", source="MONDO:GARD"}
 subset: nord_rare {source="MONDO:NORD", source="NORD:705"}
 subset: rare
 synonym: "Acanthocheilonema perstans infection" RELATED [GARD:0000004]
+synonym: "Acanthocheilonemiasis" EXACT [NORD:705]
 synonym: "Acanthocheilonemiasis" RELATED [GARD:0000004, NORD:705]
```
twhetzel, line 111958: The GARD content (gard-rare-sbset.owl) does not include any information about synonyms and the NORD content includes "Acanthocheilonemiasis" as an EXACT synonym so the the RELATED synonym will be removed.

src/ontology/mondo-edit.obo
```diff
@@ -139896,6 +139904,7 @@ comment: We place the DO class here as it is explicitly AD, but it's not clear i
 subset: gard_rare {source="GARD:15029", source="MONDO:GARD"}
 subset: nord_rare {source="NORD:702", source="MONDO:NORD"}
 subset: rare
+synonym: "Aarskog Syndrome" EXACT [NORD:702]
 synonym: "Aarskog Syndrome" RELATED [NORD:702]
 synonym: "Aarskog syndrome" RELATED [DOID:6683]
 synonym: "Aarskog syndrome, autosomal dominant" RELATED [OMIM:100050]
```
twhetzel, line 139910: @sabrinatoro should "Aarskog syndrome" be listed as an exact or related synonym? 

- the new entry from NORD includes it as an exact synonym
- [DOID:6683](https://www.ebi.ac.uk/ols4/ontologies/doid/classes?obo_id=DOID%3A6683) no longer contains "Aarskog syndrome" as a synonym
- [OMIM:100050](https://omim.org/entry/100050) title is "AARSKOG SYNDROME, AUTOSOMAL DOMINANT"
- there is also [OMIM:305400](https://omim.org/entry/305400) with title "AARSKOG-SCOTT SYNDROME"

These all seem like exact synonyms to me for [autosomal dominant Aarskog syndrome](https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMONDO_0007030).

src/ontology/mondo-edit.obo
```diff
@@ -149374,6 +149384,7 @@ synonym: "spondylocostal dysostosis caused by mutation in TBX6" EXACT [MONDO:des
 synonym: "spondylocostal dysostosis type 5" EXACT [MONDORULE:1, OMIM:122600]
 synonym: "spondylocostal dysplasia" RELATED [OMIM:122600]
 synonym: "spondylothoracic dysostosis" EXACT DEPRECATED [https://rarediseases.org/rare-diseases/jarcho-levin-syndrome/, OMIM:122600]
+synonym: "Spondylothoracic Dysplasia" EXACT [NORD:1915]
 synonym: "Spondylothoracic Dysplasia" RELATED [NORD:1915]
```
twhetzel, line 149388: This synonym is for MONDO:0007389 'spondylocostal dysostosis 5'. While 'spondylocostal dysostosis 5' and Spondylothoracic Dysplasia share some features, their genetic mechanisms are different. 
- [https://omim.org/entry/122600](https://omim.org/entry/122600) lists these alternative titles: SPONDYLOCOSTAL DYSPLASIA and SPONDYLOTHORACIC DYSOSTOSIS
- [NORD:1915](https://rarediseases.org/rare-diseases/spondylothoracic-dysostosis/) states "There is significant confusion in the medical literature regarding names for spondylothoracic dysplasia. For years, this disorder and a similar disorder, spondylocostal dysplasia (were considered the same disorder and referred to as Jarcho-Levin syndrome. Researchers now know that these disorders are separate entities with different causes and associated malformations."

Based on NORD it seems that this synonym should be related vs. exact and perhaps should not even be listed as a synonym for 'spondylocostal dysostosis 5'.

src/ontology/mondo-edit.obo
matentzn: This should have been caught by the Mondo Ingest QC system, so we need to see why it has not!

src/ontology/mondo-edit.obo
matentzn: No it should not - we need to add the "animal disease in RD QC check" to the Mondo Ingest repo and ensure this error does not come through.

src/ontology/mondo-edit.obo
matentzn: This change should not have made it through to here.. The pre-emptive QC system must be able to catch this

src/ontology/mondo-edit.obo
```diff
@@ -147175,6 +147182,7 @@ synonym: "isolated Klippel-Feil syndrome caused by mutation in GDF6" EXACT [MOND
 synonym: "Kfs" RELATED [OMIM:118100]
 synonym: "KFS1" RELATED ABBREVIATION [MONDO:Lexical, OMIM:118100]
 synonym: "Klippel-Feil Syndrome" BROAD [NORD:1336]
```
matentzn, line 147184: This should have been removed, something is going a bit wrong

matentzn (changes requested):

src/ontology/mondo-edit.obo
twhetzel: In mondo, `update-nord` only handles xref and subset annotations, e.g. `grep -vE '^(xref: NORD:|subset: nord_rare)' ...` I gather once the synonym sync is in place this will address cases where synonym scopes are different. 

In mondo-ingest, the `robot_report_external_content.txt` file does not yet include this qc sparql query, but I will add it. 

Does that cover what to check?

src/ontology/mondo-edit.obo
twhetzel: The check sparql/qc/mondo/qc-animal-disease-rare.sparql is added in the mondo-ingest `externalclingenmedgenefo` branch.

src/ontology/mondo-edit.obo
twhetzel: The check `sparql/qc/general/qc-related-exact-synonym.sparql` was not in the list of qc items to handle. I've added it now.

src/ontology/mondo-edit.obo
matentzn: > There is no existing qc check that an exact synonym can not be the same as a broad synonym.

the checked called `duplicate_synonym_scope` (a standard ODK check) should take care of that, at least if memory serves right! Need to test again with the QC checks you added, thanks you.

> Also, I see the nord.robot.tsv that it contains a synonym type column, was that expected to be added?

Yes, that was expected.

twhetzel:
This PR will be closed since these errors should be caught by the QC checks and/or show that additional QC checks need to be added into `robot_report_external_content.txt` in the mondo-ingest repo so that these statements that cause the QC failures here are removed from the externally provided content and are added into the mondo-ingest report `mondo-incl-robot-report.tsv`.

A new run of the pipelines `update-externally-managed-content` in mondo-ingest and `update-rare-disease-subset` in mondo will be done and a new PR created.

PR #8106, opened 2024-08-23

hrshdhgd:


matentzn (approved):
This is a _very_ reassuring PR. It seems that all the ICD 11 mapping work now trickled through, and most of the changes are basically ICD 11 evidence being added.

I am very happy with this @twhetzel and would suggest this is merged asap to avoid conflicts!

PR #8107, opened 2024-08-24

hrshdhgd:


twhetzel (approved):
Confirmed that these OMIMs are obsolete now.

PR #8120, opened 2024-08-28

hrshdhgd:


src/ontology/mondo-edit.obo
```diff
@@ -94517,7 +94518,7 @@ synonym: "DKD" EXACT ABBREVIATION [NCIT:C84417]
 synonym: "type 1 diabetes nephropathy" NARROW [MONDO:0005442]
 synonym: "type 2 diabetes nephropathy" NARROW [MONDO:0005443]
 xref: DOID:11503 {source="MONDO:relatedTo", source="EFO:0000401"}
-xref: EFO:0000401 {source="MONDO:equivalentTo", source="MONDO:preferredExternal", source="MONDO:EFO"}
```
matentzn, line 94520: @twhetzel I made this PR as a test to check of removing the "preferred externals" will break QC, see https://github.com/monarch-initiative/mondo-ingest/pull/628#discussion_r1731876897. It does not!

Feel free since I already made this PR and it looks good to review and merge it, or you can just close it; it does not matter. I made this only as a test!

twhetzel:
Closing since this was made as a test.

Issue #7923, opened 2024-07-05
Bulk request for new gene related syndrome terms from ClinGen Retina GCEP


wchankey:
**Preferred gene-related syndrome label**
Please see attached Excel sheet below for 23 proposed gene-related syndrome labels.

**Synonyms**
Please see attached Excel sheet below for synonyms for 23 proposed gene-related syndrome labels.

**Parent term (use [OLS](https://www.ebi.ac.uk/ols/ontologies/mondo), or your favorite ontology browser)**
Please see attached Excel sheet below for parent terms for 23 proposed gene-related syndrome labels.

**Definition**
Please see attached Excel sheet below for definitions for 23 proposed gene-related syndrome labels.

**Definition source (Please give PubMed ID, if applicable, in format PMID:####
ClinGen Retina Gene Curation Expert Panel

**Children terms (if applicable)** Should any existing terms be re-classified as children underneath this new proposed term? 
Please see attached Excel sheet below for children terms for 23 proposed gene-related syndrome labels.

**Your nano-attribution (ORCID) or URL for a working group**
ORCID ID = 0000-0001-5935-2391 / https://clinicalgenome.org/working-groups/clinical-domain/ocular-cdwg/

[Batch_TermRequests_ClinGenRetinaGCEP_07012024.xlsx](https://github.com/user-attachments/files/16112555/Batch_TermRequests_ClinGenRetinaGCEP_07012024.xlsx)

yshwetar:
Hi @wchankey!

A few changes/additions were made in the pull requests above.

For #8144
- Through literature search, found achromatoposia 5 to be a child of new term 'PDE6C-related retinopathy' and was added as so ([LSWG curation](https://search.clinicalgenome.org/kb/genes/HGNC:8787)).

For #8145
- New term submission 'KCNV2-related retinopathy' has an exact synonym of a term already present in MONDO, 'cone dystrophy with supernormal rod response (MONDO:0012475)'. No new term was made, and instead 'KCNV2-related retinopathy' was made an exact synonym of 'cone dystrophy with supernormal rod response (MONDO:0012475)'.
- Through literature search, found retinopathy of prematurity to be a child of new term 'FZD4-related exudative vitreoretinopathy' and was added as so ([LSWG curation](https://search.clinicalgenome.org/kb/genes/HGNC:4042)).

For #8151
[...]

PR #8130, opened 2024-09-05
New terms PDE6 related retinopathies

yshwetar:
Contributes to #7923 

For both PDE6C and G, the following was added throughout
- New terms
- Definitions
- Narrow synonyms
- Ocular CDWG xref / source
- ORCIDs
- PMIDs
- EquivalentTo logic added


sabrinatoro:
@yshwetar can this PR be closed (without merging) since you are now working on using ROBOT to add these new terms?

PR #8133, opened 2024-09-06
PDE6 via ROBOT

yshwetar:
Trying this.... again


sabrinatoro:
@yshwetar please share the link to your spreadsheet. I can see where the problems might be based on the errors you get.

twhetzel:
The qc check is failing because the MONDO IDs are 8 digits instead of 7. Yousif's ID range is set correctly in `idranges.owl`.

Issue #8002, opened 2024-07-25
Request for new term "mycoplasma infection"


nklsbckmnn:
### Label

mycoplasma infection

### Synonym type

None

### Definition

An infection caused by Mycoplasma.

### Parent term

MONDO_0005550 (infectious disease)

### Children term(s)

MONDO_0019377 (Mycoplasma encephalitis); MONDO_0005867 (Mycoplasma pneumoniae pneumonia)

PR #8143, opened 2024-09-07
Add new term mycoplasma infection

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8002

I modeled the term pattern from https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/specific_infectious_disease_by_agent.yaml and used MONDO:0005747 'enterovirus infectious disease' as an example.


src/ontology/mondo-edit.obo
```diff
@@ -438255,6 +438256,7 @@ xref: MEDGEN:1636101 {source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 xref: Orphanet:83482 {source="MONDO:equivalentTo"}
 xref: UMLS:C4707240 {source="MEDGEN:1636101", source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 is_a: MONDO:0020067 {source="MONDO:Redundant", source="Orphanet:83482"} ! infectious encephalitis
+is_a: MONDO:1030003 {source="https://orcid.org/0000-0002-3458-4839"} ! mycoplasma infection
 intersection_of: MONDO:0020067 ! infectious encephalitis
 intersection_of: disease_has_infectious_agent NCBITaxon:2104 ! Mycoplasmoides pneumoniae
```
sabrinatoro, line 438261: @twhetzel I think this infectious agent is not correct, and should be (maybe?) changed to "Mycoplasma pneumoniae". (please check)
The reason it caught my attention is because "Mycoplasmoides pneumoniae" is not a "Mycoplasma", and therefore I questioned whether it should be a child. But I think the annotation is what is incorrect.

src/ontology/mondo-edit.obo
```diff
@@ -584341,6 +584343,17 @@ is_a: MONDO:0022800 {source="https://clinicalgenome.org/affiliation/40065/", sou
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-3458-4839
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/7379" xsd:anyURI
 
+[Term]
+id: MONDO:1030003
+name: mycoplasma infection
```
sabrinatoro, line 584348: Mycoplasma should have a capital letter because it refers to the genus name

sabrinatoro (changes requested):
@twhetzel 
Please see comment in line regarding the child term. 
Otherwise the term itself looks good.

src/ontology/mondo-edit.obo
twhetzel: done

src/ontology/mondo-edit.obo
twhetzel: Mycoplasma pneumoniae is a basionym (the original name on which a new name is based) for Mycoplasmoides pneumoniae (https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=info&id=2104) so this looks correct to me.

sabrinatoro: @twhetzel: the reason why I am questioning this is because Mycoplanoides pneumoniae is not classified as a Mycoplasma. 

Therefore, if "Mycoplasma infection" is defined as having infectious agent Mycoplasma, 'Mycoplasma encephalitis' defined as having "infectious agent Mycoplanoides pneumoniae" should not be a child of "Mycoplasma infection" (since Mycoplanoides pneumoniae is not classified as a Mycoplasma, screenshot below). UNLESS we are saying that  'Mycoplasma encephalitis' has infectious agents both Mycoplanoides pneumoniae and Mycoplasma (but I don't think it is the case).

@nicolevasilevsky can you please take a look? Maybe I am missing something. 

Screenshot 2024-09-09 at 7 46 09 PM

Screenshot 2024-09-09 at 7 46 29 PM

Screenshot 2024-09-09 at 7 47 09 PM

Screenshot 2024-09-09 at 7 47 25 PM

twhetzel: There was a re-naming over the last few years.
- https://pubmed.ncbi.nlm.nih.gov/29556819/
- https://pubmed.ncbi.nlm.nih.gov/31385780/
- https://pubmed.ncbi.nlm.nih.gov/31971499/

Since Mycoplasma pneumoniae is a basionym for Mycoplasmoides pneumoniae I'd gather these would still be called mycoplasma infections.

sabrinatoro: I understand about the renaming. My issue is not about the renaming. My issue is that the taxon associated with the parent term is neither the same nor an ontological parent of the taxon associated with "mycoplasma encephalitis". 
If the taxon used for "mycoplasma encephalitis" is correct, then the taxon associated with the parent term should be reviewed. But as it currently is, it is not correct ontologically.

nicolevasilevsky: I agree with Sabrina, I think the correct NCBIT term is Mycoplanoides pneumoniae, according to my [google search](https://www.google.com/search?q=primary+infectious+agent+for+mycoplasm+infection&rlz=1C1GCEU_enUS1113US1114&oq=primary+infectious+agent+for+mycoplasm+infection&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIJCAEQIRgKGKABMgkIAhAhGAoYoAEyCQgDECEYChirAtIBCDc2ODJqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8), this seems to be the primary infectious agenda in mycoplasma infections:

image


See:
https://www.ncbi.nlm.nih.gov/books/NBK536927/

twhetzel: With the change in the naming, there is no longer a Mycoplasma pneumoniae, but a Mycoplasmoides pneumoniae (https://www.ebi.ac.uk/ols4/ontologies/ncbitaxon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FNCBITaxon_2104?lang=en). 

I have another suggestion I will post shortly.

twhetzel: To account for the change in names in NCBI Taxonomy from Mycoplasma pneumoniae to Mycoplasmoides pneumoniae I changed the new term name from Mycoplasma infection to Mycoplasmoides infection with a related synonym of mycoplasma infection. 

The individual disease names however do still include the word mycoplasma. For example as found in [DOID](https://www.ebi.ac.uk/ols4/ontologies/doid/classes?obo_id=DOID%3A13276) and [Orphanet](https://www.orpha.net/en/disease/detail/83482).

sabrinatoro (approved):
I approve this PR as the taxon used in the definition of the parent and children terms are taxonomically related. 
@twhetzel i think you need to add something in the main issue to make sure that the user is aware of this new name and new definition. 
I am approving this PR but leaving it un-merged so you can see this comment (and because merging will close the original issue). Feel free to merge whenever you are ready.

PR #8145, opened 2024-09-07
IRDs and EV

yshwetar:
Addresses #7923

- Several new terms
- Definitions
- Term trackers
- ORCIDs
- CDWGs
- equivalentTo logic
- Parents and children
- All children of new terms have single-child exclude subclass QC check
- IMPG1-related dominant retinopathy has no children

* Original submission has "cone dystrophy with supernormal rod response (MONDO:0012475)" as an exact synonym and child for KCNV2-related retinopathy. Should be a synonym, is a child in this submit. Will change in following submission

*Need to add PMIDs for subclass of and definitions


sabrinatoro (approved):
fantastic job @yshwetar !!
There is a conflict in the PR. I can take care of this conflict and merge. 
Thank you!

Issue #8005, opened 2024-07-25
Request for new term "hypertensive urgency"


nklsbckmnn:
### Label

hypertensive urgency

### Synonym type

None

### Definition

Severe hypertension that is characterized by rapid onset of extremely high blood pressure without hypertension-mediated organ damage.

### Parent term

MONDO:0005044 (hypertensive disorder)

twhetzel:
I agree with the user request to add this new term as it is also found in other ontologies such as SNOMED, NCIT, ICD11 and UMLS.

PR #8146, opened 2024-09-10
Add new term 'hypertensive urgency'

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8005

Created the new term based on it's existence in other ontologies such as SNOMED, ICD11, NCIT, and UMLS and that it did not appear to already exist in Mondo.


cmungall:
Should we not always add mappings when we add terms?

Does this belong in Mondo or HPO?

* [] NCIT:C35740 ! Non-Neoplastic Vascular Disorder "A term that includes all non-neoplastic vascular disorders."
    * [i] NCIT:C4796 ! Hypertensive Episode "An episode of elevated blood pressure."
        * [i] NCIT:C3122 ! Hypertensive Crisis "A severe, acute increase in blood pressure that may result in stroke or myocardial ischemia."
            * [i] **NCIT:C197914 ! Hypertensive Urgency "Hypertensive crisis without end-organ damage."**


HPO has "Hypertensive crisis" but it has no children, but presumably urgency would be a subclass?

sabrinatoro:
First: I agree with Chris: mappings/x-refs should be added if they exist

Second: this is another case of "is it a disease or a phenotype? It depends on the point of view..."
At first view, I agree that it looks like a phenotype. However, NCIT classifies this as a disease/disorder. It is also in ICD11 (though there could be things in ICD that are not disease/disorder), and users requested it-assuming it is needed for diagnosis. 

I would be tempted to add this term BUT make it very clear that the term refers to a disease/disorder. That would mean: 
- write the definition to make it clear that we are referring to a disorder (after all, the suggested term's parent is MONDO:0005044-hypertensive disorder)
- maybe add the word "isolated" in the name to indicate that the term refers to a diagnosis ( hypertension that is not part of a larger syndrome) and not to the phenotype.

@nicolevasilevsky @pnrobinson : any thoughts on this?

pnrobinson:
According to https://www.ncbi.nlm.nih.gov/books/NBK513351/
"Hypertensive urgency is a marked elevation in blood pressure without evidence of target organ damage, such as pulmonary edema, cardiac ischemia, neurologic deficits, or acute renal failure."
All of this "bundling" is an indication that this is a disease entity. 
On the other hand, there are many different etiologies:
"The etiology of acute elevations is variable. Noncompliance with antihypertensive therapy, use of sympathomimetics, and thyroid dysfunction are among the many possible causes of hypertensive urgencies. Even anxiety and pain may cause acute elevations in blood pressure and require a different treatment strategy."

So if we want to be purists, we might need multiple entries.

On the other other hand, a patient may have a specific thyroid disease and the hypertension is a manifestation therefore.

I think the reality is that there is a universe of concepts that lives somewhere in between HPO and Mondo that we currently do not do a good job with.

Practically speaking, this should be a Mondo term rather than an HPO term.

sabrinatoro:
Thank you @pnrobinson

src/ontology/mondo-edit.obo
```diff
@@ -584370,6 +584369,20 @@ is_a: MONDO:1030005 {source="PMID:31120550"} ! mosaic SMO syndrome
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-3458-4839
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/7768" xsd:anyURI
 
+[Term]
+id: MONDO:1030007
+name: hypertensive urgency
+def: "Hypertensive urgency occurs when blood pressure spikes -- blood pressure readings are 180/110 mmHg or higher -- but there is no damage to the body's organs (such as pulmonary edema, cardiac ischemia, neurologic deficits, or acute renal failure). Blood pressure can be brought down safely within a few hours with blood pressure medication." [icd11.foundation:896211058]
```
sabrinatoro, line 584375: @twhetzel Can you please make the definition such that it clearly refers to a disease/disorder. For example: "a hypertensive disorder characterized by (blablabla)" 

Please also the PMID for the [publication that Peter shared] (https://www.ncbi.nlm.nih.gov/books/NBK513351/). Some of the information from that publication can also be used for the definition (especially the first sentence)

sabrinatoro: Adding the orcid of the requester as x-ref would be good, but it doesn't look like we have it.

src/ontology/mondo-edit.obo
```diff
@@ -584370,6 +584369,20 @@ is_a: MONDO:1030005 {source="PMID:31120550"} ! mosaic SMO syndrome
...
+synonym: "hypertensive crisis without acute organ damage" EXACT [icd11.foundation:896211058]
+synonym: "severely elevated blood pressure without acute organ damage" EXACT [icd11.foundation:896211058]
+xref: icd11.foundation:896211058 {source="MONDO:equivalentTo"}
+xref: NCIT:C197914 {source="MONDO:equivalentTo"}
+xref: SCTID:443482000 {source="MONDO:equivalentTo"}
+xref: UMLS:C0745138 {source="MONDO:equivalentTo"}
+is_a: MONDO:0005044 {source="icd11.foundation:896211058"} ! hypertensive disorder
```
sabrinatoro, line 584382: Adding the orcid of the requester as x-ref would be good, but it doesn't look like we have it.

sabrinatoro (changes requested):
Requesting rewording of the definition to make it clear that the term refers to a disease/disorder.

src/ontology/mondo-edit.obo
twhetzel: I came across the requester's ORCID in another issue, https://orcid.org/0009-0009-6791-365X. I verified this is an actual ORCID, however similar to the GH profile their is no public information available for this ORCID entry.

twhetzel: the definition is updated, i will push this change shortly

src/ontology/mondo-edit.obo
twhetzel: I found the ORCID in another ticket from this submitter and added this

src/ontology/mondo-edit.obo
sabrinatoro: Instead of https://www.ncbi.nlm.nih.gov/books/NBK513351/, please use PMID:30020723

sabrinatoro (approved):
@twhetzel: I have only one change requested: use PMID instead of the link to the Pubmed page (see comment).
Since this is small change, I have approved the PR, so you can merge after making the change requested above and QC passed.

pnrobinson:
It would be good to discuss a more general plan for HPO-Mondo coordination rather than one term at a time! @mellybelly @cmungall

PR #8150, opened 2024-09-10
Orphanet alignment for October release

sabrinatoro:
Spreadsheet is here: https://docs.google.com/spreadsheets/d/18BXvv1jHp2gpO0AsBsxaZqp77dvcatk0yNXTTiLP9NE/edit?gid=0#gid=0


src/ontology/mondo-edit.obo
```diff
@@ -560385,6 +560385,63 @@ name: arterial tortuosity-bone fragility syndrome
...
+name: papillary hemangioma
+xref: Orphanet:673543 {source="MONDO:equivalentTo"}
+is_a: MONDO:0971115 {source="Orphanet:673543", source="PMID:17927585", source="https://orcid.org/0000-0002-4142-7153"} ! benign vascular tumor
+
+[Term]
+id: MONDO:0975754
+name: pseudomyogenic hemangioendothelioma
+xref: Orphanet:673556 {source="MONDO:equivalentTo"}
+is_a: MONDO:0971116 {source="Orphanet:673556", source="PMID:36108375", source="https://orcid.org/0000-0002-4142-7153"} ! borderline vascular tumor
+relationship: excluded_from_qc_check http://purl.obolibrary.org/obo/mondo/sparql/qc/general/qc-single-child.sparql
```
nicolevasilevsky, line 560413: Orphanet has several classes classified under 'borderline vascular tumor':
https://www.orpha.net/en/disease/classification/group/162973?orphaCode=673473&firstParent=0&name=Borderline-vascular-tumor&mode=name
image

I wonder if we should classify these other cancers under this class? Or maybe this is not a good grouping?

NCIT has a similar term, Intermediate Vascular Neoplasm, and has their own classification, maybe we want to follow this:
https://www.ebi.ac.uk/ols4/ontologies/ncit/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FNCIT_C7334

image

I don't think this should be a blocker for this PR, but if you think we should consider this classification, we should create a ticket.

nicolevasilevsky (approved):
Approving this PR but I'll let you merge @sabrinatoro, to give you a chance to review my comment.

src/ontology/mondo-edit.obo
sabrinatoro: Thank you @nicolevasilevsky. 
I think we should definitely review classification. I will create an issue for this, as I think it will take some time to review. (note: I am hoping that we can get a lot of these classifications with the subclass alignment)

Issue #8082, opened 2024-08-21
Clarify immunodeficiencies involving POLD1 and POLD2


cmungall:
We have a term that was added in #5261

```yaml
id: MONDO:0800145
name: non-severe combined immunodeficiency due to polymerase delta deficiency
def: "Any non-severe combined immunodeficiency in which the cause of the disease is variation in the POLD1/POLD2 gene." [PMID:31449058]
subset: clingen {source="MONDO:CLINGEN"}
subset: gard_rare {source="MONDO:GARD"}
subset: rare
is_a: MONDO:0018814 {source="https://orcid.org/0000-0002-7371-8158", source="https://orcid.org/0000-0003-0113-912X"} ! non-SCID combined immunodeficiency
relationship: curated_content_resource https://search.clinicalgenome.org/kb/conditions/MONDO:0800145 {source="MONDO:CLINGEN"}
relationship: RO:0002573 HP:0000007 {source="https://orcid.org/0000-0002-7371-8158", source="https://orcid.org/0000-0003-0113-912X"} ! has modifier Autosomal recessive inheritance
property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/5261" xsd:anyURI
```

This doesn't conform to a DP. We don't have a DP for disjunctions of genes.

It's important to conform to DPs because otherwise the ontology becomes too hard to manage. For example, we also have:

```yaml
id: MONDO:0970994
name: immunodeficiency 120
xref: OMIM:620836 {source="MONDO:equivalentTo"}
is_a: MONDO:0021094 {source="OMIM:620836", source="https://orcid.org/0000-0002-4142-7153"} ! immunodeficiency disease
```

This is caused by POLD1 (why don't we have the edge linking to gene?)

What is the relationship between these two MONDO IDs? Right now they are unlinked, but they are clearly part of the same spectrum of deficiencies involving one or more parts of the POLD complex.

cmungall:
Note that OMIM annotated the *same paper* and collected in under https://omim.org/entry/620836 (immunodeficiency 120)

aslgraefe:
USIDNET2024 for POLD1 & POLD2: 
_Inheritance_: AR
_Associated Features_: Recurrent respiratory tract infections, skin infections, warts and molluscum, short stature, intellectual disability; T cell count: Low Low CD4 T cells; low B cell count; low Immunoglobulin levels

sabrinatoro:
[...]

PR #8173, opened 2024-09-17
pattern for disease caused by one of multiple genes

sabrinatoro:
Addresses #8082

@cmungall Could you please review this? 
Is this what you had in mind? I know using "or" in logical definition is not great, but it is what would work for this (I think). 

That being said, the logical definition using "or" in mondo-edit.obo looks unlike anything else in OBO. This makes me question whether what I did is correct. Screenshots below for convenience. Thank you! 

Screenshot 2024-09-17 at 2 08 39 PM

### DIFF
--- a/src/patterns/dosdp-patterns/disease_series_by_groupofgenes.yaml
+++ b/src/patterns/dosdp-patterns/disease_series_by_groupofgenes.yaml
@@ -1,0 +1,62 @@
+pattern_name: disease_series_by_groupofgenes
+
+pattern_iri: http://purl.obolibrary.org/obo/mondo/patterns/disease_series_by_groupofgenes.yaml
+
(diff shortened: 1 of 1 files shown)


cmungall (changes requested):
This isn't the correct way to do multivalued fields in dosdp. These were introduced here https://github.com/INCATools/dead_simple_owl_design_patterns/issues/71. I will see if I can find some docs on how to use them. @matentzn may be able to help.

Note we can't represent unions in expressions, obo format, and union reasoning in Elk is incomplete

But why do we even need a logical definition here? I would focus on documenting the pattern, and only do logical definitions for really simple clear cases.

How often do we expect this? Surely in most cases the grouping will correspond to a union of gene-specific diseases (here a disjoint union is fine)

Issue #8168, opened 2024-09-17
Document QC checks run for Externally Managed Content


twhetzel:
Add a page to the Mondo User Corner, e.g. https://mondo.readthedocs.io/en/latest/editors-guide/rare-disease-subset/, to document which QC tests are run for externally managed content and what each failure type means.

PR #8186, opened 2024-09-19
Document QC checks for EMC

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8168


docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
+## Automatic Mondo Quality Control Checks for Externally Managed Content
+A subset of the extensive Mondo QC checks are automatically run on externally managed content. Any content from external providers that does not pass these QC checks is not included in Mondo until the content is updated and passes the QC checks. Reports of these QC checks will be available for review of any errors so that any needed changes can be made.
```
sabrinatoro, line 2: Add "at the source" or "by the external providers"

"is not included in Mondo until the content is updated **at the source/by the data providers** and passes the QC checks."

sabrinatoro: Maybe add something after the first sentence to explain that we do that so we don't block the update. 
Suggestion: 
**To limit delay due to QC failure,** any content from external providers (...)

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+
+### QC Checks for Externally Managed Content
+| Error | Description | SPARQL Query | 
+|:------|:------------|:-------------|
+| duplicate_scoped_synonym | An entity has duplicate synonyms with different properties (e.g. the same broad and related synonym). This causes ambiguity. | <a href='https://robot.obolibrary.org/report_queries/duplicate_scoped_synonym' target='_blank'>duplicate scoped synonym</a> |
```
sabrinatoro, line 7: The description is confusing. 
is this reporting a mondo term?

Suggestion: 
The synonym from the external source is the same as a synonym already reported on a Mondo term, but the synonym scope is not the same. For example, the synonym from the source is "exact", and the synonym in Mondo is "broad".

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| invalid_xref | A database_cross_reference is not in CURIE format. | <a href='https://robot.obolibrary.org/report_queries/invalid_xref' target='_blank'>invalid xref</a> |
```
sabrinatoro, line 8: Suggestion: 
A database_cross_reference in the EMC file is not in CURIE format.

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| qc-duplicate-exact-synonym-no-abbrev| An exact synonym, excluding abbreviations, or label is used for two different entities.| <a href='https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/general/qc-duplicate-exact-synonym-no-abbrev.sparql' target='_blank'>duplicate synonyms</a>|
```
sabrinatoro, line 9: Suggestion: 
An exact synonym, excluding abbreviations, or label is used in Mondo for two different Mondo terms/entities.

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| qc-related-exact-synonym| A value for a synonym is used with both the exact and related synonym scopes.|<a href='https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/general/qc-related-exact-synonym.sparql' target='_blank'>related exact synonym</a>|
```
sabrinatoro, line 10: how is this one different from duplicate_scoped_synonym ?

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| qc-syn-source-not-xref| A synonym does not have a database cross reference as the source of the synonym.| <a href='https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/general/qc-syn-source-not-xref.sparql' target='_blank'>missing synonym source</a>|
```
sabrinatoro, line 11: I am confused by this one.  Do we get files from external sources that do not have any x-ref and have only synonyms? Is this what this is checking? 

if so, 
A synonym from the external source file does not have a database cross reference (...)

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| qc-trailing-whitespace.sparql| An entity contains a leading or trailing whitespace. | <a href='https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/general/qc-trailing-whitespace.sparql' target='_blank'>trailing whitespace</a>|
```
sabrinatoro, line 12: Is this a EMC related QC check? I am assuming that we have a check to make sure that we remove the extra whitespace when we process these files. 

Do we really exclude information and tell the source to remove trailing whitespace? (that seems a bit weird to me)

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| qc-animal-disease-rare| An animal disease is contained within a rare disease subset. |<a href='https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/mondo/qc-animal-disease-rare.sparql' target='_blank'>animal disease rare</a>|
```
sabrinatoro, line 13: Again, is this really a QC for the EMC?

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,33 @@
...
+| qc-proxy-merges| Two Mondo classes have a mapping to the same term. See <a href='https://mondo.readthedocs.io/en/latest/editors-guide/proxy-merge/' target='_blank'>proxy merges</a> for more information.| <a href='https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/mondo/qc-proxy-merges.sparql' target='_blank'>proxy merge</a>|
```
sabrinatoro, line 14: "to the same **external source** term'

sabrinatoro (changes requested):
@twhetzel I am confused by this document. 
- If this document is for the external sources, then it needs more explanation about what external managed content is, how it is process, etc. But I can set this on the side for now and we can update later. 
- Generally, I think the descriptions need to be clearer about what information is from the external source, and what information is in Mondo. For example "a synonym from the external source is already in Mondo but with a different scope". Therefore the synonym is ignored until it can be reviewed (in Mondo and in the external source)
- I am a bit confused about the QC check contents. I thought these would be the one reported [here](https://docs.google.com/document/d/1SLCIIxbW-B-cXfg15A6Xpq9oQ9t0yaHJ_yVkaQia__c/edit#heading=h.fgkun4my6zai). However, there is a lot of other QC checks in this document which seem more general and not specific to the EMC. For example: the white space trailing.  I guess my question is: are these truly specific to the externally managed content? (if the answer is yes, then it is all good.)

twhetzel:
I don't know why the QC checks for EMC are not the ones in listed in the [doc](https://docs.google.com/document/d/1SLCIIxbW-B-cXfg15A6Xpq9oQ9t0yaHJ_yVkaQia__c/edit#heading=h.fgkun4my6zai) you referenced. The Mondo QC checks that are done for EMC are based on the failures of EMC content that has occurred during the development of this pipeline. These are all the checks of things I've had to fix manually when Nico would submit a PR for adding this content.

twhetzel:
Also, I do not know if it is still the plan to email the report to providers or not.

docs/editors-guide/externally-managed-content.md
twhetzel: updated

docs/editors-guide/externally-managed-content.md
twhetzel: updated

docs/editors-guide/externally-managed-content.md
twhetzel: updated

docs/editors-guide/externally-managed-content.md
twhetzel: updated: An exact synonym, excluding abbreviations, or label in the EMC file duplicates an existing exact synonym or label for a different Mondo term/entity.

docs/editors-guide/externally-managed-content.md
twhetzel: good call, since duplicate_scoped_synonym includes this check (qc-related-exact-synonym) I removed it here

docs/editors-guide/externally-managed-content.md
twhetzel: updated

docs/editors-guide/externally-managed-content.md
twhetzel: i think this did happen once and is why it's listed here

docs/editors-guide/externally-managed-content.md
twhetzel: well, this is an existing QC check for Mondo so if the trailing whitespace is not fixed at the source each month a curator will need to fix this manually. this is added because it has happened in past additions of EMC content

twhetzel: however, i also agree that removing any leading/trailing whitespace should be something that happens when processing the EMC content

docs/editors-guide/externally-managed-content.md
twhetzel: let me look at this one again....

docs/editors-guide/externally-managed-content.md
```diff
@@ -0,0 +1,32 @@
+## Automatic Mondo Quality Control Checks for Externally Managed Content
+A subset of the extensive Mondo QC checks are automatically run on the externally managed content. To limit delay due to QC failure, any content from external providers that does not pass these QC checks is not included in Mondo until the content is updated by the data providers and passes the QC checks. Reports of these QC checks will be available for review of any errors so that any needed changes can be made.
+
+### QC Checks for Externally Managed Content
+| Error | Description | SPARQL Query | 
+|:------|:------------|:-------------|
+| duplicate_scoped_synonym | The synonym from the external source is the same as a synonym already reported on the Mondo term, but the synonym scope is not the same. For example, the synonym from the source is "exact", and the synonym in Mondo is "broad".| <a href='https://robot.obolibrary.org/report_queries/duplicate_scoped_synonym' target='_blank'>duplicate scoped synonym</a> |
+| invalid_xref | A database_cross_reference in the EMC file is not in CURIE format.| <a href='https://robot.obolibrary.org/report_queries/invalid_xref' target='_blank'>invalid xref</a> |
```
sabrinatoro, line 8: EMC is not defined in the document yet. We should have it in the first paragraph so we know that EMC is Externally Managed Content.

sabrinatoro (approved):
@twhetzel My only comment is that EMC needs to be defined early in the text so we know what it stands for. Otherwise, this iteration is fine. 

I am approving so you can merge after having added what EMC stands for, without waiting for another review.

docs/editors-guide/externally-managed-content.md
twhetzel: done

Issue #7999, opened 2024-07-24
Request for new synonym Complex neurodevelopmental disorder with or without aldosteronism


Jciesielka1:
### Mondo term

0014200: Aldosterone-producing adenoma with seizures and neurological abnormalities

### Synonyms

Complex neurodevelopmental disorder with or without aldosteronism

### Synonym type

None

### Additional comments

Please add the following definition for this term: An autosomal dominant neurodevelopmental condition related to variants in CACNA1D. Most reported variants are de novo and functional studies have indicated a gain-of-function disease mechanism (PMIDs: 23913001, 37122292). This condition is characterized by developmental delay/intellectual disability, autism spectrum disorder, hypotonia and seizures (PMIDs: 32666275, 37122292). Other reported features include endocrine abnormalities such as primary aldosteronism and congenital hyperinsulinemic hypoglycemia, self-injurious behavior, facial dysmorphisms, and heart defects.

ErinRiggs:
Hello - just checking on the status of this request?

PR #8192, opened 2024-09-23
Add synonym and definition to MONDO:0014200: 'aldosterone-producing adenoma with seizures and neurological abnormalities'

twhetzel:
Closes https://github.com/monarch-initiative/mondo/issues/7999

I used the source PMID:32583268 instead of PMID:32666275 since the former is the primary article and latter is a correction to since the gene symbol was incorrect in the original publication.


sabrinatoro (approved):
looks good to me.

Issue #7887, opened 2024-06-30
Request for new term "persistent atrial fibrillation"


nklsbckmnn:
### Label

Persistent atrial fibrillation

### Synonym type

None

### Definition

Atrial fibrillation that is continuous and sustains for >7 d.

### Parent term

MONDO_0004981 atrial fibrillation

### Additional comments

https://www.ahajournals.org/doi/10.1161/CIR.0000000000001193

twhetzel:
@kanems this will be a new term (MONDO:1030009) mapped as 'equivalentTo' both MEDGEN:749907 and UMLS:C2585653.

PR #8194, opened 2024-09-23
Add new term persistent atrial fibrillation

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7887


twhetzel:
Closing since I can not sort why there are changes to some imports in this PR.

Issue #8003, opened 2024-07-25
Request for new term "mitral valve insufficiency"


nklsbckmnn:
### Label

mitral valve insufficiency

### Synonym type

None

### Definition

Dysfunction of the mitral valve characterized by incomplete valve closure.

### Parent term

MONDO:0003767 (mitral valve disorder)

### Children term(s)

MONDO:0001298 (congenital mitral valve insufficiency)

### ORCID Identifier

https://orcid.org/0009-0009-6791-365X

PR #8195, opened 2024-09-23
Add new term persistent atrial fibrillation

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7887


src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,20 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-3458-4839
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/8003" xsd:anyURI
 
+[Term]
+id: MONDO:1030009
+name: persistent atrial fibrillation
+def: "A disorder characterized by continuous atrial fibrillation that lasts for more than seven days." [https://orcid.org/0009-0009-6791-365X]
```
This definition implies that "any disorder" that has continuous atrial fibrillation that lasts for more than seven days is a "persistent atrial fibrillation". This is not correct. 
The definition (genus-differentia) should use the most specific genus, which here should be "atrial fibrillation" (ie the parent term). 

dbxref: 
- Please add the PMID of the paper shared in the issue as a dbxref for this term. 
- this ORCID refers to a private record. Is this the correct orcid? If so, who's orcid is it?

src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,20 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
...
+synonym: "persistent AF" EXACT [PMID:29401239]
+xref: ICD10CM:I48.1 {source="MONDO:equivalentTo"}
+xref: MEDGEN:749907 {source="MONDO:equivalentTo"}
+xref: MESH:D001281 {source="MONDO:relatedTo"}
+xref: SCTID:440059007 {source="MONDO:equivalentTo"}
+xref: UMLS:C2585653 {source="MONDO:equivalentTo"}
```
sabrinatoro, line 584975: Please make sure to tag Megan on the issue to let her know that we have a new Mondo term, and that this Mondo term is equivalent to this MEDGEN and UMLS id, so she can update her mapping files.

src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,20 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
...
+is_a: MONDO:0004981 {source="https://orcid.org/0009-0009-6791-365X"} ! atrial fibrillation
```
sabrinatoro, line 584976: Please use the PMID as a source. 
Also same question as above about the orcid

sabrinatoro (changes requested):
@twhetzel Please see comments in lines. 

We should (as much as possible) use PMID as x-refs for definition and as source for SubClass of. It is ok to add orcid, but PMID (or x-refs from other sources) indicate that there is more support behind that definition/SC than "someone said so". 
Also, the orcid used here leads to a private account. We should avoid using it (how do we know for sure it is the correct one?).

src/ontology/mondo-edit.obo
twhetzel: Sure, I'll update the definition. I've been asked to remove this kind of thing before so I did not include it here.

I added the PMID as an xref on the definition.
The user provided this orcid in a [previous request](https://github.com/monarch-initiative/mondo/issues/8003). I will remove it based on your PR summary comment.

sabrinatoro: > I've been asked to remove this kind of thing before so I did not include it here.

We discussed this previously: what was asked to remove from the definition is the disease name itself (ie the term name itself) from the definition. For instance, the definition of "persistent atrial fibrillation" should not start with "Persistent atrial fibrillation is a ..."

The Genus, ie most often parent term name, is needed for the definition to follow the "genus-differentia" pattern.

src/ontology/mondo-edit.obo
twhetzel: updated to pmid and removed orcid

src/ontology/mondo-edit.obo
twhetzel: done

sabrinatoro (approved):

Issue #7885, opened 2024-06-30
Request for new term "precerebral artery stenosis"


nklsbckmnn:
### Label

precerebral artery stenosis

### Synonym type

None

### Definition

Abnormal narrowing of a precerebral artery.

### Parent term

cerebrovascular disorder

PR #8196, opened 2024-09-24
Add new term precerebral artery stenosis

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7885


src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,18 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-3458-4839
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/8003" xsd:anyURI
 
+[Term]
+id: MONDO:1030010
+name: precerebral artery stenosis
+def: "A disorder characterized by an abnormal narrowing of a precerebral artery." [https://orcid.org/0009-0009-6791-365X]
```
This definition implies that "any disorder" that has an abnormal narrowing of a precerebral artery is a "precerebral artery stenosis". This is not correct.
The definition (genus-differentia) should use the most specific genus, which here should be "cerebrovascular disorder" (ie the parent term).

dbxref:
this ORCID refers to a private record. Is this the correct orcid? If so, who's orcid is it?
Please find another x-ref to add here. Maybe one of the x-ref would work?

src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,18 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
...
+synonym: "narrowing of precerebral artery" EXACT [SCTID:1055001]
+xref: ICD10CM:I65 {source="MONDO:relatedTo"}
+xref: SCTID:1055001 {source="MONDO:equivalentTo"}
+xref: UMLS:C0265089 {source="MONDO:equivalentTo"}
+is_a: MONDO:0011057 {source="https://orcid.org/0009-0009-6791-365X"} ! cerebrovascular disorder
```
sabrinatoro, line 584974: This orcid is private. Please add another source for this SC.

sabrinatoro (changes requested):
Please see comments.

src/ontology/mondo-edit.obo
twhetzel: updated

src/ontology/mondo-edit.obo
twhetzel: updated

sabrinatoro (approved):
@twhetzel 
I am approving this PR. 
The conflict needs to be resolved before merging. Let me know if you want me to do it.

sabrinatoro (approved):

Issue #7886, opened 2024-06-30
Request for new term "paroxysmal atrial fibrillation"


nklsbckmnn:
### Label

Paroxysmal atrial fibrillation

### Synonym type

None

### Definition

Atrial fibrillation that is intermittent and terminates within ≤7 d of onset

### Parent term

MONDO_0004981 atrial fibrillation

### Additional comments

https://www.ahajournals.org/doi/10.1161/CIR.0000000000001193

PR #8197, opened 2024-09-24
Add new term paroxysmal atrial fibrillation

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7886


src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,21 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-3458-4839
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/8003" xsd:anyURI
 
+[Term]
+id: MONDO:1030011
+name: paroxysmal atrial fibrillation
+def: "A disorder characterized by intermittent atrial fibrillation that lasts less than seven days from onset." [https://orcid.org/0009-0009-6791-365X]
```
This definition implies that "any disorder" that has continuous atrial fibrillation that lasts for more than seven days is a "persistent atrial fibrillation". This is not correct.
The definition (genus-differentia) should use the most specific genus, which here should be "atrial fibrillation" (ie the parent term).

dbxref:

Please add the PMID of the paper shared in the issue as a dbxref for this term.
this ORCID refers to a private record. Is this the correct orcid? If so, who's orcid is it?

src/ontology/mondo-edit.obo
```diff
@@ -584963,6 +584963,21 @@ is_a: MONDO:0003767 {source="NCIT:C50888", source="https://orcid.org/0009-0009-6
...
+synonym: "paroxymsal AFib" EXACT [NCIT:C80391]
+synonym: "paroxysmal AF" EXACT [NCIT:C80391]
+xref: ICD10CM:I48.0 {source="MONDO:equivalentTo"}
+xref: icd11.foundation:542703670 {source="MONDO:equivalentTo"}
+xref: MESH:D001281 {source="MONDO:relatedTo"}
+xref: NCIT:C80391 {source="MONDO:equivalentTo"}
+xref: SCTID:282825002 {source="MONDO:equivalentTo"}
+is_a: MONDO:0004981 {source="https://orcid.org/0009-0009-6791-365X"} ! atrial fibrillation
```
sabrinatoro, line 584977: Please use the PMID as a source, and/or some of the x-ref (as appropriate)
Also same question as above about the orcid

sabrinatoro (changes requested):
Please see comments

src/ontology/mondo-edit.obo
twhetzel: updated

src/ontology/mondo-edit.obo
twhetzel: updated

sabrinatoro (approved):

Issue #7637, opened 2024-04-25
[Revise subclass] congenital myasthenic syndrome 12 (MONDO:0012518)


ValWood:
**Mondo term (ID and Label)**

congenital myasthenic syndrome 12 (MONDO:0012518)
Any congenital myasthenic syndromes with glycosylation defect in which the cause of the disease is a mutation in the GFPT1 gene.

should be is_a
is_a
"disorder of glycosylation" (MONDO:0024322)

**Suggested revision and reasons**

**Your nano-attribution (ORCID)**
If you don't have an ORCID, you can sign up for one [here](https://orcid.org/)

sagehrke:
Dear @ValWood,
Thank you for reaching out to the Mondo Disease Ontology, part of the [Monarch Initiative](https://monarchinitiative.org/)!

We have received your ticket and will review it as soon as possible. 

If you have any additional information or updates relevant to your inquiry, please add them in a comment on this ticket. Your input is valuable in helping us understand and resolve the issue effectively.

Thank you for your patience and for your contributions to Mondo! 

Sincerely,
The Mondo Team

PR #8198, opened 2024-09-24
Add new parent term to congenital myasthenic syndrome 12

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7637

Confirmed new parent suggestion from description in https://omim.org/entry/610542 and icd11 description.


src/ontology/mondo-edit.obo
```diff
@@ -284615,9 +284615,11 @@ xref: Orphanet:353327 {source="OMIM:610542"}
 xref: Orphanet:590 {source="OMIM:610542"}
 xref: UMLS:C3552335 {source="MONDO:equivalentTo", source="MEDGEN:765249", source="MONDO:MEDGEN"}
 is_a: MONDO:0000182 {source="DC-OMIM:610542", source="OMIM:610542"} ! congenital myasthenic syndrome with tubular aggregates
+is_a: MONDO:0024322 {source="https://orcid.org/0000-0001-6330-7526"} ! disorder of glycosylation
```
sabrinatoro, line 284618: If you confirmed the SC with OMIM and ICD11, please add these x-ref as a source.

sabrinatoro (changes requested):
Please see comment about source for SC

src/ontology/mondo-edit.obo
twhetzel: added omim and pubmed references

sabrinatoro (approved):
I approve and merge

PR #8213, opened 2024-09-27
added orcid

sabrinatoro:
Following the discussion on slack, I am adding the user's orcid which I had previously requested to be removed.


twhetzel (approved):
This looks fine, but there are other new term submissions from [the user that Yousif curated](https://github.com/monarch-initiative/mondo/issues?q=is%3Aissue+author%3Anklsbckmnn+assignee%3Ayshwetar) that may also need this orcid.

Issue #7742, opened 2024-05-27
Request for new synonym Parkinson Disease MONDO:0005180


dilshans2k:
### Mondo term

Parkinson disease MONDO:0005180

### Synonyms

PD

### Synonym type

exact

### Additional comments

Currently `PD` and `LOPD` is linked to `late-onset Parkinson disease` `MONDO:0008199`, but only `LOPD` is required to be linked with `MONDO:0008199` and `PD` should be linked with `MONDO:0005180`

dilshans2k:
Anyone with the medical expertise can shed some light as to how to handle acronyms better?
For eg:
Searching `LOPD` on google gives `late onset pompe disease`, but in `MONDO` ontology, it is linked to `late onset parkinson disease`

sagehrke:
Dear @dilshans2k,
Thank you for reaching out to the Mondo Disease Ontology, part of the [Monarch Initiative](https://monarchinitiative.org/)!

We have received your ticket and will review it as soon as possible.

If you have any additional information or updates relevant to your inquiry, please add them in a comment on this ticket. Your input is valuable in helping us understand and resolve the issue effectively.

Thank you for your patience and for your contributions to Mondo!

Sincerely,
The Mondo Team

PR #8240, opened 2024-10-15
add exact synonym abbreviation for MONDO:0005180 'Parkinson disease'

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7742

The user requested the synonym 'PD' for MONDO:0005180 'Parkinson disease'. 

They also mentioned that 'PD' does not need to be on MONDO:0008199 'late-onset Parkinson disease', however the  abbreviation 'PD' is from the [OMIM](https://omim.org/entry/168600) record so I left it as-is. However, I can see some arguments to mark 'PD' as excluded (despite only 12 of 600 excluded synonyms are also exact synonyms). Let me know if you prefer that EXCLUDE be added to 'PD' on MONDO:0008199 'late-onset Parkinson disease'.


sabrinatoro (commented):
@twhetzel 
This looks good. However, based on the comment, LOPD should also be added as an abbreviation for late onset pompe disease. 

Also, please make sure you answer the requester's answer about acronyms.

src/ontology/mondo-edit.obo
```diff
@@ -417738,8 +417739,8 @@ synonym: "GSD type 2, late onset" EXACT [Orphanet:420429]
 synonym: "GSD type 2, late-onset" EXACT [Orphanet:420429]
 synonym: "GSD type II, late onset" EXACT [Orphanet:420429]
 synonym: "GSD type II, late-onset" EXACT [Orphanet:420429]
+synonym: "LOPD" EXACT ABBREVIATION [Orphanet:420429]
 synonym: "Pompe disease, late onset" EXACT [Orphanet:420429]
-synonym: "Pompe disease, late-onset" EXACT [Orphanet:420429]
```
twhetzel, line 417742: I removed the duplicate synonym

sabrinatoro: This synonym is not duplicated. There is a dash between "late" and "onset", which is not in the other synonym. I know it looks like a detail, but if people use a dash in their search, they might not find this term. 
Please reinstate this synonym.

sabrinatoro (changes requested):
@twhetzel 
There is one minor change request.

sabrinatoro (commented):
@twhetzel Approving and merging. 
Please make sure to answer the user's question about how we handle acronyms.

sabrinatoro (approved):

Issue #7664, opened 2024-05-02
Revise NF2 to NF2-related schwannomatosis 


allaway:
**Mondo term (ID and Label):**
MONDO:0007039 neurofibromatosis type 2

**Suggested new label:**
NF2-related schwannomatosis 

**Your nano-attribution (ORCID)**
0000-0003-3573-3565

**_Optional_: Any additional information (like supporting evidence, PubMed ID, etc.)**
Renamed in 2022 along with "schwannomatosis": https://www.sciencedirect.com/science/article/pii/S1098360022007730

sagehrke:
Dear @allaway,
Thank you for reaching out to the Mondo Disease Ontology, part of the [Monarch Initiative](https://monarchinitiative.org/)!

We have received your ticket and will review it as soon as possible. 

If you have any additional information or updates relevant to your inquiry, please add them in a comment on this ticket. Your input is valuable in helping us understand and resolve the issue effectively.

Thank you for your patience and for your contributions to Mondo! 

Sincerely,
The Mondo Team

allaway:
Hello! I just wanted to follow up on this request. Thank you.

twhetzel:
Hi @allaway - we are working on this ticket now so it can be included in the next Mondo release.

allaway:
Thank you @twhetzel!

PR #8241, opened 2024-10-15
Relabel request for MONDO:0007039 neurofibromatosis type 2

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7664

The user requested that the term label be updated. However, in reading over the [OMIM entry](https://omim.org/entry/101000) despite that OMIM entry references the same article the user mentions, [Updated diagnostic criteria and nomenclature for neurofibromatosis type 2 and schwannomatosis: An international consensus recommendation](https://pubmed.ncbi.nlm.nih.gov/35674741/), OMIM says "Because the schwannomatosis syndrome related to NF2 is usually characterized by bilateral vestibular schwannomas (formerly called acoustic neuromas), OMIM uses `vestibular schwannomatosis` as the preferred designation for the disorder." Therefore I changed the name to 'vestibular schwannomatosis' and added the OMIM abbreviation for this label. The term label the user suggested already exists as a synonym.


src/ontology/mondo-edit.obo
```diff
@@ -140186,7 +140186,7 @@ is_a: MONDO:0003847 {source="https://orcid.org/0000-0001-5208-3432"} ! hereditar
 
 [Term]
 id: MONDO:0007039
-name: neurofibromatosis type 2
```
sabrinatoro, line 140189: This previous name should be added as an exact synonym so that users can still find this term using this name. (note that there should be an x-ref for this exact synonym)

sabrinatoro:
1) @twhetzel, as we previously discussed, please add the notes about your research and decision(s) in the main issue, so the requester and other users can see it and have a chance to respond to it. 

2) Even though we usually follow OMIM nomenclature, in this case, I think we should update the label. 
Based on the paper above, "The process extended over 3 years, and included a wide array of NF specialists, nonspecialists, patient advocacy groups, patients, and family members. (...) The Delphi process worked well in both reaching consensus in some areas and highlighting the issues with varying opinions for discussion at the NYC New York City meeting."
Therefore it looks like the community as a whole agreed on this new official name. I would personally advocate for using "NF2-related schwannomatosis" as a primary label as suggested. 
@nicolevasilevsky can you please share your thoughts on this?

twhetzel:
1 - I wanted to get your feedback on this before writing in the issue in this case. I can ask you these types of questions (e.g. is it ok to not follow Mondo naming policy in this case for this request) in issues in the future if that is where you prefer these discussions to be.
2 - Following the curation call, I will re-name this to "NF2-related schwannomatosis" and add the OMIM label as a synonym.

sabrinatoro (approved):
@twhetzel 
I am approving this PR.
While the issue reported has been resolved, I am finding some inconsistencies in the synonyms: e.g. neurofibromatosis type 2 and ~type II (same names) are both exact and related. This should be reviewed either here or in another PR. Feel free to either merge or make additional changes to this PR. 
Screenshot 2024-10-17 at 12 40 00 PM

PR #8252, opened 2024-10-16

twhetzel:


twhetzel:
Closing since a new qc check is now added to `src/ontology/config/robot_report_external_content.txt` in mondo-ingest branch `externalclingenmedgenefo` so that this error will be identified and the content causing the error will be removed to the relevant report file. I will re-run the pipeline.

PR #8254, opened 2024-10-17

twhetzel:


twhetzel:
I had the wrong directory path in the `src/ontology/config/robot_report_external_content.txt` file in mondo-ingest in the branch `externalclingenmedgenefo`. That is fixed now so I'll run the EMC pipeline again.

PR #8257, opened 2024-10-17

twhetzel:


src/ontology/mondo-edit.obo
```diff
@@ -367554,6 +367554,7 @@ xref: SCTID:44423001 {source="MONDO:equivalentTo"}
 xref: UMLS:C0270855 {source="MEDGEN:124373", source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 is_a: MONDO:0000412 {source="DOID:308"} ! neonatal period electroclinical syndrome
 is_a: MONDO:0005560 {source="DOID:308/inferred", source="EFO:1001900/inferred", source="NCIT:C116593"} ! brain disorder
+is_a: MONDO:0016801 {source="Orphanet:1935"} ! mitochondrial substrate carrier disorder
```
sabrinatoro, line 367557: @twhetzel this SC has been excluded in Mondo (see line 367559) and therefore, it should not have been added. 
There is a bug in the pipeline which should be updated. 
How do you want to proceed? 
1) update the code and rerun this
2) manually remove this incorrect SC annotation, merge this PR, update the code in a separate PR before the next time we update the SC.

sabrinatoro (commented):
@twhetzel 
please see my comment in line, and let me know how you would like to proceed.

src/ontology/mondo-edit.obo
twhetzel: For now I'd manually make the change to fix the incorrect SC annotation.

twhetzel: I have now removed this and submitted a ticket for [mondo-ingest](https://github.com/monarch-initiative/mondo-ingest/issues/680) about this.

sabrinatoro (approved):
looks good. Approving and Merging.

PR #8262, opened 2024-10-22
update imports

twhetzel:
I refreshed the imports following this [page](https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/) and updating the 'imports/manual_seed.txt' file using the _new_ content from https://docs.google.com/spreadsheets/d/1UME3pTeR42hwNt1I6RPl0nvJKVfTIeJqhx80laAlD50/edit?gid=0#gid=0.

I had to use 60GB of memory in order to get this to complete without getting an Out of Memory Error cc: @matentzn

### DIFF
--- a/src/ontology/imports/hgnc_terms.txt
+++ b/src/ontology/imports/hgnc_terms.txt
@@ -1473,6 +1473,7 @@ http://identifiers.org/hgnc/20856
 http://identifiers.org/hgnc/20858
 http://identifiers.org/hgnc/20859
 http://identifiers.org/hgnc/20862
+http://identifiers.org/hgnc/2088
 http://identifiers.org/hgnc/20889
 http://identifiers.org/hgnc/20893
 http://identifiers.org/hgnc/20914
--- a/src/ontology/imports/manual_seed.txt
+++ b/src/ontology/imports/manual_seed.txt
@@ -197,5 +197,6 @@ http://identifiers.org/hgnc/16850
 http://identifiers.org/hgnc/1462
 http://identifiers.org/hgnc/17735
 http://identifiers.org/hgnc/9247
-
+http://identifiers.org/ncbigene/100127354
+http://identifiers.org/hgnc/2088
 


src/ontology/mondo-edit.obo
```diff
@@ -570873,7 +570873,7 @@ synonym: "coat color albinism, oculocutaneous type IV, Japanese medaka" RELATED
 xref: OMIA:001821-8090 {source="MONDO:equivalentTo"}
 is_a: MONDO:1011321 {source="OMIA:001821-8090", source="https://orcid.org/0000-0002-5002-8648"} ! hereditary disease, non-human animal
 is_a: MONDO:1012579 {source="OMIA:001821-8090", source="https://orcid.org/0000-0002-5002-8648"} ! oculocutaneous albinism, non-human animal
-relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/ncbigene/100049347 {source="OMIA:001821-8090", source="https://orcid.org/0000-0002-5002-8648"}
+relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/ncbigene/100049347 {source="OMIA:001821-8090", source="https://orcid.org/0000-0002-5002-8648"} ! slc45a2
```
twhetzel, line 570876: Confirmed that this ncbigene UID does have a lowercase label: https://www.ncbi.nlm.nih.gov/gene/?term=100049347

twhetzel:
I did run `git pull origin master` before creating this feature branch so I am not sure why there would be content that would not pass the qc check:
```
FAIL Rule ../sparql/qc/general/qc-single-child.sparql: 1 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0004548,http://www.w3.org/2000/01/rdf-schema#subClassOf,http://purl.obolibrary.org/obo/MONDO_0020541
```
Looking at Mondo in Protege, I am also not seeing a single child class.

twhetzel:
NOTE: The qc check failed on the first push:
```
FAIL Rule ../sparql/qc/general/qc-single-child.sparql: 1 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0004548,http://www.w3.org/2000/01/rdf-schema#subClassOf,http://purl.obolibrary.org/obo/MONDO_0020541
```

I did pull from `master` before creating the feature branch and refreshing the imports so I am not sure why this error is only happening now. I did add an 'exclude from qc' check for single_child and pushed that change.

twhetzel:
NOTE: Also testing out a new PR for refreshing the imports https://github.com/monarch-initiative/mondo/pull/8264

twhetzel:
The [new PR](https://github.com/monarch-initiative/mondo/pull/8264) is returning the same error so I will close this older PR.

PR #8264, opened 2024-10-22
refresh imports 22-oct-2024

twhetzel:
Testing out a new PR for refreshing the imports. I ran the refresh imports on a fresh clone of Mondo to verify this is happening as a result of refreshing the imports. The same errors as found originally occur. See the error log [issue here](https://github.com/monarch-initiative/mondo/actions/runs/11467370674/job/31910098595?pr=8264#step:4:83).

Note: this error is also posted as it's own [issue](https://github.com/monarch-initiative/mondo/issues/8268).

### DIFF
--- a/src/ontology/imports/cl_terms.txt
+++ b/src/ontology/imports/cl_terms.txt
@@ -766,6 +766,7 @@ http://purl.obolibrary.org/obo/CL_0008035
 http://purl.obolibrary.org/obo/CL_0009000
 http://purl.obolibrary.org/obo/CL_0009001
 http://purl.obolibrary.org/obo/CL_0009004
+http://purl.obolibrary.org/obo/CL_0009005
 http://purl.obolibrary.org/obo/CL_0009006
 http://purl.obolibrary.org/obo/CL_0009018
 http://purl.obolibrary.org/obo/CL_0009019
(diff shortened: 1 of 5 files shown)


twhetzel:
Two imports were not previously in the [spreadsheet](https://docs.google.com/spreadsheets/d/1UME3pTeR42hwNt1I6RPl0nvJKVfTIeJqhx80laAlD50/edit?gid=0#gid=0) so they were missed in this refresh of the imports (luckily Sabrina caught this!). I will refresh the imports again with these new entities.

twhetzel:
I re-ran the refresh-merged import pipeline with the new additions to the manual_seed.txt file, but that did not seem to actually import these two entities since they still show up as numbers. Other '*_terms.txt' files were updated this time though.

src/ontology/mondo-edit.obo
```diff
@@ -83836,6 +83836,7 @@ xref: UMLS:C1515284 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source=
 is_a: MONDO:0003395 {source="DOID:8394", source="NCIT:C39946"} ! testicular granulosa cell tumor
 intersection_of: MONDO:0003395 ! testicular granulosa cell tumor
 intersection_of: has_characteristic HP:0003581 ! Adult onset
+relationship: excluded_from_qc_check http://purl.obolibrary.org/obo/mondo/sparql/qc/general/qc-single-child.sparql
```
sabrinatoro, line 83839: @twhetzel - this single child excluded_from_qc_check is not necessary since this term is not a single child. But it is not "hurting" anything. 
At this point, I think it is ok to keep it as is so we can merge this PR. 
Let's remove it in a separate PR though.

sabrinatoro (approved):
@twhetzel 
I reviewed this PR. I am approving and merging. 
Please see the comment about the single child excluded_from_qc_check.

src/ontology/mondo-edit.obo
twhetzel: Ok, in the first run of the qc checks this was an issue. As I look at it again, this was the sign of bigger problems to come (the logical issue you resolved).

PR #8269, opened 2024-10-23
Adding the pipeline for synchronising confirmed synonym evidence

matentzn:
Build:
- #8270


src/ontology/mondo.Makefile
```diff
@@ -1298,6 +1302,24 @@ update-subclass-sync: $(TMPDIR)/subclass-confirmed.robot.owl tmp/subclass-axioms
 		make NORM
 		mv NORM $(SRC)
 
+tmp/synonyms-axioms.owl: $(SRC)
+	$(ROBOT) filter --input $(SRC) --term oboInOwl:hasExactSynonym --axioms annotation --preserve-structure false --trim false \
+		--drop-axiom-annotations "oboInOwl:source=~'(DOID|ICD10CM|icd11.foundation|NCIT|OMIM|OMIMPS|Orphanet):.*'" \
+		-o $@
```
twhetzel, line 1308: why does `drop-axiom-annotations` not include all sources in [sync-synonyms.confirmed.robot.tsv](https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/refs/heads/sync1-synonyms/src/ontology/reports/sync-synonym/sync-synonyms.confirmed.robot.tsv)

what about confirmed narrow and related synonyms?

src/ontology/mondo.Makefile
```diff
@@ -1298,6 +1302,24 @@ update-subclass-sync: $(TMPDIR)/subclass-confirmed.robot.owl tmp/subclass-axioms
...
+
+# This command updates mondo-edit with all the confirmed synonyms evidence from the mondo-ingest repo
+.PHONY: update-synonyms-sync
+update-synonyms-sync: $(TMPDIR)/synonyms-confirmed.robot.owl tmp/synonyms-axioms.owl
+	grep -vE '^synonym:.*EXACT' $(SRC) > tmp/$(SRC)
```
twhetzel, line 1313: why only EXACT?

twhetzel (commented):
comments inline

src/ontology/mondo.Makefile
```diff
@@ -1298,6 +1302,24 @@ update-subclass-sync: $(TMPDIR)/subclass-confirmed.robot.owl tmp/subclass-axioms
 		make NORM
 		mv NORM $(SRC)
 
+tmp/synonyms-axioms.owl: $(SRC)
+	$(ROBOT) filter --input $(SRC) --term oboInOwl:hasExactSynonym --axioms annotation --preserve-structure false --trim false \
+		--drop-axiom-annotations "oboInOwl:source=~'(DOID|ICD10CM|icd11.foundation|NCIT|OMIM|OMIMPS|Orphanet):.*'" \
```
joeflack4, line 1307: Never seen this syntax before!

Guess there's no easy way to DRYify this and interpolate these prefixes from elsewhere.

matentzn: Yes this is a new ROBOT feature..

I would have liked to dryify this as well but I cant see an easy way right now.. somewhere there would have to live a source of truth for the resources that we want to sync with in this pipeline, so I thought to myself, might as well be here, but yes, its not DRY

joeflack4: @matentzn IC10WHO is missing here. Not sure if anything else, but you should double check the prefix_maps in the metadata files in `mondo-ingest` if you want to double check for any other missing ones.

RE: DRY: 1st step: Maybe could then add these sources to a global variable and interpolate it? Not too important at this point.

matentzn: > RE: DRY: 1st step: Maybe could then add these sources to a global variable and interpolate it? Not too important at this point.

Yeah, good point, did that! Immediately refactored the respective code for the subclass sync as well. Definetely DRYer!

src/ontology/mondo.Makefile
matentzn: My bad, good we have reviews.. Fixed now!

src/ontology/mondo.Makefile
matentzn: > why does drop-axiom-annotations not include all sources in [sync-synonyms.confirmed.robot.tsv](https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/refs/heads/sync1-synonyms/src/ontology/reports/sync-synonym/sync-synonyms.confirmed.robot.tsv)

Did I miss any apart from ICD10WHO? it was just an oversight, my bad.. All should be there now.

> what about confirmed narrow, related, or broad synonyms?

All fixed now, thank you!

Note there are a NO broad synonyms coming in from the sync.. Maybe check if its true that there should be zero broad synonyms in the syn.

matentzn: Can any of these ever end up in the synonym sync? Theoretically?

joeflack4: I'll say with 95+% certainty that no. I think the icd11 ones get filtered out in the signature step. `ICD10WHO2010` I've never seen.

src/ontology/mondo.Makefile
```diff
@@ -1299,6 +1303,44 @@ update-subclass-sync: $(TMPDIR)/subclass-confirmed.robot.owl tmp/subclass-axioms
...
+		--term oboInOwl:hasRelatedSynonym \
+		--axioms annotation --preserve-structure false --trim false \
+		--drop-axiom-annotations "oboInOwl:source=~'($(SOURCES_REGEX)):.*'" \
+		-o $@
+
+# This command updates mondo-edit with all the confirmed synonyms evidence from the mondo-ingest repo
+.PHONY: update-synonyms-sync
+update-synonyms-sync: $(TMPDIR)/synonyms-confirmed.robot.owl tmp/synonyms-axioms.owl
+	$(ROBOT) \
+		remove --input $(SRC) \
```
joeflack4, line 1328: ## Dropping?
@matentzn Can you explain this part?

It looks like it is dropping all existing synonym annotations before re-adding the confirmed ones and their source annotations?

If it is really dropping all of them, aren't there situations in which we have a synonym or different scope that is the result of manual curation and we don't want to destroy that?

src/ontology/mondo.Makefile
joeflack4, line 1: ## General review & results
I think the main approver for this one is Trish, but I guess I'll hold off on my approval for now because I think I would like to see the results of running this command. Sort of like how we don't merge in `mondo-ingest` until we see a build.

Is that something we can do?

src/ontology/mondo.Makefile
matentzn: You need to check this target: tmp/synonyms-axioms.owl; it contains all (!) synonyms in mondo-edit, with their provenance removed. That is also being merged back in! So this process will never drop anything!

joeflack4: #### 1. What do you mean by _(!)_?

#### 2. Unconfirmed source not being dropped?
I did look at that target, but is it doing what it is supposed to?

Observe this case. Checking the diff in #8270, `MONDO:0000004`, this line was unmodified:
`synonym: "adrenal cortical insufficiency" EXACT [DOID:10493, NCIT:C26691, NCIT:C27150]`

But in the [confirmed cases](https://docs.google.com/spreadsheets/d/1ftEZ47-JDe-1Z_fRWyO5NS39qT_-8MZsHXHEKz9aZ9Q/edit?gid=978570179#gid=978570179), `NCIT:C27150` is not a confirmed case:

|mondo_id     |mondo_label                 |synonym_scope_source|synonym_scope_mondo|synonym                       |source_id  |
|-------------|----------------------------|--------------------|-------------------|------------------------------|-----------|
|MONDO:0000004|adrenocortical insufficiency|oio:hasExactSynonym |                   |adrenal cortical insufficiency|DOID:10493 |
|MONDO:0000004|adrenocortical insufficiency|oio:hasExactSynonym |                   |adrenal cortical insufficiency|NCIT:C26691|

So shouldn't provenance for `NCIT:C27150` be removed?

matentzn: Oooppppsss. see https://github.com/monarch-initiative/mondo/commit/b87630ec7d56d82aeaf28e2f06e6715d393d63d5

Copy paste error :D THANKS for seeing that, thank Open Science for reviews!!

matentzn: (Updated the build as well)

src/ontology/mondo.Makefile
joeflack4: Non-confirmed sources are dropped now! Everything looks good now to me.

src/ontology/mondo.Makefile
joeflack4: Thanks! Resolving.

joeflack4 (approved):

.github/workflows/synonyms.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+#  schedule:
+#    - cron: '0 14 * * 1'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.5.2
```
twhetzel, line 11:
```suggestion
    container: obolibrary/odkfull:v1.5.3
```

src/ontology/mondo.Makefile
twhetzel: Since the synonym sync uses the sssom file and then only works with the sources there are components for in mondo-ingest it's limited to that.

src/ontology/mondo.Makefile
joeflack4: Not gonna drop my approval, but just FYI there are [QC failures in the build branch](https://github.com/monarch-initiative/mondo/actions/runs/11713965163/job/32627744719?pr=8270).

twhetzel (approved):
LGTM

NOTE: Will need to change [this line](https://github.com/monarch-initiative/mondo/pull/8269/files#diff-4a43f792a4363fc6010b30eee1955eaeb517bc4b20911038c950f100fecaf2b0R1269) once synonym sync pipeline is in `main` in mondo-ingest.

Issue #8288, opened 2024-10-29
Update ODK to version 1.5.3


twhetzel:
There is a new release of ODK (1.5.3) and the mondo repo should update to this new version. The ODK version needs to be updated [here](https://github.com/monarch-initiative/mondo/blob/master/src/ontology/run.sh#L14).

- NOTE: Sabrina will need to do a mock release of Mondo using the new ODK version before merging the PR with this change into master

PR #8289, opened 2024-10-30
update odk version to 1.5.3

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8288

@sabrinatoro - a mock mondo release should be done to test this update to ODK v1.5.3 before merging this branch

### DIFF
--- a/src/ontology/run.sh
+++ b/src/ontology/run.sh
@@ -11,7 +11,7 @@
 # See README-editors.md for more details.
 
 MEMORY_GB=${MEMORY_GB:-8}
-IMAGE=${IMAGE:-odkfull:v1.5.2}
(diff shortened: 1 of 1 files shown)


sabrinatoro:
@twhetzel should this be merged for this upcoming release? Can we merge after the November release instead? There won't be enough time for a mock release before then.

twhetzel:
I missed your comment. It's fine to do after the release :)

sabrinatoro (approved):
It looks good to me-looks simple, however if something more should be done, I wouldn't know :-)

twhetzel:
@sabrinatoro as long as a mock release still works then all should be good.

I have another ticket to update the ODK version in the GH Action so I'll aim to do that today as well

PR #8291, opened 2024-10-30
Add QC check to look for excluded subclassof axioms which are asserted

### LINKED ISSUE #8290: Add QC check - class with excluded_subClassOf value does not also have a subClassOf statement
Add QC check to catch cases where (due to timing of content creation) a class with excluded_subClassOf relationship is not also stated to be a subclass of that excluded class as a result of the SubClass sync pipeline. 

This issue was discussed [here](https://github.com/monarch-initiative/mondo-ingest/issues/680) and determined that it was not an issue with the SubClass sync pipeline, but of the timing of when data was created in mondo-ingest vs. the latest changes in the mondo repo.

matentzn:
Fixes #8290 

This tests a case where we said, deliberately:

This subclass axiom should not exist, but it turns out, it does!

There are quite a few cases like this; good call to look for that check, I would have thought that existed!


twhetzel (approved):
Queries looks good - thanks Nico!

I will ask if there is another curator who can review the QC errors now caught by this check.

twhetzel:
@sabrinatoro can we discuss or is there someone else that can help sort out the remaining qc issues?

twhetzel:
- qc-excluded-subclass-is-asserted.sparql: 34 violation(s) - I looked at a few and these are cases where the excluded class has a logical definition. Did the query to clean this up take that condition into account?
- qc-no-superclass.sparql: 4 violation(s)
- qc-single-child.sparql: 2 violation(s) --> I added the exclude from qc check for these two terms

A curator will need to fix the other issues.

Issue #8312, opened 2024-11-06
MONDO:0009054 'autosomal recessive cutis laxa type 2, classic type' has incorrect synonym


twhetzel:
This term has an incorrect synonym, ADCL2, as pointed out by Chris M. at the curation call today since AD references autosomal dominance and this term is called 'autosomal recessive....'

The referenced [DOID term](https://www.ebi.ac.uk/ols4/ontologies/doid/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FDOID_0070136) is 'autosomal dominant cutis laxa 2', clearly different than 'autosomal recessive cutis laxa type 2, classic type'.

PR #8313, opened 2024-11-07
Remove incorrect synonym

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8312

- remove incorrect synonym based on review of [DOID:0070136](https://www.ebi.ac.uk/ols4/ontologies/doid/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FDOID_0070136) term


sabrinatoro (approved):
I think it is ok. However, this synonym comes from the xref: DOID:0070136 {source="MONDO:relatedTo"}.  It would make sense to remove this x-ref too. 
@twhetzel feel free to either merge this PR as is and remove the x-ref in another PR, or update this PR. Whatever you prefer. I am approving, so you are not stuck.

twhetzel:
Good call! I'll remove the xref here too.

Issue #8276, opened 2024-10-26
Multiple gene associations represented as EquivalentTo


twhetzel:
- MONDO:0011067 'autosomal recessive nonsyndromic hearing loss 12'  https://omim.org/entry/601386
  - remove gene associations since >1 gene associated and one of these is in [{}](https://omim.org/help/faq#1_6) 
- MONDO:0011974 'retinitis pigmentosa 7' https://omim.org/entry/608133 --> This is CORRECT since it's digenic.
  -  Leave as-is  
- MONDO:0012157 'congenital myasthenic syndrome 4C' https://omim.org/entry/608931 
  - keep only the 1 gene that OMIM lists
- MONDO:0014866 'Charcot-Marie-Tooth disease axonal type 2T' https://omim.org/entry/617017 
  - keep only the 1 gene that OMIM lists
- MONDO:0100050 'Usher syndrome, type 1D/F' https://omim.org/entry/601067 --> This is CORRECT since it's digenic!

Results from SPARQL query of Mondo Oct2024 release, followed by reading the SPARQL results into Pandas dataframe and getting Mondo IDs that occurred more than once in the SPARQL results.

SPARQL query
```
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX obo: <http://purl.obolibrary.org/obo/>
PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>

SELECT DISTINCT ?mondoCurie ?roProperty ?hgncIdentifier (COALESCE(?source, "No Source") AS ?sourceAnnotation)
WHERE {
  ?class owl:equivalentClass ?equivClass .

  ?equivClass owl:intersectionOf/rdf:rest*/rdf:first ?component .
  
  ?component rdf:type owl:Restriction ;
             owl:onProperty obo:RO_0004003 ;
             owl:someValuesFrom ?hgncIdentifier .

  FILTER (STRSTARTS(STR(?hgncIdentifier), "http://identifiers.org/hgnc/"))

  # Retrieve source annotation if available
  OPTIONAL {
    ?annotation a owl:Axiom ;
                owl:annotatedSource ?class ;
                owl:annotatedProperty obo:RO_0004003 ;
[...]

PR #8315, opened 2024-11-07
Fix incorrect multiple gene associations

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8276

See ticket for reasons for updates


src/ontology/mondo-edit.obo
```diff
@@ -275865,11 +275862,9 @@ xref: UMLS:C1837091 {source="MEDGEN:373251", source="MONDO:equivalentTo", source
 is_a: MONDO:0018940 {source="DOID:0110679", source="MONDO:Redundant", source="OMIM:608931"} ! congenital myasthenic syndrome
 is_a: MONDO:0020344 {source="MONDO:Redundant", source="Orphanet:98913/btnt"} ! postsynaptic congenital myasthenic syndrome
 is_a: MONDO:1040021 {source="https://orcid.org/0000-0002-9698-992X"} ! congenital myasthenic syndrome 4
-intersection_of: MONDO:0018940 ! congenital myasthenic syndrome
-intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/1966 ! CHRNE
-intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/4241 ! GFPT1
```
twhetzel, line 275870: Sabrina - removing the incorrect relationship `intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/4241 ! GFPT1` then caused an [error with the qc check for omim subsumption](https://github.com/monarch-initiative/mondo/actions/runs/11716038830/job/32633442177#step:4:32). The reasoner was showing that the '4C' entry was a parent of '4A' and '4B', which did not look correct so that is partly why I removed the logical statement and moved the gene association relationship to the SubClassOf. There are some other ways this could be resolved, please let me know if this should be handled differently.

sabrinatoro (approved):

PR #8325, opened 2024-11-11
Extract only the relevant gene entries from ncbigene dump

matentzn:
We use this pattern:

1. Given the terms of ncbigenes in the seed ($(IMPORTDIR)/merged_terms_combined.txt), only extract those from the triples dump to make import refresh much faster (avoid loading the dump into OWL API)
2. The seed (genes) are pasted into a sparql template, which materialises as a sparql query
3. The query is used to extract the ncbigene mirror. The good thing is that there was already such a query (for another reason).

This idea rests on the fact that

```
arq --data=mirror/ncbi_gene.nt --query=
```

Is much much more efficient then ROBOT (which I am pretty sure it is)

@twhetzel I am testing this, but feel free to test it yourself.. Memory needs should be back below 48 GB now.


twhetzel:
I ran `sh run.sh make refresh-merged` from this branch and it completed (MEMORY_GB=50, time taken=18 min).
BUT, I don't see the `manual_seed.txt` file in this branch, which is the file we currently add identifiers for entities to import following these instructions https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/

twhetzel (approved):

twhetzel:
@matentzn this is working well for me. If you are done your testing can you merge?

Issue #8329, opened 2024-11-12
Updates to fix QC errors for multiple gene associations


twhetzel:
Updates to fix or add qc exclusions for new qc check failures for https://github.com/monarch-initiative/mondo/issues/8316

FAIL Rule ../sparql/qc/mondo/qc-multiple-gene-associations.sparql: 4 violation(s)
entity,label,geneIdentifiers
- [x]  http://purl.obolibrary.org/obo/MONDO_0011974,retinitis pigmentosa 7,"http://identifiers.org/hgnc/10254, http://identifiers.org/hgnc/9942" --> exclude from QC check, see https://github.com/monarch-initiative/mondo/issues/8276
- [x] http://purl.obolibrary.org/obo/MONDO_0100188,combined ApoA-I and ApoC-III deficiency,"http://identifiers.org/hgnc/600, http://identifiers.org/hgnc/610" --> see comment in post below
- [x] http://purl.obolibrary.org/obo/MONDO_0100050,"Usher syndrome, type 1D/F","http://identifiers.org/hgnc/13733, http://identifiers.org/hgnc/14674" --> exclude from QC check, see https://github.com/monarch-initiative/mondo/issues/8276
- [x] http://purl.obolibrary.org/obo/MONDO_0011908,juvenile myelomonocytic leukemia,"http://identifiers.org/hgnc/1541, http://identifiers.org/hgnc/9644, http://identifiers.org/hgnc/7765, http://identifiers.org/hgnc/17073" --> see comment in post below

twhetzel:
- MONDO_0100188 'combined ApoA-I and ApoC-III deficiency' - based on the xref for the definition (PMID: 3118360) this term seems more related to [OMIM:620058](https://www.omim.org/entry/620058), which does have a single disease defining gene so I removed the two gene associations. In addition, if we agree that MONDO_0100188 is the same as OMIM:620058 then there is already another MONDO term for this, ie MONDO:0859290.

katiermullen:
@twhetzel Thanks for all of your work on this!

It seems based on [this publication ](https://pmc.ncbi.nlm.nih.gov/articles/PMC6942922/)that these (MONDO:0100188 and MONDO:0859290) are different diseases:

[...]

PR #8330, opened 2024-11-12
Address cases of multiple gene associations

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8329

Please see ticket for explanation of changes


src/ontology/mondo-edit.obo
```diff
@@ -271426,6 +271422,7 @@ is_a: MONDO:1040055 {source="https://clinicalgenome.org/working-groups/clinical-
 intersection_of: MONDO:0019200 ! retinitis pigmentosa
 intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/10254 ! ROM1
 intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/9942 ! PRPH2
+relationship: excluded_from_qc_check http://purl.obolibrary.org/obo/mondo/sparql/qc/mondo/qc-multiple-gene-associations.sparql
```
sabrinatoro, line 271425: I don't think this is correct. 
According to [OMIM](https://www.omim.org/entry/608133?search=608133&highlight=608133), this record refers to the disease for which the RDS gene (PRPH2) is mutated, however, "A digenic form of retinitis pigmentosa resulting from a mutation in the RDS gene ([179605.0004](https://www.omim.org/entry/179605#0004)) and a null mutation of the ROM1 gene (see [180721.0001](https://www.omim.org/entry/180721#0001)) has been reported."

Therefore, for this record, we cannot say that in 100% of the case, the disease is caused by a mutation in both genes (or in a mutation in only one gene). 

The gene annotations should be removed for this disease. 

Note: the definition for this disease should also be changed to make this disease concept not specific to the mutation in the PRPH2 gene.

sabrinatoro (changes requested):
@twhetzel 
I think retinitis pigmentosa 7 should not have association with any genes. Please see my explanation in the comment. I can make the changes if you want, but I first want to make sure you have a chance you check my explanation in case you don't disagree.

src/ontology/mondo-edit.obo
twhetzel: Thanks Sabrina. I agree and will make the changes.

twhetzel: Updated

sabrinatoro (approved):
@twhetzel I updated the definition for retinitis pigmentosa 7.
I think this PR can be merged after QC have passed. 

[OMIM:608133](https://omim.org/entry/608133) indicates that there are 2 included entries. We should add these with the "omim included" x-ref in a different PR (note, i have not created an issue for this).

Issue #7641, opened 2024-04-26
Several issues


galyea123:
Hi Sabrina!  Following your suggestions I included several issues in the same document for this week:

**Duplicated term**
“Weiss-kruszka syndrome”  MONDO:0032836 is duplicated with “metopic ridging-ptosis-facial dysmorphism syndrome” MONDO:0044715
Please merge and add the respective synonyms.

**TAG the following condition as NORD rare and add synonym**
Condition: Neurodevelopmental-craniofacial syndrome with variable renal and cardiac abnormalities MONDO:0859190
Synonyms: NECRC,  ZMYM2-related neurodevelopmental disorder with multiple anomalies

**ADD a term**
We have a request to write a report about disorders related to DNM1 which will be entitled:  "DNM1-Encephalopathy and Neurodevelopmental Disorder".
OMIM list two conditions related to the DNM1 gene mutations:
[Developmental and epileptic encephalopathy 31A, autosomal dominant](https://omim.org/entry/616346) and [Developmental and epileptic encephalopathy 31B, autosomal recessive](https://omim.org/entry/620352). 
I only found developmental and epileptic encephalopathy 31B in MONDO. MONDO:0957248
I did not find the subtype 31A. 
Our request is to add the subtype “developmental and epileptic encephalopathy  31A" and tag NORD in both subtypes as rare as well to include the respective synonyms:
Developmental and epileptic encephalopathy 31A, autosomal dominant 
Developmental and epileptic encephalopathy 31B, autosomal recessive
NOTE: Our report in NORD entitled “DNM1-Encephalopathy and Neurodevelopmental Disorder” which will include both subtypes and therefore can be include as a broad synonym too.

Please let me know if this way is better or if you would rather I made a ticket for each issue.  

Thank you,

Gioconda Alyea

sagehrke:
Dear @galyea123,
Thank you for reaching out to the Mondo Disease Ontology, part of the [Monarch Initiative](https://monarchinitiative.org/)!

We have received your ticket and will review it as soon as possible. 

If you have any additional information or updates relevant to your inquiry, please add them in a comment on this ticket. Your input is valuable in helping us understand and resolve the issue effectively.

Thank you for your patience and for your contributions to Mondo! 

Sincerely,
The Mondo Team

galyea123:
[...]

PR #8356, opened 2024-11-12
Added Nord rare + Github tracker

yshwetar:
Addresses part of #7641


src/ontology/mondo-edit.obo
```diff
@@ -556240,6 +556240,7 @@ id: MONDO:0957248
 name: developmental and epileptic encephalopathy, 31B
 def: "Any developmental and epileptic encephalopathy in which the cause of the disease is a homozygous mutation in the DNM1 gene." [https://orcid.org/0000-0002-0587-4693, OMIM:620352]
 subset: gard_rare {source="MONDO:GARD"}
+subset: nord_rare {source="MONDO:NORD"}
```
sabrinatoro, line 556243: @yshwetar : while not "wrong", we do not manually add any rare subset annotations. 
All subsets are maintained automatically. Therefore anything you add manually will be removed when we update the subsets. You can see more information [here](https://mondo.readthedocs.io/en/latest/editors-guide/rare-disease-curator/)

It is ok to leave here, but just keep this in mind for the future.

sabrinatoro (approved):
This looks good

Issue #8265, opened 2024-10-22
DHX30-related complex neurodevelopmental disorder


Jciesielka1:
**Preferred gene-related syndrome label**
DHX30-related complex neurodevelopmental disorder

**Synonyms**
Neurodevelopmental disorder with variable motor and speech impairment (MIM 617804)

**Parent term (use [OLS](https://www.ebi.ac.uk/ols/ontologies/mondo), or your favorite ontology browser)**
Complex neurodevelopmental disorder (MONDO:0100038)

**Definition**
This is a complex neurodevelopmental disorder in which the cause of disease is variation in DHX30. Individuals with variants in DHX30 have been found to have variable presentations including intellectual development, delayed or absent speech development, delayed motor development, hypotonia, feeding difficulties, and ataxic gait or the inability to walk. Other phenotypic features commonly reported include sleep disorders, autistic features, seizures, and joint hypermobility (PMID: 29100085 and 3402070).

**Definition source (Please give PubMed ID, if applicable, in format PMID:#######)
PMID: 29100085 and 3402070

**Children terms (if applicable)** Should any existing terms be re-classified as children underneath this new proposed term? 
No child terms that we are aware of.

**Your nano-attribution (ORCID) or URL for a working group**
ClinGen Intellectual Disability and Autism Expert Panel (https://clinicalgenome.org/affiliation/40006/)

@ErinRiggs

PR #8369, opened 2024-11-13
New term DHX30-related complex neurodevelopmental disorder

yshwetar:
Addresses part of #8265

Need DHX30 gene in Ontology to make Equivalent To definition


sabrinatoro:
@yshwetar we already have a term (which the requester mentioned as a synonym): MONDO:0060622-'neurodevelopmental disorder with severe motor impairment and absent language'

Based on your research, does it look like the term requested is different from the term we already have? If we don't know, we should ask the requester. 
If the requested term is the same as MONDO:0060622 (= OMIM:617804), then we should add "DHX30-related complex neurodevelopmental disorder" as a synonym and ClinGen preferred label. We should also update the definition of the term and add "Complex neurodevelopmental disorder" (MONDO:0100038) as a parent.

yshwetar:
Started a new PR #8376 to address the issue #8265

PR #8380, opened 2024-11-13
NTR DHDDS-related syndrome

### LINKED ISSUE #8152: Request to add new gene related syndrome term for DHDDS
DHDDS-related syndrome

**Synonyms**
NA

**Parent term (use [OLS]
hereditary neurological disease MONDO:0100545

**Definition**
Any neurological disorder in which the cause of the disease is a mutation in the DHDDS gene.

PMID:34034154

**Children terms (if applicable)** Should any existing terms be re-classified as children underneath this new proposed term?
MONDO:0044326

0000-0002-9698-992X

katiermullen:
closes #8152

note: did not add child term as requested since it does not have all of the features of the parent term.


sabrinatoro:
@katiermullen: I think the gene association should be in the logical definition. 
I made the changes, please take a look and see if you agree. 

By adding the logical definition including the gene association, I see that there are 2 additional children. I think these are ok because the new term is only defined for being a neurological disorder and having a variation in the gene. 

Screenshot 2024-11-13 at 8 32 01 PM


So I don't follow why "did not add child term as requested since it does not have all of the features of the parent term."
Let's talk about this.

katiermullen:
@sabrinatoro Thank you for the review and comments. This makes total sense now. I misread the term MONDO:0044326 and I did not see that it had material basis in germline mutation in some DHDDS. This is my mistake.

When adding a gene, how do I know if I should add it as an EC, SC or both?

PR #8383, opened 2024-11-14
Update GNB1-related disorder

### LINKED ISSUE #8167: GNB1-Related Disorder
### Mondo term

intellectual disability, autosomal dominant 42

### Synonyms

_No response_

### Synonym type

exact

### Additional comments

This is related to 6600.  We have a NORD report about this condition with the requested synonym name. 
Thank you! Gioconda

yshwetar:
Closes #8167


sabrinatoro (commented):
@yshwetar : I don't know that this is correct. 
According to [OMIM:139380](https://www.omim.org/entry/139380), a variation in the GNB1 gene can lead to multiple diseases, including, but not limited to 'intellectual disability, autosomal dominant 42'. Therefore this synonym should not be exact. However, the other involvement of the GNB1 genes are for somatic mutation in cancer. 
Let's confirm with the requester.

PR #8386, opened 2024-11-14
Update Synonyms of cholesterol embolism

### LINKED ISSUE #8049: MONDO:0005568 cholesterol embolism
**Mondo term (ID Label)**
MONDO:0005568 cholesterol embolism

Some of the synonyms on this term seem like they are incorrectly mapped:
purple toe syndrome
trash foot
warfarin blue toe syndrome

Perhaps they would be better matched to MONDO:0001883 blue toe syndrome ?
FYI Trash foot is a separate concept in UMLS, but no definition is provided (CUI: C0585266) 

yshwetar:
Closes #8049

Cholesterol Embolism primarily occurs in the lower extremities, mainly the toes. However, can arise anywhere - thus 'purple toe syndrome', 'trash foot', and 'warfarin blue toe syndrome' were moved to be synonyms of 'Blue toe syndrome'.

Also added term tracker to both


src/ontology/mondo-edit.obo
```diff
@@ -31153,11 +31153,15 @@ property_value: IAO:0000589 "bacteriuria (disease)" xsd:string
 id: MONDO:0001883
 name: blue toe syndrome
 def: "A condition that is caused by recurring atheroembolism in the lower extremities. It is characterized by cyanotic discoloration of the toes, usually the first, fourth, and fifth toes. Discoloration may extend to the lateral aspect of the foot. Despite the gangrene-like appearance, blue toes may respond to conservative therapy without amputation." [MESH:D018438]
+synonym: "purple toe syndrome" EXACT [DOID:1461]
+synonym: "trash foot" EXACT [DOID:1461]
+synonym: "warfarin blue toe syndrome" EXACT [DOID:1416]
```
sabrinatoro, line 31158: @yshwetar these synonyms should not be on this term because DOID:1461 is not an equivalent mapping

src/ontology/mondo-edit.obo
```diff
@@ -108056,9 +108060,6 @@ def: "Blocking of a blood vessel by cholesterol-rich atheromatous deposits, gene
 subset: otar {source="MONDO:OTAR"}
 synonym: "atheroembolism" EXACT [DOID:1461]
 synonym: "cholesterol crystal embolism" EXACT [DOID:1461]
-synonym: "purple toe syndrome" EXACT [DOID:1461]
-synonym: "trash foot" EXACT [DOID:1461]
-synonym: "warfarin blue toe syndrome" EXACT [DOID:1461]
```
sabrinatoro, line 108061: I agree that these should not be exact, even though they are reported as synonym in DO. 
Instead of removing them, we should update the synonym scope to "narrow synonym" since the synonyms refers to a child term of this term.

sabrinatoro (commented):
@yshwetar please take a look at the comments and let me know if you need more explanations. I can make the updates in the ontology.

sabrinatoro (approved):
I made some changes to the original PR. 
I am approving and will merge after QC have passed.

PR #8387, opened 2024-11-14
Update autosomal recessive hyper-IgE syndrome due to ZNF341 deficiency

### LINKED ISSUE #7844: MONDO:0957426 autosomal recessive hyper-IgE syndrome
MONDO:0957426 autosomal recessive hyper-IgE syndrome

**Suggested new label:**
If this is, indeed an exact match to Orphanet 641368 , then the pref name should be Autosomal recessive hyper-IgE syndrome due to ZNF341 deficiency
and this would suggest it is the same concept as OMIM 618282
HYPER-IgE SYNDROME 3, AUTOSOMAL RECESSIVE, WITH RECURRENT INFECTIONS (associated with AR mutation in ZNF341)

BUT the Orphanet term maps to multiple OMIM records, https://www.orpha.net/en/disease/detail/641368 which suggests that there may be some error with the Orphanet record either in name or mappings?
It seems like either it should be broader and generic in name for all AR hyper-IgE syndromes, as named in Mondo OR it should be narrow and exactly match 1 and only 1 MIM record. I will send a message via Orphanet's site as well to try and get clarification. 

yshwetar:
Closes #7844

Submitter has provided references and correspondence that show this disease has been updated from autosomal recessive hyper-IgE syndrome to autosomal recessive hyper-IgE syndrome due to ZNF341 deficiency. Additionally added a term tracker, and OMIM link.


sabrinatoro:
@yshwetar let me know if you need help with this issue.

twhetzel:
Looking at the proxy merge terms, this looks like a case where the two terms should be merged, i.e. merge MONDO:0957426 into MONDO_0032654.

twhetzel:
The changes requested in https://github.com/monarch-initiative/mondo/issues/7844 resulted in finding two terms that should be merged. Therefore, the OMIM xref will not be added (yet) as requested in the other ticket and will be effectively handled when the terms are merged. Rather than continue in the separate [Merge ticket](https://github.com/monarch-initiative/mondo/issues/8401), Sabrina asked the merge be continued on in the original ticket.

PR #8389, opened 2024-11-14
Disorder of the tendon

### LINKED ISSUE #8042: MONDO:0100010 tendinopathy [Revise text def]
**Mondo term (ID and Label)**
MONDO:0100010 tendinopathy 

Disorders that are cause**d** by overuse of tendons. (instead of causes)

If this is the preferred definition then I recommend correcting this X-ref from exact to related: [snomedct/68172002](http://identifiers.org/snomedct/68172002) 
The preferred concept on this SNOMED CT term is "Disorder of tendon" which could cover infection, congenital issues, gout,  or injuries/disorders from over use. 
Also remove the syn of "disorder of tendon" if the intent for the mondo term is specific to "over use"-related tendon ailments.

yshwetar:
Closes #8042

Changed label tendinopathy to disorder of tendon. Changed definition to not be overuse injury only using location_top pattern. Tendinopathy is an exact synonym with my ORCID as a source. Included term tracker.


sabrinatoro (approved):
I added the Mondo pattern to the definition x-ref, and I updates the issue tracker annotation type.

Issue #8010, opened 2024-07-28
Revise subclass relationship of "hypertensive nephropathy"


nklsbckmnn:
"hypertensive nephropathy" (MONDO:0024633)

This can be acquired. I would like to suggest removing the restriction to "inherited kidney disorder" (MONDO:0100191) and put it under "kidney disorder" (MONDO:0005240) or "chronic kidney disease" (MONDO:0005300) instead.

twhetzel:
Thank you @nklsbckmnn for suggesting this change. We have reviewed this and will remove the parent class 'hereditary disease' since looking at the [OMIM page](https://omim.org/entry/608026) it is not clear that this is considered hereditary and will add the parent 'kidney disorder'. To document, [PMID:37793351 ](https://pubmed.ncbi.nlm.nih.gov/37793351/) does mention that "HN can be viewed as an "acquired error of metabolism.""

PR #8396, opened 2024-11-14
revise hierarchy for MONDO:0024633 'hypertensive nephropathy'

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8010

- removed 'hereditary disease' as parent since hereditary nature is not clear from OMIM page https://omim.org/entry/608026
- added 'kidney disorder' as parent --> I know this does become an inferred parent based on the logical definition so if it should not also be an explicit subClassOf relationship please let me know


src/ontology/mondo-edit.obo
```diff
@@ -493066,7 +493066,7 @@ xref: NCIT:C4757 {source="MONDO:equivalentTo"}
 xref: OMIM:608026 {source="MONDO:equivalentTo"}
 xref: SCTID:38481006 {source="MONDO:equivalentTo"}
 xref: UMLS:C0848548 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source="MEDGEN:167258"}
-is_a: MONDO:0003847 {source="https://orcid.org/0000-0001-5208-3432"} ! hereditary disease
+is_a: MONDO:0005240 ! kidney disorder
```
sabrinatoro, line 493069: this SubClassOf needs a source.

sabrinatoro (approved):

src/ontology/mondo-edit.obo
twhetzel: thanks for adding the source!

Issue #7588, opened 2024-04-10
congenital myopathy with excess of thin filaments


galyea123:
Hi!  This term was obsoleted in the list MONDO provided to NORD.  I did not find it currently listed at MONDO but MONDO does list nemaline myopathy 3 which is a synonym.  I noticed that other synonyms are lacking in MONDO: Congenital myopathy 2A, typical, autosomal dominant (which is the actual preferred term in OMIM).
Thank you,
Gioconda

PR #8398, opened 2024-11-14
Add synonyms and re-label term MONDO:0008070 

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/7588

- re-label term to match OMIM name
- add synonyms


sabrinatoro (approved):
I approve and will merge after QC have passed

PR #8407, opened 2024-11-15
combined oxidative phosphorylation deficiency (MONDO:0000732) definit…

### LINKED ISSUE #8216: combined oxidative phosphorylation deficiency (MONDO:0000732) definition is truncated
**Mondo term (ID and Label)**

combined oxidative phosphorylation deficiency (MONDO:0000732)

A mitochondrial oxidative phosphorylation disorder in which multiple mitochondrial respiratory chain complexes.

is not a complete sentance

katiermullen:
…ion is truncated

Closes #8216


src/ontology/mondo-edit.obo
```diff
@@ -9654,7 +9654,7 @@ replaced_by: MONDO:0019347
 [Term]
 id: MONDO:0000732
 name: combined oxidative phosphorylation deficiency
-def: "A mitochondrial oxidative phosphorylation disorder in which multiple mitochondrial respiratory chain complexes." [https://orcid.org/0000-0002-6601-2165]
+def: "A mitochondrial oxidative phosphorylation disorder in which multiple mitochondrial respiratory chain complexes are affected." [https://orcid.org/0000-0002-5002-8648, https://orcid.org/0000-0002-6601-2165]
```
sabrinatoro, line 9657: This definition doesn't make much sense to me, especially when looking at the children and how each are associated with one specific gene. 

Are multiple complexes affected in one disease? That would mean that in the children term, one gene affects multiple complexes.

sabrinatoro (commented):
@katiermullen please see my comment.

src/ontology/mondo-edit.obo
sabrinatoro: For what I can find, could it be that this disease is a multisystem disorder with variable manifestations resulting from a defect in the mitochondrial oxidative phosphorylation (OXPHOS) system? (involving any of the complexes)

sabrinatoro: See "Genetic Heterogeneity of Combined Oxidative Phosphorylation Deficiency" [here](https://omim.org/entry/609060#:~:text=Genetic%20Heterogeneity%20of%20Combined%20Oxidative%20Phosphorylation%20Deficiency)

katiermullen: Yes, this makes sense. I will update with the OMIM definition. Thank you, @sabrinatoro!

sabrinatoro (commented):
@katiermullen 
I didn't do the research, so maybe it is correct, but this definition seems to be for the parent term,  "mitochondrial oxidative phosphorylation disorder' (note, we could add this definition since it currently doesn't have one). 
What makes it a "combined ~" disease? instead of "non-combined"?

katiermullen:
@sabrinatoro It is difficult to find a definition for the grouping term. Based on the definitions for term 'combined oxidative phosphorylation defect type 8' and 'combined oxidative phosphorylation defect type 9' the original truncated definition attributed to Chris would be correct if the words, "are affected" are added, i.e. A mitochondrial oxidative phosphorylation disorder in which multiple mitochondrial respiratory chain complexes are affected. See for more evidence: https://pmc.ncbi.nlm.nih.gov/articles/PMC2563189/

The OMIM definition "A multisystem disorder with variable manifestations resulting from a defect in the mitochondrial oxidative phosphorylation (OXPHOS) system." [OMIM:609060] does seem to describe MONDO:0016387 'mitochondrial oxidative phosphorylation disorder'

Not to open a can of worms, but if we have combined oxidative phosphorylation deficiency shouldn't we also have isolated oxidative phosphorylation deficiency?

Please let me know your thoughts and I will update the PR. Thank you :)

sabrinatoro:
I completely agree! If we have "combined" we should have "isolated"
This [publication](https://pmc.ncbi.nlm.nih.gov/articles/PMC4493854/) might have answers. Looking (very briefly) at Table 2, it looks like a variation in 1 gene (most often from the nuclear DNA) can affect more than one respiratory complex. 

In summary, what you originally had makes sense (I didn't have the background to understand). The answer to my original question, Are multiple complexes affected in one disease? (even if there is only 1 gene affected), is therefore yes. 

After all this back-and-forth (sorry about that), I think that
1) the original definition (1 disease with multiple complexes affected) is correct.
2) since we are here, let's update the definition for the parent term "MONDO:0016387 'mitochondrial oxidative phosphorylation disorder'"
3) just FYI, there is a "mitochondrial disease" group who contacted us and will ask for new terms. We can ask them to review the classification.
4) we need to add "combined/isolated" to our patterns to review. 

Thank you @katiermullen, sorry for the back and forth.

sabrinatoro (approved):
That looks good. 
I am approving and merging.

PR #8414, opened 2024-11-20

twhetzel:


twhetzel:
The SPARQL query was updated and the EMC pipeline re-run creating a [new PR](https://github.com/monarch-initiative/mondo/pull/8420).

PR #8420, opened 2024-11-20

twhetzel:


src/ontology/mondo-edit.obo
```diff
@@ -331468,6 +331476,7 @@ subset: ordo_malformation_syndrome {source="Orphanet:404448"}
 subset: orphanet_rare {source="Orphanet:404448"}
 subset: otar {source="MONDO:OTAR"}
 subset: rare
+synonym: "ADNP Syndrome" EXACT [NORD:1965]
```
sabrinatoro, line 331479: @twhetzel I just want to bring your attention to this synonym addition. The only difference with the following line is the capitalization of "Syndrome" (reminder, in mondo, we do not capitalize words except for first name). 
This is a detail, and not enough to stop this PR merge, but I still want to point this out for future review of the pipeline.

sabrinatoro (approved):
This looks good and can be merged.

src/ontology/mondo-edit.obo
twhetzel: Ok, I added a ticket to track/work on this.

PR #8431, opened 2024-11-21
Fix subclass of sync

matentzn:
Previously the syncing was wrong as the method was deleting _all_ subclass of axioms, regardless of whether they were existential restrictions or subclass between names. It took a while to find this solution, but I think it should work:

1. delete _all_ subclass of axioms after
2. backing up all subclass of axioms with existentials
3. backing uo all subclass of axioms without existentials, and then removing evidence from all the mondo ingest primary sources from it.

I accidentally merged the synonym sync commit with the URL change but this will disappear since #8430 is merged.

- [ ] Depends on https://github.com/monarch-initiative/mondo-ingest/issues/708 (and subsequent dataload) to ensure no other issues are caused by this PR

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -1283,19 +1283,36 @@ update-%-mappings: $(TMPDIR)/new-exact-matches-%.owl
 		make NORM
 		mv NORM $(SRC)
 
-tmp/subclass-axioms.owl: $(SRC)
-	$(ROBOT) filter --input $(SRC) --axioms SubClassOf --preserve-structure false --trim false \
+# We want to preserve all axioms with relationships as they are here
+# As we only care about subclass relations between names.
+$(TMPDIR)/subclass-anonymous-axioms.owl: $(SRC)
+	$(ROBOT) filter --input $(SRC) \
+		--select "object-properties" \
+		--axioms SubClassOf \
+		--preserve-structure false \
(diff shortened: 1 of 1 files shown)


twhetzel:
@matentzn does this mean that https://github.com/monarch-initiative/mondo/pull/8422 should not be merged? The results of running the Synonym Sync pipeline _before_ this change?

matentzn:
@twhetzel this is such a tiny PR, I would just finalise and merge rather than doing the thinking again!

twhetzel (approved):
Works for me based on what we discussed. However, I did not run this goal locally to test.

twhetzel:
@matentzn - do you still want this ticket merged despite the data issues with the Synonym Sync posted here: https://github.com/monarch-initiative/mondo/pull/8432? In this PR it's just code, but I also do not know why you didn't merge this after it was approved so that's why I am asking.

matentzn:
I am on the fence - the diff was so large on the data load #8432 that we cannot with all confidence say that, once the bug is fixed, no other issue emerges.. I am fine merging it, but I am also fine waiting, as we are not going to run subclass sync again until the bug is fixed

twhetzel:
Ok, sounds like we should merge it then to clear out PRs and make any future updates as needed once the bug in generating the subclass sync data is sorted since as you say the subclass sync will not be run otherwise.

PR #8432, opened 2024-11-21
After refactoring the subclass of sync, a lot of changes

matentzn:
I eyeballed some of the, and they seem correct to me.. To review this:

1. Check out this file: https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/main/src/ontology/reports/sync-subClassOf.confirmed.tsv (this is what is merged into mondo-edit)
2. Confirm that the changes reflect the changes you see to the edit file
3. Confirm that the changes to the sync-subClassOf.confirmed.tsv are intended

The most important is (3), as I already did 1 and a few spotchecks re 2.

I suggest @joeflack4 you take a look here is well.


twhetzel:
I'm not sure I am looking at this in the "right" direction. For example, for MONDO:0000022 it has these changes:
```
- is_a: MONDO:0024290 {source="NCIT:C118172", source="icd11.foundation:1048673005"} ! enuresis
+ is_a: MONDO:0024290 {source="icd11.foundation:1048673005"} ! enuresis
```
So in Mondo MONDO:0000022 'nocturnal enuresis' is a subclassof MONDO:0024290 'enuresis'. The Mondo term has `xref: icd11.foundation:1048673005 {source="MONDO:equivalentTo"}` and icd11.foundation has this same subclassof relation between 'nocturnal enuresis' and 'enuresis' and is a source for subclassof relationship in Mondo. But the Mondo term also has `xref: NCIT:C118172 {source="MONDO:equivalentTo"}` and the subclassof relationship between between 'nocturnal enuresis' and 'enuresis' exists in NCIT, but _NCIT:C118172 was removed as a source_ for the subclassof relationship in Mondo.

matentzn:
> Also, many of the subclassof sources that were removed are from DOID and NCIT. When I looked in the mondo-ingest repo at tmp/component-download-ncit.owl.owl I could find 'nocturnal enuresis' and 'enuresis', but when I looked at components/ncit.owl I could not find either class 'nocturnal enuresis' and 'enuresis'

At first I found this concerning, but then I checked our specs and they clearly state that we only ever sync the _neoplasm_ branch with NCIT. enuresis is in the psychiatric disorder branch.. 

The DO is more interesting/concerning:

```
[Term]
id: MONDO:0000248
name: dengue shock syndrome
xref: DOID:0050125 {source="MONDO:equivalentTo"}
is_a: MONDO:0005502 {source="DOID:0050125", source="MONDO:Redundant"} ! dengue disease
```

```
[Term]
id: MONDO:0005502
name: dengue disease
xref: DOID:12205 {source="MONDO:equivalentTo", source="EFO:0005547"}
```

And in DO:

```
<owl:Class rdf:about="http://purl.obolibrary.org/obo/DOID_0050125">
    <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/DOID_12205"/>
    <rdfs:label xml:lang="en">dengue shock syndrome</rdfs:label>
</owl:Class>
```

The confirmed subclass table is clearly missing this, only containing these two entries related to dengue:

```
MONDO:0005358	Dengue hemorrhagic fever	MONDO:0005502	DOID:12206	DOID:12205	dengue disease
MONDO:0000259	asymptomatic dengue	MONDO:0005502	DOID:0050143	DOID:12205	dengue disease
```

In my opinion its a bug in the subclass sync, one that we didn't notice because my previous subclass pipeline didn't delete _all_ the evidence.. 

So, I think we need to:

- [ ] 1. File a bug report (if you agree it is one) (@twhetzel)
- [ ] 2. Fix that bug (@joeflack4)
- [ ] 3. Merge asap into main (maybe we should considering ditching the whole idea of a `develop` branch, its slowing us down a lot) (@twhetzel)
- [ ] 4. Make a new data release (@twhetzel)
- [ ] 5. Run the subclass pipeline #8431 again (@matentzn)
- [ ] 6. Update this PR again (@matentzn)
- [ ] 7. Review again (@twhetzel)

let me know what you think

twhetzel:
Converting to Draft until this is fixed https://github.com/monarch-initiative/mondo-ingest/issues/708

src/ontology/mondo-edit.obo
joeflack4, line 1: ## General review
I spot checked some things. Looks like I'd expect; the deletions and lack thereof.

joeflack4, line 1: ## Double checking the `is_a` `source`
Just using the example case from https://github.com/monarch-initiative/mondo-ingest/issues/708, even though that is an direct-source-indirect-mondo and not part of this PR.

|subject_mondo_id|subject_mondo_label|object_mondo_id|subject_source_id|object_source_id|object_mondo_label|
|-|-|-|-|-|-|
|ID| |SC %|>A oboInOwl:source| | |
|MONDO:0000248|dengue shock syndrome|MONDO:0005502|DOID:0050125|DOID:12205|dengue disease|

```
id: MONDO:0000248
name: dengue shock syndrome
xref: DOID:0050125 {source="MONDO:equivalentTo"}
is_a: MONDO:0005502 {source="DOID:0050125", source="MONDO:Redundant"} ! dengue disease
```

I'm pretty sure `source="DOID:0050125"` is correct but I just wanted to double check that this shouldn't be `source="DOID:12205", because in a way I feel like it makes sense to express the evidence that way as well.

joeflack4: ## `-added`?
The PR title just says "subclass sync", but your comments are only about `-confirmed`, so that's what I checked. Have we not implemented `-added` for the subclass sync, or has it just not been run for this PR?

joeflack4, line 1: ## `-added`?
The PR title just says "subclass sync", but your comments are only about `-confirmed`, so that's what I checked. Have we not implemented `-added` for the subclass sync, or has it just not been run for this PR?

joeflack4, line 1: > Confirm that the changes to the `sync-subClassOf.confirmed.tsv` are intended

@matentzn What are you asking precisely? I don't think you mean _to_ the TSV; Like I don't think you're asking me to go to `mondo-ingest` and look at the diff of that file over time.

It sounds like what you're asking is that the changes _by_ the TSV are intended.

By _intended_, it sounds like you want to take this time to double check the `mondo-ingest` subclass sync pipeline itself; to spot check a few cases where we can match up what we'd expect to see appear in the `-confirmed` TSV, and then check to see if those entries do indeed appear in the TSV. Is that what you are asking?

twhetzel:
Following up from the Tech call and analysis of the missing subclassOf source annotations from NCIT (https://github.com/monarch-initiative/mondo-curation-analysis/tree/main/mondo-subclass-sync-analysis), the decision was to add the source provenance of `MONDO:notVerified`. I also decided to keep the existing source provenance as well since multiple provenance annotations can be maintained and _if_ this information is ever curated it's easier to remove the `MONDO:notVerified` after verification of the existing subclass provenance than start from scratch to find the provenance. That decision can be easily changed by removing from the ROBOT header from the file [here](https://github.com/monarch-initiative/mondo-curation-analysis/blob/main/mondo-subclass-sync-analysis/ncit_subclass_provenance_robot_df.tsv), header column `ncitSource2`. Also, this currently static file can be moved into mondo-ingest, for time sake it was easier to pull from mondo-curation-analysis.

matentzn:
> @matentzn can you let me know if you prefer any changes to this?

Thank you @twhetzel for the NCIT stuff this is fine, and indeed the most _correct_ way; thanks for going the extra mile!

PR #8433, opened 2024-11-22
rm quick check from build

matentzn:
In order to build and test Mondo with externally managed content (which is sometimes broken), we need to be more permissive during the build. This moves the quick-check.pl  out of the normal release build pipeline into the qc (`test`) pipeline.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -1296,7 +1296,12 @@ check: mondo-edit.obo
 	 ../utils/quick-check.pl $<
 
 mondo-edit.owl: mondo-edit.obo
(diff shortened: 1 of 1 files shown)


twhetzel:
When is the test pipeline run?

twhetzel (approved):
Approving since this seems to be needed to move forward, but would like to know the plan for running `test`.

matentzn:
@twhetzel `test` is just the default test pipeline, so the test will be run as part of 

https://github.com/monarch-initiative/mondo/blob/a76a99300de25c0f747a5b77e2c0433435665e34/.github/workflows/main.yaml#L32

Issue #8437, opened 2024-11-25
Refresh imports


twhetzel:
Use the documentation [here](https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/) to refresh the imports. The source of new imported terms to add is [here](https://docs.google.com/spreadsheets/d/1UME3pTeR42hwNt1I6RPl0nvJKVfTIeJqhx80laAlD50/edit?gid=0#gid=0).

twhetzel:
Running the refresh imports, leaves 2 NCBI gene classes at the root: http://identifiers.org/ncbigene/101150212 (from Gorilla gorilla gorilla (9595)), http://identifiers.org/ncbigene/488381 (from Canis lupus familiaris (taxon ID 9615)). Each of these taxon identifiers are already in https://github.com/monarch-initiative/ncbi-gene/blob/main/src/ncbi_gene/transform.yaml.

PR #8438, opened 2024-11-25
refresh imports from 24Nov2024

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8437

Need to sort out if it's ok to merge this based on what changes need to be added from Mondo SOP and mondo-ingest pipelines. The refresh imports only takes ~20 min to run now with 40GB Memory so this can easily be re-done if needed.


matentzn (approved):
I know its in draft but looks great (waiting for this for another PR)

twhetzel:
@matentzn do you need this PR to be merged before any other pipeline PRs you're working on? If yes, then this can be merged now. I left this as draft to make sure it would not interfere with other pipeline PRs you're working on, but if this one is needed first then it's fine to merge it then.

Issue #8442, opened 2024-11-26
Test run of EMC pipeline


twhetzel:
Following updates to get the EMC pipeline code into mondo-ingest `main`, run the EMC pipeline GH Action as a test to make sure everything is still working as expected.

twhetzel:
See PR https://github.com/monarch-initiative/mondo/pull/8444

PR #8444, opened 2024-11-26

twhetzel:


twhetzel:
EMC workflow runs without qc check errors (ie the MONDO:MONDO:1234567 issue run into before). 
closes https://github.com/monarch-initiative/mondo/issues/8442

PR #8447, opened 2024-11-26

twhetzel:

This PR is the result of running the Synonym Sync pipeline and adding provenance for confirmed Synonyms between Mondo and the external source. 

The changes that should be present are the addition of source provenance for existing synonyms, so changes like:
```
- synonym: "disease" EXACT [NCIT:C2991]
+ synonym: "disease" EXACT [DOID:4, NCIT:C2991, Orphanet:377788]
```

There were also 17 errors from the [qc-related-exact-synonym errors](https://github.com/monarch-initiative/mondo/pull/8447/commits/d5722c80a92cf9fac9ea364049daba4e8d7033d4) qc check that were also fixed manually by removing the duplicated RELATED synonym.


matentzn:
@twhetzel are _all_ changes (apart from your manual stuff) confirmed to be of the kind

```
- synonym: "disease" EXACT [NCIT:C2991]
+ synonym: "disease" EXACT [DOID:4, NCIT:C2991, Orphanet:377788]
```

In this case I would be inclined to approve without too much of a review..

matentzn (approved):
This is a great step forward. 

apart from the case I mention here https://github.com/monarch-initiative/mondo/pull/8447#issuecomment-2504687453 where we have changed the synonym scope ourselves (and will now loose a lot of source evidence because of that), this is great.

Note that while this is a very important PR for presenting evidence, it fails a bit on the attribution and provenance front:

Many synonyms _come from_ some source, but where later to changed to a different scope. Its hard to model such cases in a primitive provenance system like OBO..

twhetzel:
@matentzn yes, I believe generally all cases of "confirmed" synonyms should simply be the addition of provenance to an existing synonym. However, as you point out that may be the addition of provenance to a synonym where previously there was no provenance so my example of the kinds of changes that could be seen in the PR was not complete. 

Do you think there is an issue adding provenance where there was none before? The "confirmed" cases were defined as: If the synonym string does exist in Mondo and has the same synonym scope. I do not see adding provenance where there was none before as an issue and this is also not a case of a scope difference. A scope difference was defined as  when Mondo has an exact synonym, for example, and the external source has the synonym as related (or narrow or broad).

matentzn:
> Do you think there is an issue adding provenance where there was none before? The "confirmed" cases were defined as: If the synonym string does exist in Mondo and has the same synonym scope. I do not see adding provenance where there was none before as an issue and this is also not a case of a scope difference. A scope difference was defined as when Mondo has an exact synonym, for example, and the external source has the synonym as related (or narrow or broad).

I practically agree! I am simply pointing out that 

1. If we import "X" EXACT from Source S
2. and we later decide that "X" is related
3. We lose the fact we got it from S in the first place

A problem for the future, not a blocker.

twhetzel:
@matentzn for the EDIT comment in [this post](https://github.com/monarch-initiative/mondo/pull/8447#issuecomment-2504687453) I see now the provenance is being removed in this PR since it includes only "confirmed" cases, but generally the provenance is then accounted for in "updated" cases where the synonym scope is different between Mondo and the source ontology. So with this PR only it is a bit incomplete, but will be sorted out with the other cases.

Issue #8443, opened 2024-11-26
Add qc check to make sure that no Mondo class has gene associations using both a germline and somatic properties


twhetzel:
Related to the OMIM ingest and OMIM gene pipeline, since the OMIM data does not distinguish somatic from germline gene associations using a structured format despite finding gene associations that meet our rules for being disease defining, some of these may actually be somatic, but are also added as somatic. This qc check is being added due to this.

This case did occur with MONDO:0010438 'paroxysmal nocturnal hemoglobinuria 1'

PR #8448, opened 2024-11-26
Add qc check so Mondo class does not have both somatic and germline gene associations

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8443


twhetzel:
MONDO:0009891 'acquired polycythemia vera' was caught as a qc error of this check. Based on reviewing the OMIM entry ( https://omim.org/entry/263300), both the germline and somatic gene associations were removed

OMIM entry Text:
```
A number sign (#) is used with this entry because of evidence that most cases of polycythemia vera (PV) are associated with a somatic mutation in the JAK2 gene ([147796](https://omim.org/entry/147796)) on chromosome 9p24.

Somatic mutations in the TET2 gene ([612839](https://omim.org/entry/612839)) and the NFE2 gene ([601490](https://omim.org/entry/601490)) have also been found in cases of polycythemia vera.
```

sabrinatoro (commented):
I don't know about the sparql query, but the update to the mondo-edit make sense and look good to me.

matentzn (approved):
lgtm!

PR #8488, opened 2024-12-11
Update to ODK 1.5.4

twhetzel:
ROBOT and SSSOM-Java versions are updated in ODK 1.5.4. Full https://github.com/INCATools/ontology-development-kit/releases/tag/v1.5.4 for ODK 1.5.4.

While the process for updating also typically includes a mock release to ensure no surprises when the release is run, if there is not time to update before the release in January this PR can wait until after the January release. Given the libraries updated, I do not expect issues with this update.


sabrinatoro:
I ran a mock-release on this branch, and the release failed. (note that the mock release did not fail on master)
(@twhetzel @matentzn)

Here the error log: 

runoak -i simpleobo:tmp/mondo-released.obo diff -X simpleobo:mondo-base.obo -o reports/difference_release_base.yaml --output-type yaml
runoak -i simpleobo:tmp/mondo-released.obo diff -X simpleobo:mondo-base.obo \
-o reports/difference_release_base.tsv --output-type csv --statistics --group-by-property oio:hasOBONamespace
TypeError: 'NoneType' object is not iterable
make: *** [mondo.Makefile:966: reports/difference_release_base.tsv] Error 1
rm imports/foodon_terms.txt imports/omo_terms.txt imports/ncbigene_terms.txt imports/envo_terms.txt imports/ncit_terms.txt

twhetzel:
Thanks for testing this Sabrina! Which command in the release process (https://mondo.readthedocs.io/en/latest/developer-guide/release/) resulted in the error? And when you say "note that the mock release did not fail on master", did you also create a new feature branch from the current master and include this update and run a mock release?

matentzn:
```
sh run.sh make reports/difference_release_base.tsv IMP=false MIR=false PAT=false
```

I can confirm this fails with the above error, which means there is a problem in OAK..

twhetzel:
Ok, @matentzn are you taking it from here then?

matentzn:
It will be blocked for long - I am too backed up on other issues. I am not an OAK developer, so you can just share a reproducible error on the OAK issue tracker and send a PM to Chris and ask who might be able to take care of it; unfortunately our go-to person for this task is not longer with us (Harshad :( )

twhetzel:
I ran `sh run.sh make reports/difference_release_base.tsv IMP=false MIR=false PAT=false` from this PR branch and did not get an error.

matentzn:
Maybe add a `-B`?

twhetzel:
I tried `sh run.sh make reports/difference_release_base.tsv IMP=false MIR=false PAT=false -B` and still did not get an error.
The file is created:
```
$ ls reports/difference_release_base.tsv 
-rw-r--r--  1 twhetzel  staff  296 Jan  9 07:30 reports/difference_release_base.tsv
```

twhetzel:
Can someone re-try the command and see if the error still happens?

matentzn:
here i broke it again so you can replicate the error:

https://github.com/monarch-initiative/mondo/commit/34c42ee594d1aa1de2c109ea57729c4461cc4f6f

be careful not to merge this commit to master. it passed now because the latest release did not touch the problematic parts of the diff.

matentzn:
I am not that confident for saying it but indeed, the problem does not appear anymore. I tested it and it does work.

@sabrinatoro can you when you close for the day just check out this branch and run a test release, and see if it fails? I tried, and it did not for me.. Maybe something miraculously fixed itself, but I fear that the real reason it works now is that there was something bad in the diff between the last release and the master branch when we last ran the pipeline and failed - a badness that does not seem to be there anymore.

PR #8503, opened 2024-12-19
DO NOT MERGE - Monthly confirmed subclass evidence update

twhetzel:


twhetzel:
There are still a lot of changes here, mainly removal of DOID and NCIT provenance. One issue Joe identified is that there is no confirmed subclass file for NCIT.

PR #8507, opened 2024-12-20
remove gard from obsoletion workflow

twhetzel:
PR to fix https://github.com/monarch-initiative/mondo-ingest/issues/735

The Obsoletion workflow fails because the GARD file is missing since it was removed as part of a clean-up effort in mondo-ingest to remove unused files for GARD. This PR removes `gard` from being used in the Obsoletion workflow to fix the issue.

I am looking to see if/where else removal of `gard` here effects other make goals.

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -561,7 +561,7 @@ subset-metrics:
 
 SOURCE_VERSION = $(TODAY)
 # snomed
-SOURCE_IDS = doid ncit ordo omim gard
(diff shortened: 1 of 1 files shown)


matentzn (approved):
I assume you checked that SOURCE_IDS is not used in any other pipeline where the use of gard was legitimate (I cant think of any). LGTM

twhetzel:
I did look to see if this removal of `gard` from SOURCE_IDS effects anything else and as of now my eyes are not seeing any other places where this would cause a problem.

PR #8519, opened 2025-01-03
NCIT subclass neoplasm analysis

joeflack4:
Just an analysis.
@twhetzel for review


twhetzel:
Rather than add this one-off analysis to confirm the changes for the subclass sync here, the files could go into a new folder in https://github.com/monarch-initiative/mondo-curation-analysis.

twhetzel:
I did the analysis, no need to add anything into the mondo repo.

Issue #8509, opened 2024-12-20
Issues with Obsoletion workflow not updating xref equivalent annotations correctly


twhetzel:
The Obsoletion workflow based on [this PR](https://github.com/monarch-initiative/mondo/pull/8508) is not updating the Mondo annotation on xrefs correctly. ~In some cases it was changed to `MONDO:equivalentObsolete` when the external source term was still active.~ 

Edit: In other cases, annotations of both `MONDO:equivalentTo` and `MONDO:equivalentObsolete` were found on an xref.

sabrinatoro:
@twhetzel I am not sure that there is any problem with the pipeline. 
See my comments in the PR above.

twhetzel:
Thanks Sabrina for pointing out the additional information in the Summary section on the Orphanet pages and explaining the Orphanet Summary in "Mondo-words". That handles many of the cases that led to this ticket being created. 

However, there are still some cases where the pipeline did not make the full and correct update to obsolete the xref:
1. This [original line](https://github.com/monarch-initiative/mondo/pull/8508/commits/0ffa2ed48091aceef0506a7bef4125f55b4958b5#diff-889e0a9ba62d62cc6ee7c5c9bead0075444d506253590ef0dcf4786904ac6768L158048):`xref: Orphanet:79506 {source="OMIM:143470", source="MONDO:equivalentTo"}` was changed to `xref: Orphanet:79506 {source="MONDO:equivalentObsolete", source="OMIM:143470", source="MONDO:equivalentTo"}`
2. This [original line](https://github.com/monarch-initiative/mondo/pull/8508/commits/0ffa2ed48091aceef0506a7bef4125f55b4958b5#diff-889e0a9ba62d62cc6ee7c5c9bead0075444d506253590ef0dcf4786904ac6768L158048) `xref: Orphanet:79506 {source="MONDO:equivalentObsolete", source="OMIM:143470", source="MONDO:equivalentTo"}` was changed to `xref: Orphanet:79506 {source="MONDO:equivalentObsolete", source="OMIM:143470", source="MONDO:equivalentTo"}`
[...]

PR #8508, opened 2024-12-20
Obsoletion status of xref update

twhetzel:
Update obsoletion status of xrefs.

PR #8526, opened 2025-01-04
add additional deprecation merge patterns

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8509

Added additional sed commands to the `deprecated_annotation_merging` make goal  based on the issues listed [here](https://github.com/monarch-initiative/mondo/issues/8509#issuecomment-2569825604).

I tested these changes by running: `sh run.sh make update_deprecated_mappings` and comparing the diff to the changes made in the [monthly obsoletion PR](https://github.com/monarch-initiative/mondo/pull/8508). The same changes were found in both the [PR](https://github.com/monarch-initiative/mondo/pull/8508) and from running the `deprecated_annotation_merging` make goal with the additional sed commands. The run with the updated `deprecated_annotation_merging` make goal had one additional update (xref: OMIM:210550 {source="MONDO:equivalentObsolete"} for MONDO:0008868), which does match the OMIM status of this entry since it is also deprecated.

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -1083,6 +1083,8 @@ deprecated_annotation_merging:
 	sed -i 's/source="MONDO:equivalentObsolete",\ source="MONDO:obsoleteEquivalentObsolete"/source="MONDO:obsoleteEquivalentObsolete"/g' mondo-edit.obo || true
 	sed -i 's/source="MONDO:equivalentObsolete",\ source="MONDO:equivalentTo"/source="MONDO:equivalentObsolete"/g' mondo-edit.obo || true
 	sed -i 's/\(.*\)source="MONDO:equivalentObsolete"\(.*\)source="MONDO:obsoleteEquivalentObsolete"\(.*\)/\1source="MONDO:obsoleteEquivalentObsolete"\2\3/g' mondo-edit.obo || true
+	sed -i '/source="MONDO:equivalentObsolete"/ s/, source="MONDO:equivalentTo"//g' mondo-edit.obo || true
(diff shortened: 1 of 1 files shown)


src/ontology/mondo.Makefile
```diff
@@ -1083,6 +1083,8 @@ deprecated_annotation_merging:
 	sed -i 's/source="MONDO:equivalentObsolete",\ source="MONDO:obsoleteEquivalentObsolete"/source="MONDO:obsoleteEquivalentObsolete"/g' mondo-edit.obo || true
 	sed -i 's/source="MONDO:equivalentObsolete",\ source="MONDO:equivalentTo"/source="MONDO:equivalentObsolete"/g' mondo-edit.obo || true
 	sed -i 's/\(.*\)source="MONDO:equivalentObsolete"\(.*\)source="MONDO:obsoleteEquivalentObsolete"\(.*\)/\1source="MONDO:obsoleteEquivalentObsolete"\2\3/g' mondo-edit.obo || true
+	sed -i '/source="MONDO:equivalentObsolete"/ s/, source="MONDO:equivalentTo"//g' mondo-edit.obo || true
```
matentzn, line 1086: I am not too familar with the syntax, but is the `s/` in the right place? Shouldnt it be in the beginning?

twhetzel: I had that same question, but the make goal does work with these changes as-is and when i asked chatgpt it said this line is ok with the `s/` is in the middle of the line since it's combined with a pattern address (the part before the s).

twhetzel: ok, can you approve the PR?

matentzn (approved):

Issue #8423, opened 2024-11-21
Add import refresh to release SOP?


matentzn:
We should try to avoid releasing Mondo with terms that have not been imported yet. If you look at the current release, you see at least 30 if not more classes without a classification and label (all genes or taxa):

<img width="299" alt="image" src="https://github.com/user-attachments/assets/4db6edb3-d839-49b0-8e52-f34b5ce0674a">

To solve this, I would suggest, contrary to what I might have said in the past, we run releases every Monday before a release.

@twhetzel @sabrinatoro let me know what you think!

matentzn:
EDIT: just to be clear this is _not_ a big problem, all our tooling for accessing the ontology can handle this case.

sabrinatoro:
Do you mean running the import? It is on our to-do list for the Friday before the release.
So, unless it is different from what I understand, it "should" already take care of itself.

matentzn:
I thought so, but why do we have so many genes then that are not imported?

sabrinatoro:
I would have to dig deeper into this, but maybe it has to do with all the new non-human diseases that were added after the last import. That, plus the new terms we created during the curateathon (again, I would have to check)

Question for you, @matentzn : are these going to update themselves when updating the import, or do we need to add them to a list to be added manually (as explained [here](https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/))? If the latter, it might be why there are so many.

matentzn:
they will update themselves i believe.

Import refresh should perhaps be the last thing to do.. Maybe move to Monday evening?

twhetzel:
I believe these are a combination of curate-a-thon additions, non-human, and the previous refresh imports issue that is now resolved. When I tested the updates Nico added to the refresh imports all but ~2 "updated themselves". I'll need to get these into 'https://github.com/monarch-initiative/ncbi-gene/blob/main/src/ncbi_gene/transform.yaml'.... and should create a ticket for this.

twhetzel:
The majority of these are now taken care of with the latest refresh of the imports https://github.com/monarch-initiative/mondo/pull/8540

PR #8528, opened 2025-01-05
Update documentation for refreshing imports

twhetzel:
Update documentation for refreshing the imports. Related to #8423


sabrinatoro (approved):
The instructions are clear, though I haven't tested them. 
I will approve and merge for now. We can revisit if we find that we need more details when following the instructions.

PR #8540, opened 2025-01-06
Refresh imports from 6Jan2024

twhetzel:
Refresh of imports following https://mondo.readthedocs.io/en/latest/editors-guide/import-terms-for-logical-axioms/

All imports at root except for http://identifiers.org/ncbigene/488381 were updated. I am not sure why this one is still here since it is in https://github.com/monarch-initiative/ncbi-gene/blob/main/src/ncbi_gene/transform.yaml and https://github.com/obophenotype/ncbitaxon/blob/master/subsets/taxon-subset-ids.txt. However, that "issue" can be sorted and should not block merging of this PR. 

Also, based on [past refresh import PRs](https://github.com/monarch-initiative/mondo/pull/8438) it's ok that the `sparql` file is also in the PR.


sabrinatoro (approved):
looks good. 
approving and merging

PR #8546, opened 2025-01-07
January 2025 release

sabrinatoro:


twhetzel (approved):
It generally looks ok, but one thing looks off. In the file "reports/obsoletion-candidates.tsv" it shows the 3 new obsoletion candidates, but in "src/ontology/reports/mondo_obsoletioncandidates.tsv" there is only 1 new obsoletion candidate. I thought these files represented the same content changes (ie all new obsoletion candidates should be in each file), but just with different formatting.

sabrinatoro:
@twhetzel How did you determine the new obsoletion candidates in these files (since they are both a list of terms, and not only the new one). 

I searched both files for the Mondo ID of the new obsoletion candidates, and I was able to find them in both files.

twhetzel:
I looked at the diff for both files here in GitHub (https://github.com/monarch-initiative/mondo/pull/8546/files) and looked for new additions, ie lines highlighted in green.

sabrinatoro:
I think src/ontology/reports/mondo_obsoletioncandidates.tsv is updated weekly (?) as part of the "Mondo_diff" weekly automated PR, therefore 2 obsoletion candidates were already in the file on Master (and therefore they don't show on the diff). @matentzn can you confirm?

Since all the obsoletion candidate terms are in both files, I think it is safe to merge.

twhetzel:
Ok, let me look again. I was wondering since the obsoletion candidates were generated in two rounds if something like that may have happened.

twhetzel:
I checked out the branch and confirm all new obsoletion candidates in "reports/obsoletion-candidates.tsv" are also in "src/ontology/reports/mondo_obsoletioncandidates.tsv".

PR #8551, opened 2025-01-08
Fix docs GH Action workflow error

twhetzel:
This PR removes the pip install command, since the command fails and tabulate is in ODK so `sh run.sh make ../patterns/dosdp-patterns/README.md` appears to run fine without the pip command needed. 

Update: Sabrina reported in Slack that this failed as part of creating the Mondo release and this GH Action run 
https://github.com/monarch-initiative/mondo/actions/runs/12656667270

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -75,7 +75,6 @@ pattern_ontology: ../patterns/pattern.owl
 	filter --select "<http://purl.obolibrary.org/obo/mondo/patterns*>" --select "self annotations" --signature true --trim true -o ../patterns/pattern-simple.owl
 
 ../patterns/dosdp-patterns/README.md: .FORCE
-	pip install tabulate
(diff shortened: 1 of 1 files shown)


src/ontology/mondo.Makefile
```diff
@@ -75,7 +75,6 @@ pattern_ontology: ../patterns/pattern.owl
 	filter --select "<http://purl.obolibrary.org/obo/mondo/patterns*>" --select "self annotations" --signature true --trim true -o ../patterns/pattern-simple.owl
 
 ../patterns/dosdp-patterns/README.md: .FORCE
-	pip install tabulate
```
joeflack4, line 78: ## General review
Ah, I see. Yes, this is in ODK now, as you say.

I don't see a link to an issue or a failed action, but I assume that, as you are saying, it failed due to this `pip install`? I wonder why.

joeflack4 (approved):

PR #8463, opened 2024-11-29
NEVER MERGE: Test PR for adding synonyms

matentzn:
@twhetzel this is for us to see how "adding synonyms" would look like. 

If you are happy with the _enormous_ result (this is a huge impact change potentially), I will create a new PR.

## READ THIS

- [ ] Do not merge this PR. Create a new one with the correct pipeline step, the ones in hear are simplified by omitting confirmed subclasses so the review is easier.

PR #8562, opened 2025-01-12
NEVER MERGE: Test PR for adding synonyms - `qc-related-exact-synonym`

joeflack4:
Confirms bug fix for `qc-related-exact-synonym` failures in the following PR is working:
- https://github.com/monarch-initiative/mondo-ingest/pull/747

Context: 
- #8463 which is part of the Super Ticket https://github.com/monarch-initiative/mondo-ingest/issues/754

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -1288,11 +1288,12 @@ all: config/exclusion_reasons.tsv
 $(TMPDIR)/subclass-confirmed.robot.tsv:
 	wget "https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/main/src/ontology/reports/sync-subClassOf.confirmed.tsv" -O $@
 
+# TODO tmp: check bug fix branch and see if addresses any QC failures for https://github.com/monarch-initiative/mondo/pull/8463
 $(TMPDIR)/synonyms-confirmed.robot.tsv:
-	wget "https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/refs/heads/main/src/ontology/reports/sync-synonym/sync-synonyms.confirmed.robot.tsv" -O $@
+	wget "https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/refs/heads/bugfix-syn-sync-exclusions--mini-build/src/ontology/reports/sync-synonym/sync-synonyms.confirmed.robot.tsv" -O $@
 
 $(TMPDIR)/synonyms-added.robot.tsv:
-	wget "https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/refs/heads/main/src/ontology/reports/sync-synonym/sync-synonyms.added.robot.tsv" -O $@
+	wget "https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/refs/heads/bugfix-syn-sync-exclusions--mini-build/src/ontology/reports/sync-synonym/sync-synonyms.added.robot.tsv" -O $@
 
 $(TMPDIR)/new-exact-matches-%.tsv:
 	wget "https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/main/src/ontology/lexmatch/unmapped_$*_lex_exact.tsv" -O $@


joeflack4 (commented):
https://github.com/monarch-initiative/mondo/actions/runs/12729914779

src/ontology/mondo.Makefile
joeflack4, line 1: ## Results
Look at: [CI run](https://github.com/monarch-initiative/mondo/actions/runs/12730217064)

I would hope/expect that the results from that will be the same as what I have now locally:
```
FAIL Rule ../sparql/qc/general/qc-related-exact-synonym.sparql: 7 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0007039,oboInOwl:hasExactSynonym,ACN
http://purl.obolibrary.org/obo/MONDO_0007039,oboInOwl:hasExactSynonym,BANF
http://purl.obolibrary.org/obo/MONDO_0010753,oboInOwl:hasExactSynonym,EDS5
http://purl.obolibrary.org/obo/MONDO_0011811,oboInOwl:hasExactSynonym,SCA24
http://purl.obolibrary.org/obo/MONDO_0013426,oboInOwl:hasExactSynonym,LDS1C
http://purl.obolibrary.org/obo/MONDO_0014167,oboInOwl:hasExactSynonym,FCMTE5
http://purl.obolibrary.org/obo/MONDO_0033005,oboInOwl:hasExactSynonym,SCAR5
```

Basically, https://github.com/monarch-initiative/mondo-ingest/pull/747 addresses all but these 7.

These 7 are cases in which we have `skos:exactMatch` to terms from 2 different sources, and the scope from those 2 sources are different.

**Example: `MONDO:0007039` "ACN"**
`mondo-edit.obo`:
```
synonym: "ACN" EXACT ABBREVIATION [DOID:0111252]
synonym: "ACN" RELATED ABBREVIATION [OMIM:101000]
```

`components/doid.owl`:
```
    <owl:Class rdf:about="http://purl.obolibrary.org/obo/DOID_0111252">
        <oboInOwl:hasExactSynonym xml:lang="en">ACN</oboInOwl:hasExactSynonym>
    </owl:Class>
```

`components/omim.owl`:
```
    <owl:Class rdf:about="https://omim.org/entry/101000">
        <oboInOwl:hasRelatedSynonym>ACN</oboInOwl:hasRelatedSynonym>
    </owl:Class>
```

Proposed solution for these remaining:
a. Some kind of manual curation before running pipeline in `mondo`
b. Update synonym sync: Filter out any cases where sources disagree

I would recommend (b). And instead of only filtering them out, I think we should generate another output that shows all of these conflicts, so that we have the opportunity to curate.

joeflack4 (commented):
See comment thread.

twhetzel (approved):
The changes in this PR comparing the previous Synonym Sync code to the update in https://github.com/monarch-initiative/mondo-ingest/pull/747 generally looks good. For the "added" case, the `EXCLUDE` annotation is now being taken into account properly by the [code](https://github.com/monarch-initiative/mondo-ingest/pull/747/files) and not adding new synonyms where the existing synonym has the same string value, but has the `EXCLUDE` annotation.

src/ontology/mondo.Makefile
twhetzel: @joeflack4 
(1) Since these are the results from your latest updates to the Synonym Sync why would another output be needed that shows _all_ of these conflicts, in other words why would this list of these `qc-related-exact-synonym.sparql` qc check errors not be complete?

(2) In all of these 7 cases the issue is not between any generic combination of sources, it is between DOID and OMIM and specifically an OMIM abbreviation from an alternative symbol marked as formerly. In all of these cases the scope of the OMIM symbol marked as formerly should take precedence and the DOID synonym abbreviation should be filtered from the synonym sync output file. I will create a new ticket for this.

joeflack4: (1) My recommendation is for ease of running things. If we filter out problematic cases before they hit the QC, then this allows us to seamlessly run the synonym sync monthly without the hold up of handling the QC failures. Having `mondo-ingest` put them in a special curation output instead will allow us to handle those cases at any time. So the idea is that we can curate at a time of choosing, rather than the stress/rush of needing to do it when we're trying to get a monthly build update done on time.

PR #8598, opened 2025-01-18
update equivalent/other hierarchy

sabrinatoro:
Some x-ref had both "equivalentTo" and "otherhierarchy" (probably from another PR). Resolving that in this


twhetzel (approved):
Hmm, I'm not what's been run lately that caused this, but I think I have seen this in the past.

PR #8616, opened 2025-01-20

twhetzel:


twhetzel (commented):
I spot checked a few updates against the source data files and it looks good.

src/ontology/mondo-edit.obo
```diff
@@ -368109,6 +368081,14 @@ property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/
 [Term]
 id: MONDO:0016025
 name: obsolete myoclonic-astatic epilepsy
+subset: gard_rare {source="GARD:2169", source="MONDO:GARD"}
+subset: nord_rare {source="MONDO:NORD"}
+subset: ordo_disorder {source="Orphanet:1942"}
+subset: otar {source="MONDO:OTAR"}
```
matentzn, line 368087: Not related to this PR: it is very important @twhetzel you also have an eye on our commitment to our community to protect the terms they are using, because any time we obsolete a `otar`, `nord`, `clingen`, `gard`, etc term, we encur bigger costs on the side of these users!

matentzn (approved):
Looks great! Lots of changes from Medgen, and inferred rare (from the epilepsy restructuring), happy here!

PR #8639, opened 2025-01-31
updates based on OMIM information

sabrinatoro:
1) updated the definition of MONDO:0007171 based on [OMIM:108770](https://omim.org/entry/108770) "A number sign (#) is used with this entry because of evidence that atrial standstill-1 (ATRST1) is caused by coinheritance of a variant in the SCN5A gene (600163) in combination with a rare connexin-40 (GJA5; 121013) genotype."

2) Updated the Gene-association for MONDO:0010218 based on [OMIM:278850](https://omim.org/entry/278850) "evidence that familial 46,XX sex reversal is caused by heterozygous duplication or triplication of a 68-kb regulatory region (XXSR) -584 to -516 kb upstream of the SOX9 gene (608160) on chromosome 17q24."


twhetzel (approved):
Great updates!

PR #8645, opened 2025-02-03
FOXG1 syndrome/Disorder; Rett syndrome, congenital variant

### LINKED ISSUE #7850: FOXG1 syndrome/Disorder; Rett syndrome, congenital variant [Merge]
**Mondo term (ID and Label)**
MONDO:0035383 FOXG1 syndrome

**Reason for deprecation**
GeneReviews uses FOXG1 syndrome as their preferred name for this chapter: https://www.ncbi.nlm.nih.gov/books/NBK604176/  that points to OMIM [613454](https://www.ncbi.nlm.nih.gov/omim/613454)	RETT SYNDROME, CONGENITAL VARIANT (See Table B for OMIM entries)

But I see there is also FOXG1 disorder (MONDO:0100040) that is parent to Rhett syndrome, congenital variant (MONDO:0013270), so I don't know if that is a conceptual match to FOXG1 syndrome? 

If UMLS is correct, then Orphanet 561854 and OMIM 613454 are the same concept on CUI C3150705. 

**Term to be merged with**
???
Not sure, but I would seem that there it at least one too many Mondo concepts here among this trio of entries. Or maybe it's not a merge but a restructuring of FOXG1 syndrome under the Disorder term?

katiermullen:
Closes #7850


sabrinatoro (approved):
@katiermullen I added an x-ref for the definition. I am approving and merging after QC checks have passed. 

Let's please follow-up about the name and the definition for the disease (FOXG1 disorder is not great, we should check if we can make it "neurological disease" specific)

PR #8656, opened 2025-02-05
Fix pattern_mkdocs goal 

twhetzel:
The error occurs when trying to pip install tabulate. Since ODK includes tabulate, I removed this line from the pattern_mkdocs goal and ran the goal locally to get the updated files.


twhetzel:
@matentzn can you review this?

matentzn (approved):

twhetzel:
@sabrinatoro this should fix the GH Action errors you've been seeing. If any of these happen again in the future, please tag me on Slack since I do not receive the error emails.

Issue #8730, opened 2025-02-17
Fix broken link and list order for ClinGen doc page


twhetzel:
The link to this page from the menu is not working https://mondo.readthedocs.io/en/latest/ClinGen-lumping-splitting.md

PR #8731, opened 2025-02-17
fix link and re-order item in menu for new ClinGen doc page

twhetzel:
closes [8730](https://github.com/monarch-initiative/mondo/issues/8730)


sabrinatoro (approved):
Approving

Issue #8736, opened 2025-02-18
Add OMIM SubclassOf axioms


twhetzel:
Use the `src/ontology/reports/omim.subclass.added.robot.tsv` ROBOT file [here](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/reports/omim.subclass.added.robot.tsv) and update Mondo.

Merge ROBOT template docs https://mondo.readthedocs.io/en/latest/editors-guide/robot-template/#merging-a-robot-template-into-mondo_1

**UPDATE**

In the ROBOT template, in row 2 the text `AI obo:mondo#excluded_subClassOf` should be changed to `SC %`. The updated ROBOT file, [omim.subclass.added.robot - MODIFIED](https://docs.google.com/spreadsheets/d/1FaOD7JHrZ-c4FOLuNADaZGoD7T96QIjx_2wwLAvW3Uw/edit?usp=sharing), is in the gDrive in Mondo-->Templates.

nicolevasilevsky:
@twhetzel do you need my help with this?

twhetzel:
Thanks @nicolevasilevsky, while I am still assigned this issue Nico has taken over this due to the QC issues encountered in this PR https://github.com/monarch-initiative/mondo/pull/8741 when I merged the ROBOT template. 

I think this is now the latest PR related to this issue https://github.com/monarch-initiative/mondo/pull/8896. I do not know Nico's overall plan here so best to check-in with him about what is needed.

matentzn:
Are excluded subclass of axioms no longer included in the omim robot template?

PR #8738, opened 2025-02-18
run merge template against omim.subclass.added.robot.tsv template

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8736

following slack chat[1], it was suggested to run the merge template workflow against [omim.subclass.added.robot.tsv](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/reports/omim.subclass.added.robot.tsv)

[1] When we import a new OMIMPS (ie add an omimps equivalent in Mondo), do we have a workflow to update the classification in Mondo such that all the omim that are part of this phenotypic series have their mondo equivalent subclass of the omimps equivalent?


matentzn (changes requested):
Oops I forgot the template uses "excluded subclass of" and not subclass of. 

Please update the template locally to using subclass of rather than excluded subclass of and update..

twhetzel:
ok, i'll close this PR and do that --> update the template locally to use "subclass of" rather than "excluded subclass of" and update

PR #8742, opened 2025-02-18

twhetzel:


src/ontology/mondo-edit.obo
```diff
@@ -509734,10 +509757,13 @@ id: MONDO:0032829
 name: neurodevelopmental disorder with hypotonia and variable intellectual and behavioral abnormalities
 def: "Neurodevelopmental disorder in which the cause of the disease is a variation in the POLR2A gene; it is characterized by early-onset hypotonia, delayed walking, poor speech, and impaired intellectual development. Other features may include feeding difficulties, dysmorphic features, and visual problems. Brain magnetic resonance imaging tends to show delayed myelination, thin corpus callosum, and/or enlarged ventricles." [https://orcid.org/0000-0001-9310-0163, PMID:31353023, PMID:35328024]
 subset: clingen {source="MONDO:CLINGEN"}
+subset: nord_rare {source="MONDO:NORD"}
+subset: rare
 synonym: "NEDHIB" EXACT ABBREVIATION [OMIM:618603]
 synonym: "neurodevelopmental disorder with hypotonia and variable intellectual and behavioral abnormalities" EXACT CLINGEN_LABEL [OMIM:618603]
 synonym: "POLR2A-associated neurodevelopmental disability" EXACT [https://orcid.org/0000-0001-9310-0163, PMID:31353023, PMID:35328024]
 synonym: "POLR2A-related disorder" EXACT [https://orcid.org/0000-0001-9310-0163, PMID:31353023]
+synonym: "POLR2A-Related Disorders" EXACT []
```
sabrinatoro, line 509766: This synonym is added without an x-ref. Where does this come from?

src/ontology/mondo-edit.obo
```diff
@@ -513271,6 +513299,7 @@ subset: rare
 synonym: "IDDSELD" RELATED ABBREVIATION []
 synonym: "INTELLECTUAL DEVELOPMENTAL DISORDER WITH SEIZURES AND LANGUAGE DELAY" RELATED []
 synonym: "SETD1B-NDD" EXACT ABBREVIATION [https://orcid.org/0000-0001-9310-0163]
+synonym: "SETD1B-Related Neurodevelopmental Disorder" EXACT []
```
sabrinatoro, line 513302: This synonym was added without an x-ref. Also, it is the same as the synonym below, but with different capitalization. We should ignore capitalization except for Gene names.

src/ontology/mondo-edit.obo
```diff
@@ -533162,6 +533196,7 @@ synonym: "intellectual developmental disorder, autosomal dominant 57" EXACT [OMI
 synonym: "intellectual disability, autosomal dominant 57" EXACT []
 synonym: "mental retardation, autosomal dominant 57" RELATED DEPRECATED []
 synonym: "MRD57" RELATED DEPRECATED []
+synonym: "TLK2-Related Neurodevelopmental Disorder" EXACT []
```
sabrinatoro, line 533199: added without x-ref; same synonym as the one below but with different capitalization

sabrinatoro (approved):
@twhetzel 
I have found a few things (see comments). 
I think these are not crucial and should not block this PR. Therefore I am approving, but I will let you merge after you have seen the comments. 

Summary of the issues: 
- some synonyms were added without x-ref
- some synonyms (same as the one added without x-ref) were the same as already existing synonym, but only differ by capitalization
- we are adding subset annotation to obsoleted class (most or all of them were merged to another term). The solution is to set up a report to share with the external sources as this need to be resolved at the source.

PR #8741, opened 2025-02-18
omim added subclassOf

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8736

see ticket for details

PR #8746, opened 2025-02-19
Add goal to create a branch diff and see unsats

matentzn:
reports/mondo_unsats.md is not used anywhere so its safe to change it; here we add a goal to generate a random sample of unsats, plus a markdown diff using robot for branch review.

Context: #8741

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -640,9 +640,17 @@ tmp/mondo-lastbase.owl:
 reports/mondo_diff.md: mondo-base.owl tmp/mondo-lastbase.owl
 	$(ROBOT) diff --left tmp/mondo-lastbase.owl --right $< -f markdown -o $@
 
-reports/mondo_unsats.md: mondo.obo
-	$(ROBOT) explain -i $< --reasoner ELK -M unsatisfiability --unsatisfiable all --explanation $@ \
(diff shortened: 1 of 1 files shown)


src/ontology/mondo.Makefile
```diff
@@ -640,9 +640,18 @@ tmp/mondo-lastbase.owl:
...
+tmp/mondo-main-edit.owl:
+	wget "https://raw.githubusercontent.com/monarch-initiative/mondo/refs/heads/master/src/ontology/mondo-edit.obo" -O $@
+
+reports/mondo_branch_edit_diff.md: mondo-edit.obo tmp/mondo-main-edit.owl
+	$(ROBOT) diff --left tmp/mondo-main-edit.owl --right $< -f markdown -o $@
+
+reports/mondo_unsats.md: mondo-edit.obo
+	$(ROBOT) explain -i $< -M unsatisfiability --unsatisfiable random:10 --explanation $@
+
+.PHONY: mondo_branch_diff reports/mondo_branch_edit_diff.md reports/mondo_unsats.md tmp/mondo-main-edit.owl
```
twhetzel, line 652: @matentzn I added all files as `.PHONY` so the `mondo_branch_diff` would always run with new files (aka effect of using `-B`). Any objections?

twhetzel (approved):
lgtm

PR #8783, opened 2025-02-25
Add AARS1-related leukoencephalopathy

### LINKED ISSUE #8482: Add AARS1-related leukoencephalopathy
Please create a new term: “AARS1-related leukoencephalopathy” 

Description: Any leukoencephalopathy in which the cause of the disease is a variant in the AARS1 gene

This new term should be a parent term to the following 3 Mondo Terms: MONDO:0014593 and MONDO:0030517 and MONDO:0030634

Thank you1

katiermullen:
closes #8482

-created new term
-added children terms (existing Mondo terms)
-added gene ID for MONDO:0030634 because it was missing


nicolevasilevsky (approved):
Looks good.

I wonder if we need a disease term for leukoencephalopathy? It is a phenotype in HPO, so maybe not?

Issue #8816, opened 2025-03-06
Smushed IRI's with |


ThomasThelen:
I found a blank node, _:b287734 that has the following predicate

oboInOwl:hasDbXref "DOID:0070508|PMID:34722187"

Two IRI's smushed together separated with a bar (pretty sure we want to use semantics to describe whatever function | is playing). It also makes it a pain to query (have to know to split on |)

matentzn:
Thank you @ThomasThelen for taking the time to writing this issue, much appreciated! I have fixed it in #8831 and added a QC check to prevent it from happening again in the future #8830.

sabrinatoro:
This issue has been fixed, and the changes will be available in the next Mondo release. 
Thank you @ThomasThelen for reporting this issue.

PR #8830, opened 2025-03-10
Add QC check to prevent pipes in annotations

matentzn:
Prevents #8816 from happening again, probably a rogue ROBOT template without a SPLIT declaration

Errors are fixed in https://github.com/monarch-initiative/mondo/pull/8831


sabrinatoro (approved):
Thank you @matentzn !!

Issue #8821, opened 2025-03-08
Dump Mondo terms


twhetzel:
Per Julie's request, get this information from the latest release of Mondo:
- term ID
- label
- definition
- comment
- parents
- synonyms
- OMIM equivalent
- orphanet equivalent

PR #8827, opened 2025-03-09
Add goal to create a dump of Mondo terms

twhetzel:
Closes https://github.com/monarch-initiative/mondo/issues/8821

PR #8812, opened 2025-03-05
Add goals to create general stats and synonym stats and rare and community stats

twhetzel:
closes #8809
closes #8906 
closes #8916 
closes #8908

Create general stats as requested in this [gSheet](https://docs.google.com/spreadsheets/d/1TGD1ta2RyoLMvAne1dp-b3LTj6b5Y_jTvsz04HPHmEk/edit?gid=1264043076#gid=1264043076)
The queries can either be run individually or as: 
`sh run.sh make create-general-mondo-stats-all`
- [x] curator review and approval of general statistics values
- [ ] developer review of general statistics queries and make goals

---

Update: added make goal and query to create synonym stats. The query can either be run as a "report" or as
`sh run.sh make create-synonym-mondo-stats-all`
- [x] curator review and approval of synonym statistics values
- [ ] developer review of  synonym statistics queries and make goals

---

Update 27Mar2025: added make goal and query to create rare subset stats. The query can either be run as a "report" or as `sh run.sh make create-rare-mondo-stats-all`
- [x] curator review and approval of rare statistics values
- [ ] developer review of rare statistics queries and make goals

---
Update 29Mar2025: added make goal to create report file of opened and closed issues between two dates with the count within each open/closed issue by tag. The dates can be entered manually or by default it uses the last two most recent dates of the tagged releases compared to the date the goal is run on. It also creates a file of unique users, the count of issues they opened or closed, and unique tags for these issues is also created. 
This can be run from the command line given some prerequisites since it does not use ODK or from the GitHub Action "Generate GitHub Issue Statistics"
- [ ] curator review
Example:
```
# Stats generated on 2025-03-30 06:02:16Z for GitHub issues from 2025-03-15 to 2025-03-18
type	label	count
new_issue		1
closed_issue		12
label_new	New term request	1
label_new	user request	1
label_closed	New term request	2
...
```

and 

```
# Stats generated on 2025-03-30 04:58:22Z for GitHub issues from 2025-03-15 to 2025-03-18
type	github_handle	count	labels
opened	J-Siew	1	New term request,user request
closed	galyea123	3	effort-XS,relabel term,synonym,user request
closed	sagehrke	1	outreach
closed	eedoh01	1	New term request,user request
...
```

PR #8833, opened 2025-03-10
Add make goal to generate data file for curation tool

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8821 

change from IRI to CURIE and use `mondo.owl` based on Nico's comment in [another PR](https://github.com/monarch-initiative/mondo/pull/8812#discussion_r1987645824) and also updated the docs (update: this is an update from the earlier merged PR https://github.com/monarch-initiative/mondo/pull/8827)


twhetzel:
Leaving as Draft until Julie confirms the sample file I sent can be loaded into Airtable

sabrinatoro:
I cannot review this PR: the sparql query is not ".sparql", and therefore I cannot run it as I would run a similar report. 

@matentzn could you please check that the changes in mondo.Makefile are ok?

@twhetzel could you please run the query and attach the result, so a curator can review that it is as expected? In addition, we should also share the resulting file with Julie to make sure it is what she needs.

twhetzel:
@sabrinatoro I updated the query to have a ".sparql" extension, however the query is intended to be run with the specific `make` goal as documented `sh run.sh make dump-mondo-terms`.

sabrinatoro (approved):
This looks good to me. 
I am approving, however, I don't know whether the changes in the mondo.Makefile are correct (@matentzn do you want to take a look? ). Therefore, I am not merging this PR. 

Note that if the change is trivial, it is ok to merge.

PR #8835, opened 2025-03-11
DO NOT MERGE: Evaluating synonym sync QC checks (1 of 2)

joeflack4:
Trying to address:
- https://github.com/monarch-initiative/mondo-ingest/issues/801

Related:
- #8838 

All QC from above issue passed in this PR using `mondo-ingest` branch `_main_data-build-10Mar2025`: QC results ([1](https://github.com/monarch-initiative/mondo/actions/runs/13825101000/job/38678481593?pr=8835), [2](https://github.com/monarch-initiative/mondo/actions/runs/13825707345/job/38680119661?pr=8835))

PR #8838, opened 2025-03-13
DO NOT MERGE: Evaluating synonym sync QC checks (2 of 2)

joeflack4:
Trying to address:
- https://github.com/monarch-initiative/mondo-ingest/issues/801

Related:
- #8835

> [!note]
> - Includes: #8463


src/ontology/mondo-edit.obo
joeflack4, line 1: ## Results
Running in the same way as #8835.

That PR ran off the `-confirmed` only and everything passed. This one runs `-confirmed` and `-added` and has these failures.

Both PRs are using `mondo-ingest` branch `_main_data-build-10Mar2025` (https://github.com/monarch-initiative/mondo-ingest/pull/808)

There are merge conflicts. Trying to fix that so I can run QC in the PR.

Unfortunately all the QC are failing on this branch, unlike with #8835:

```
FAIL Rule ../sparql/qc/general/qc-duplicate-exact-synonym-no-abbrev.sparql: 2 violation(s)
entity,property,value
MONDO:0016002-MONDO:0020066,oboInOwl:hasExactSynonym-oboInOwl:hasExactSynonym,Cutis hyperelastica
MONDO:0020066-MONDO:0016002,oboInOwl:hasExactSynonym-oboInOwl:hasExactSynonym,Cutis hyperelastica
```

```
FAIL Rule ../sparql/qc/mondo/qc-exact-synonyms-non-exact-mappings.sparql: 1 violation(s)
entity,label,xref,synonym,code
http://purl.obolibrary.org/obo/MONDO_0005645,ancylostomiasis,NCIT:C35805,Ancylostomiasis,MONDO:directSiblingOf
```

```
FAIL Rule ../sparql/qc/general/qc-syn-equal-label.sparql: 1 violation(s)
term,label,related,synonym
http://purl.obolibrary.org/obo/MONDO_0032569,"isolated growth hormone deficiency, type 5","isolated growth hormone deficiency, type 5",http://www.geneontology.org/formats/oboInOwl#hasRelatedSynonym
```

I will have to investigate further.

matentzn: I think the qc-exact-synonyms-non-exact-mappings.sparql should probably not be fixed pre-emptively. In my view this is a useful indicator of a potentially wrong mapping, not a wrong synonym

twhetzel: Sounds good. Overall, I think these few types of QC errors are within the limits of a curator fixing these during the monthly run of the Synonym Sync pipeline. My interest in checking on these additional QC errors when "-added" synonyms are added is to make sure these qc errors are not pointing out any items that are a gap in the logic of the synonym sync pipeline.

The main item I had questions about is `qc-duplicate-exact-synonym-no-abbrev.sparql` since there was a fix for this previously so it would be good to know if something else is needed. In looking at those Mondo terms, I see `MONDO:0020066` has the synonym "Cutis hyperelastica" with a synonym_type of http://purl.obolibrary.org/obo/mondo#EXCLUDE "Synonym to be removed from public release but maintained in edit version as record of external usage".

joeflack4: @twhetzel You saved me some time! Yes, `MONDO:EXCLUDE` is the issue here.

 "Cutis hyperelastica" appears on `MONDO:0020066` in `mondo-edit.obo` but not `mondo.owl`. The synonym sync queries `mondo.owl`. This is why it is not detecting it.

I think that this one is easier to leave to curation rather than try and build a solution into the synonymy sync pipeline at this point.

FYI, this synonym also appears in `mondo-ingest` `mondo-exclusion.configs.yml`, though I don't think this matters in this case. If the synonym appeared in `mondo.owl`, I believe that the synonym sync would catch this collision and prevent another "Cutis hyperelastica" from being added.

@twhetzel Based on your and Nico's comments, perhaps it is best for me to call this a conclusion to this issue, then?
- https://github.com/monarch-initiative/mondo-ingest/issues/801

I will move onto other issues. But let me know if there's anything more I should do on this.

Issue #8515, opened 2024-12-31
[Merge] MONDO:0005259 - Asperger syndrome into appropriate term related to Autism Spectrum Disorder


DnlRKorn:
**Mondo term (ID and Label)**
MONDO:0005259 - Asperger syndrome

**Reason for deprecation**

"Note: Individuals with a well-established DSM-IV diagnosis of autistic disorder, Asperger’s disorder, or pervasive developmental disorder not otherwise specified should be given the diagnosis of autism spectrum disorder. Individuals who have marked deficits in social communication, but whose symptoms do not otherwise meet criteria for autism spectrum disorder, should be evaluated for social (pragmatic) communication disorder." - https://www.autismspeaks.org/autism-diagnostic-criteria-dsm-5

"Today, those who would have previously met the criteria for Asperger syndrome are diagnosed with level 1 ASD. There is no difference between Asperger's and autism level 1 diagnoses." - https://www.autismspeaks.org/asperger-syndrome

**Term to be merged with**
I feel we have two options here; either 
A) Merge with MONDO:0005258 - autism spectrum disorder; or
B) Following the DSM-5 (https://www.autismspeaks.org/levels-of-autism) create three children of MONDO:0005258/autism spectrum disorder: ASD Level 1, ASD Level 2, ASD Level 3. Merge MONDO:0005259 with the created term for ASD Level 1.

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-1780-9872

MeeSiing:
@DnlRKorn The merge is complete. The changes will be reflected in the upcoming release. Thanks again for submitting the issue.

PR #8849, opened 2025-03-15
add new autism terms

nicolevasilevsky:
- 'autism spectrum disorder 1'
- 'autism spectrum disorder 2'
'autism spectrum disorder 3'
- add obs tag to 'Asperger syndrome'
address https://github.com/monarch-initiative/mondo/issues/8515

@DnlRKorn do you want to check this PR? is it okay to use the definitions from the website?

address #8515


DnlRKorn:
Looks good to me!

sabrinatoro (changes requested):
I think we need to be more specific with the definition. The first sentence is: "a type of autism spectrum disorder that required support/substantial/very substantial support". 
If one requires "very substantial support", they also require "support". 
Therefore we need to be more specific. 

Could we review the definitions such that the difference between the terms is clearly at the beginning of the definition. (e.g. ~ characterized by XXX, where XXX is clearly different between the different types)?
(Note, I would need to do more reading to be more helpful. If necessary, we can reach out to experts)

nicolevasilevsky:
@DnlRKorn could you suggest more specific definitions? Thank you!

nicolevasilevsky:
@sabrinatoro I updated the definitions and merged Asperger syndrome into 'autism spectrum disorder 1'

nicolevasilevsky:
I need to redo this - add the new terms and obs tags but it is premature to do the obsoletion

nicolevasilevsky:
I redid this on a new PR: https://github.com/monarch-initiative/mondo/pull/9093

PR #8854, opened 2025-03-16
reclassify 'anemia, nonspherocytic hemolytic, due to G6PD deficiency'

### LINKED ISSUE #8665: Embed in the same ontology tree: 0010480 and 0005775
The ClinGen G6PD VCEP requests that the CNSHA disease entity (MONDO: 0010480) and the G6PD deficiency disease entity (MONDO: 0005775) be embedded within the same ontology tree.

We are not sure if this qualifies as a "revise subclass" request or other. Our apologies in advance if this is not the correct label for this request.

nicolevasilevsky:
as a child of 'G6PD deficiency'
close #8665


sabrinatoro (approved):
The changes look good. 
I have issues with the name "G6PD deficiency", and the fact that some SubClass don't have sources. However this is out of scope for this issue. 
Approving and merging

PR #8856, opened 2025-03-16
clean up synonyms on IBD1

### LINKED ISSUE #8226: [Revise subclass] inflammatory bowel disease 1 MONDO_0009960 too broad?
inflammatory bowel disease 1 MONDO_0009960

It seems inflammatory bowel disease 1 is defined as IBD caused by NOD2 mutations but is synonymous with all IBD terms including ulcerative colitis. To the best of my knowledge there is no evidence that NOD2 causes ulcerative colitis or other forms of IBD, only Crohn's, so it seems inappropriate to treat IBD1 as synonymous with all IBD terms. Related, I am also a little confused by IBD30 and IBD5: IBD30 has no description and IBD5 is connected to 5q31 but there are many monogenic IBD genes/regions so it's unclear why this region has its own term**

0000-0003-1622-0502

nicolevasilevsky:
and add def to IBD30

close #8226


sabrinatoro (approved):

Issue #8627, opened 2025-01-25
Revise toxic diffuse goiter


nklsbckmnn:
toxic diffuse goiter (0001104) is currently not asserted nor inferable as a descendant of goiter (0005397)

nicolevasilevsky:
@nklsbckmnn do you have an ORCID that you could share? Thanks!

nicolevasilevsky:
thank you!

PR #8869, opened 2025-03-18
add source to superclass for 'toxic diffuse goiter'

nicolevasilevsky:
address #8627


sabrinatoro (approved):
approving and merging

PR #8870, opened 2025-03-18

twhetzel:


twhetzel:
The majority of deletions in the PR are from changes in NORD content (which NORD confirmed) as seen in the mondo-ingest [data PR](https://github.com/monarch-initiative/mondo-ingest/pull/808/files#diff-25f309059e697c84d27c86dba119764cd01714a63fd8383a8e674e59a9da348c), which is the source of the data applied here in this PR.

Initially, there was a QC error:
```
FAIL Rule ../sparql/qc/mondo/qc-ordo-subset-exact-mapping.sparql: 1 violation(s)
entity,property,value
http://purl.obolibrary.org/obo/MONDO_0700282,http://purl.obolibrary.org/obo/mondo#ordo_disorder,Orphanet:289494
```

The mondo-ingest data build was run on Mar. 10 and there was a Mondo PR that was merged into `mondo-edit.obo` on Mar. 14 that effected which Mondo term contained the Orphanet:289494 xref and as a result the placement of which Mondo term the `ordo_disorder` subset annotation was added to leading to the QC error. 

To expand, since the version of `mondo-edit.obo` used to create the subset annotation had changed between when it was created and when the subset annotation was added, the term `MONDO:0700282` no longer has the mapping to Orphanet:289494 (as seen [here](https://github.com/monarch-initiative/mondo/pull/8800/files#diff-889e0a9ba62d62cc6ee7c5c9bead0075444d506253590ef0dcf4786904ac6768L548727)) and the Orphanet:289494 xref is now on MONDO:0100605 (as seen [here](https://github.com/monarch-initiative/mondo/pull/8800/files#diff-889e0a9ba62d62cc6ee7c5c9bead0075444d506253590ef0dcf4786904ac6768R544901)) and therefore MONDO:0100605 should have the `ordo_disorder` subset annotation.

Issue #8464, opened 2024-12-02
Modify OMIM gene pipeline so gene associations are not added to obsolete terms


twhetzel:
In the first run of the OMIM gene pipeline, gene associations were added to obsolete Mondo terms, but this causes caused [qc check errors ](https://github.com/monarch-initiative/mondo/pull/8108/commits/99fea11078e8203056e1159f6229cb1ab3ee8643) with qc-deprecated-class-reference.sparql since these genes were also referenced in gene associations in other Mondo classes.

PR #8902, opened 2025-03-25
Add query to remove gene associations from obsolete classes - OMIM pipeline update

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8464

This query and additional step in the OMIM gene pipeline is to remove any gene association that was added to a Mondo class that is now obsolete so that this case does not trigger errors with the Mondo QC checks.


twhetzel:
@matentzn can you review this today?

matentzn (approved):
This looks correct, but only if you are certain that a gene axiom only has exactly one source annotation and certainly nothing else. 

If to works it works :) Thanks for poking!

twhetzel:
| gene axiom only has exactly one source annotation and certainly nothing else
This query is intended to be used with the OMIM gene pipeline so yes, the gene association on the obsolete classes are those added by the OMIM gene pipeline and only have a source annotation from what I've seen so far.

PR #8934, opened 2025-04-02
combine goals so dates are accessible as script params

twhetzel:
Combine the `make` goals so the dates are accessible to the scipt


twhetzel:
I tested the GitHub Action on this branch and provided dates and it worked. This branch does not "know" of the tags so the GH Action still fails when no dates are provided _and_ running the action from this branch. I'll merge into `master` to re-test.

PR #8954, opened 2025-04-14
Add synonym provenance for EMC

matentzn:


src/sparql/update/insert_emc_synonym_provenance.ru
```diff
@@ -0,0 +1,24 @@
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+PREFIX mondo: <http://purl.obolibrary.org/obo/mondo#>
+prefix owl: <http://www.w3.org/2002/07/owl#>
+
+INSERT {
+  ?syn_annot oboInOwl:hasDbXref ?xref_value .
+}
+WHERE {
+    VALUES ?syn_type {  
+        mondo:NORD_LABEL
```
twhetzel, line 10: Mondo does not have a `NORD_LABEL`.

matentzn: It should and I cannot for the like of me figure out why it does not

matentzn: You can see the annotation in the robot template owl file.. It just does not end up in the OBO file for some reason. Can you make an issue and investigate? That problem is separate from this one here.

twhetzel: Sure, ticket created and I'll look into it.

twhetzel (approved):
LGTM

PR #8966, opened 2025-04-17
DO NOT MERGE- tutorial on unsatisfiable class

MeeSiing:
mycotoxin allergy was created as an unsatisfiable class to demonstrate logic and debugging. however, the class was not reported as unsatisfiable.


sabrinatoro:
@MeeSiing I think we can close this PR since it was used to share information with Jim, and we don't need it anymore. Do you agree? If so, please close. Thank you!

Issue #8968, opened 2025-04-19
Add new terms related to the RAPID proposal


sabrinatoro:
New terms based on the work for the RAPID proposal. 
Spreadsheet with new terms is here: https://docs.google.com/spreadsheets/d/1M8LiK6aCOEUatsrRE02O2CcCbpwUZCaac5Kfx0g5wFo/edit?gid=0#gid=0

Explanation for these new terms can be found in the proposal spreadsheet https://docs.google.com/spreadsheets/d/1LtK93GgeX0rTP6epThzLfQmDgoPWWcTfatMx2eeG2E0/edit?gid=1862201541#gid=1862201541

sabrinatoro:
Note: name for MONDO:0700293 will have to be reviewed.

nicolevasilevsky:
note from @sabrinatoro on the PR:

Note: we should consider obsoleting MONDO:0976138

PR #8969, opened 2025-04-19
New term for the rapid proposal

sabrinatoro:
Addresses #8968


sabrinatoro:
Note: 
- name for MONDO:0700293 will have to be reviewed.
- the terms we added with no definition. We should get back to this as soon as we can

sabrinatoro:
Note: we should consider obsoleting MONDO:0976138

nicolevasilevsky (approved):

PR #8979, opened 2025-04-23
change since date to 2017, add cell to save all github issues

twhetzel:
minor update to github issue analysis notebook to change since date to 2017 and add cell to save all fetched github issue data (limited to certain fields) to a local file


matentzn (approved):
Not sure what needs to be reviewed; generally I prefer using dedicated tools like grainyhead rather than storing such code in notebooks (which will inevitably result in loss when you are moving to another project), but that has nothing to do with this specific PR.

PR #8981, opened 2025-04-23

twhetzel:


sabrinatoro:
@twhetzel there are a lot of changes (~3700+ additions). It looks like most of them are from NORD - so I assume we got all the one we lost last month. Could you please confirm that this big amount of addition is as expected? Thanks!

twhetzel:
Yes, I already have an email out about that since we (developers) see those changes earlier in our other processes.

sabrinatoro:
I did not see an email about it, but I am glad these changes can be explained. I will approve and merge.

sabrinatoro (approved):

twhetzel:
Just noting that I haven’t received a response yet to the follow-up email I sent about the additions, so I wasn’t expecting this PR to be merged until that confirmation came through. I’ll still follow up to ensure we’ve verified that the changes are fully expected.

Also, to clarify—my earlier “yes” referred to the fact that I had sent the email and that we (on the dev side) had already seen these changes earlier in our internal process, not that I was confirming the changes were expected.

PR #8990, opened 2025-04-24
Merge Erb palsy

sabrinatoro:
Addresses #8968

Terms merged because:

Comment from RAPID proposal:

MONDO:0700303 - Erb palsy

http://purl.obolibrary.org/obo/MONDO_0004569 would be a synonym for this type of birth injury.

Note: I merged the terms instead of removing the not-released yet ID, because this x-ref was used in the proposal


src/ontology/mondo-edit.obo
```diff
@@ -84302,19 +84302,26 @@ property_value: IAO:0000589 "paralytic ileus (disease)" xsd:string
 [Term]
 id: MONDO:0004569
 name: brachial plexus neuropathy from injury
+def: "A palsy characterized by a paralysis of the arm caused by an injury to the upper group of the main nerves supplying it, specifically the upper trunk C5-C6 of the brachial plexus. Erb palsy most commonly, though not exclusively, arises as a result of an injury sustained during birth." [https://orcid.org/0000-0002-4142-7153, PMID:30020632]
 subset: gard_rare {source="MONDO:GARD"}
```
katiermullen, line 84306: @sabrinatoro can you kindly confirm that brachial plexus neuropathy from injury is the same as Erb palsy? In the current definition for Erb palsy, the condition arises from injury to C5-C6 which is only part of the brachial plexus (C5-T1). A brachial plexus neuropathy from injury could affect all 5 nerves C5-T1. Please see https://www.cincinnatichildrens.org/health/b/brachial-plexus

katiermullen (changes requested):

src/ontology/mondo-edit.obo
sabrinatoro: Thank you @katiermullen. 
One of the experts said it was the same, but according to your reference, erb palsy is different, and should be a subclass of. In addition, the other types of branchial plexus could be added.

sabrinatoro:
Closing this PR in favor of https://github.com/monarch-initiative/mondo/pull/9036

Issue #8940, opened 2025-04-03
[Revise text def] MONDO:0859083 is lacking a definition


ikeseler:
MONDO:0859083 systemic lupus erythematosus 17

The definition within OMIM includes the text "evidence that systemic lupus erythematosus-17 (SLEB17) is caused by heterozygous mutation in the TLR7 gene"

Similar to MONDO:0013743 (autosomal systemic lupus erythematosus type 16), the definition for MONDO:0859083 could be "An instance of systemic lupus erythematosus (disease) that is caused by mutations in TLR7."

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0003-1738-6117

sabrinatoro:
Closed by mistake

PR #9004, opened 2025-04-28
add definition to MONDO:0859083

MeeSiing:
added definition to systemic lupus erythematosus 17 (MONDO:0859083) per issue#8940


src/ontology/mondo-edit.obo
```diff
@@ -555743,6 +555743,7 @@ relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/
 [Term]
 id: MONDO:0859083
 name: systemic lupus erythematosus 17
+def: "An instance of systemic lupus erythematosus (disease) that is caused by mutations in TLR7." [https://orcid.org/0000-0003-1738-6117, OMIM:301080]
```
sabrinatoro, line 555746: @MeeSiing 

While the definition is not incorrect, we should follow the definition as defined in this pattern: [disease_series_by_gene](https://github.com/monarch-initiative/mondo/blob/master/src/patterns/dosdp-patterns/disease_series_by_gene.yaml)

In addition, please add the following as cross-reference to the definition:
MONDO:patterns/disease_series_by_gene

sabrinatoro (changes requested):
@MeeSiing please update the following:
- definition should follow the pattern for disease defined based on gene (see comments for details)
- add the link to the GH issue to the term
- in the PR, please make sure to use "close #8940", so that the issue is closed when this issue is closed. (note that right now, the issue is not linked from the PR because you need a space before the hash sign)

Thank you!

sabrinatoro (changes requested):
@MeeSiing 
Requesting a very minor thing.

sabrinatoro (approved):
Looks great! Approving and merging

PR #9006, opened 2025-04-29
Add xref and synonym queries for stats

twhetzel:


twhetzel:
@matentzn are you able to review this PR or can suggest someone else?

src/sparql/reports/COUNT-all_diseases_synonym_stats.sparql
```diff
@@ -36,21 +40,32 @@ WHERE {
...
+      ("NORD" "exact_NORD")
+      ("ClinGen" "exact_ClinGen")
+      ("icd11.foundation" "exact_ICD11")
+      ("OMIA" "exact_OMIA")
+      ("SCTID" "exact_SCTID")
+    }
 
-    FILTER(?category != "exact_other")
+    BIND(STRBEFORE(STR(?xref), ":") AS ?actual_prefix)
+    FILTER(?actual_prefix = ?xref_prefix)
```
matentzn, line 60: Never used it like this (FILTER with a constraint on a key of a VALUES statement) but you would know if it does not work!

matentzn (approved):
Looks ok! I am concerned generally about maintainability of queries which have hard coded URLs and CURIE prefixes in them, but its not always possible to generalise.

twhetzel:
Thanks for the review! Is the concern with [this section](https://github.com/monarch-initiative/mondo/pull/9006/files#diff-b85620797e1e2fc3fb3952435eaba96c5974dffa839e738878799b2a789d99c7R28-R45)? If yes, I can put in a ticket in the Mondo repo to get this cleaned up in the ontology itself.

PR #9036, opened 2025-05-08
Erb Palsy child of "brachial plexus neuropathy"

sabrinatoro:
Address #8968

This PR replaces https://github.com/monarch-initiative/mondo/pull/8990

[This reference](https://www.cincinnatichildrens.org/health/b/brachial-plexus) notes that "Horner Syndrome" is also a brachial plexus neuropathy. However, I think Horner syndrome can involve a few neurons outside the brachial plexus. Therefore I did not add this child.


katiermullen (approved):
Looks great!

PR #9044, opened 2025-05-09
merge autosomal recessive hyper-IgE syndrome

### LINKED ISSUE #7844: MONDO:0957426 autosomal recessive hyper-IgE syndrome

nicolevasilevsky:
close #7844


sabrinatoro (approved):
Looks good. 
Approving and merging

PR #9052, opened 2025-05-11
Updated stats

twhetzel:
- The sources page is updated
- New tables added to Stats section


matentzn (approved):

matentzn:
We need some way to keep the docs up to date with every release, but as a one-off update for the paper this is ok.

PR #9059, opened 2025-05-15
Update malacards to xsd:anyURI

matentzn:
@twhetzel 

This is the only solution I could see - switch from <IRI> to `xsd:anyURI` syntax :P 

We use the same solution for seeAlso annotations in Mondo.


twhetzel (approved):
I checked out the branch and reviewed the examples mentioned in the Tech call notes ('Alzheimer disease' - MONDO:0004975 and 'thalassemia minor' MONDO:0044210 ) in Protege and these links do resolve. Odd this was needed for malacards, but not clingen entries.

matentzn:
@twhetzel I think the reason is that the OBO Format parser somehow recognises the `_` in the final part of the URL as a OBO PURL style separator..

PR #9077, opened 2025-05-20
Fix `replaced_by` annotation in `MONDO:0010107`

cthoyt:
OQuat identified that [`MONDO:0010107` (obsolete testicular regression syndrome)](https://bioregistry.io/mondo:0010107) had an invalid CURIE in its `replaced_by` annotation: `MONDO:800012` only has 6 characters.

It doesn't appear that it was missing a leading `0`, as `MONDO:0800012` does not exist. Neither does adding a trailing `0`. Adding a middle `0` lead me to [`MONDO:8000012`
 (neurologic, endocrine, and pancreatic disease, multisystem, infantile-onset 1)](https://bioregistry.io/mondo:8000012), which isn't relevant, _but_ right next to it is [`MONDO:8000017` (obsolete testicular regression syndrome)](https://bioregistry.io/mondo:8000017).

Seems like a bingo, yet another term that matches! Interestingly, this one is replaced by [`MONDO:8000015` (46,XY sex reversal 11)](https://bioregistry.io/mondo:8000015).

So it looks like `MONDO:0010107` got split, then merged back together, but we have some vestigial terms.

```mermaid
flowchart LR
   a["MONDO:0010107<br>obsolete testicular regression syndrome"] --> b["MONDO:8000017<br>obsolete testicular regression syndrome"]
   a --> c["MONDO:8000015<br>46,XY sex reversal 11"]
   b --> c
```

Should we therefore update to point to MONDO:8000017 or directly to MONDO:8000015?


cthoyt:
@matentzn @nicolevasilevsky r.e. mondo paper updates from oquat. Do mondo PRs usually need @ mentions to get reviewed?

matentzn:
I have unfollowed all repos so if you want my attention then yes, @ mention me better

PR #9096, opened 2025-05-22
add new term Lipschütz ulcer MONDO:1060113

### LINKED ISSUE #8956: Add new term: Lipschütz ulcer
### Label

Lipschütz ulcer

### Synonyms

Acute genital ulceration, ulcus vulvae acutum

### Synonym type

exact

### Definition

Acute genital ulceration, also known as "Lipschütz ulcer" or "ulcus vulvae acutum," is an uncommon, self-limited, nonsexually transmitted condition characterized by the rapid onset of painful, necrotic ulcerations of the vulva or lower vagina. It typically occurs in sexually inactive adolescent females or young females and may be preceded by influenza-like or mononucleosis-like symptoms. Acute genital ulceration has been associated with acute Epstein-Barr virus (EBV) infection or other viral and bacterial infections. However, in many cases a cause cannot be determined. https://www.uptodate.com/contents/acute-genital-ulceration-lipschutz-ulcer. 

### Parent term

vulvar disorder or vulvar disease

### Children term(s)

_No response_

### ORCID Identifier

0000-001-9310-0163

### Website URL

_No response_

### Additional comments

We had a request from a patient and a researcher to add the term.

 The precise prevalence is unknown but , but it is considered uncommon with sparse cases reported in the literature. 
(https://pmc.ncbi.nlm.nih.gov/articles/PMC10238287/).
There is a paper (https://pubmed.ncbi.nlm.nih.gov/26297242/) that is cited by many where is said that it causes around 30% of the cases of genital ulcers in  general (see: https://pmc.ncbi.nlm.nih.gov/articles/PMC5483547/). However, the 30% estimate applies only in a specific subset of patients (as per calculations it would affect around 150000 women at the most if we consider that around 500000 women are diagnosed with any kind of genital ulcers in USA per year), so it does not contradict its rarity and there are only few cases reported in the literature. Also, about 90% of reported cases occur in females aged 20 years or younger).  That said I believe that it may be considered rare (despite being underdiagnosed). Please consider it for inclusion.
Thank you, Gioconda Alyea

MeeSiing:
close #8956
- add this new term to "ulceration of vulvar" because it is a more specific form, where it is typically non-sexually transmitted, with rapid onset of painful and necrotic ulceration.
- add the definition, xref, and exact synonyms with the new term
- add "exclude from qc" - single child


nicolevasilevsky (approved):
looks good!

Issue #9097, opened 2025-05-22
Add ICD10 x-refs


katiermullen:
Add ICD10 x-refs that have been reviewed by a medical expert and deemed "certainly correct."

See https://docs.google.com/spreadsheets/d/1wwQi7HTm7u2B0IBhkvjNPx-360YXgGNhz9aKysClNZ4/edit?gid=1096054646#gid=1096054646

PR #9108, opened 2025-05-23
Add ICD10 x-refs

katiermullen:
Addresses #9097

(7 of 11)


katiermullen:
Fixed proxy merge failure following expert advice:

"Note: MONDO:0008797 anodontia is defined as "developmental dental anomaly characterized by the complete absence of all teeth", while ICD10CM:K00.0 Anodontia implies also partial anodontia / hypodontia / oligodontia. Therefore, from my point of view, MONDO:0005486 tooth agenesis is the correct match."

I agree with this expert advice and I removed the ICD10CM:K00.0 MONDO:equivalentTo on the term MONDO:0008797 'anodontia.' 

ICD10CM:K00.0 is equivalent to MONDO:0005486 'tooth agenesis.'

nicolevasilevsky (approved):

PR #8570, opened 2025-01-14
Generate Overall Mondo Stats

twhetzel:
Generate Mondo Stats as requested [here](https://docs.google.com/document/d/1vPl2x8gfTypjzcb9ND8qXXfKIjPDE4RwuExmzAUpM6Q/edit?tab=t.0). See the README for stats and how they were generated.

@sabrinatoro As future work, it would be helpful to know if any of these stats should be generated as report(s) for the Mondo releases. This was mentioned at the very end of the Tech call.

PR #9118, opened 2025-05-24
add notebook and new queries to analyze overlaps in subsets

twhetzel:
This is the notebook and additional SPARQL queries used to generate the counts of overlaps of subsets.

### DIFF
--- a/src/scripts/notebooks/MondoSubsetOverlaps.ipynb
+++ b/src/scripts/notebooks/MondoSubsetOverlaps.ipynb
@@ -1,0 +1,643 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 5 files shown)


matentzn (approved):
I checked only the queries, but they look right. Note that benign tumours are not always considered to be cancers so this concept may be overcounting the cancer category. Not my concern though😅

twhetzel:
These "parent" classes for the categories were given to me before by one of the curators (Nicole or Sabrina) and these have been used in [PRs](https://github.com/monarch-initiative/mondo/pull/8570/files#r1917474165) since Jan 2025.

Issue #8984, opened 2025-04-23
Investigate why NORD_LABEL is not being added into Mondo from EMC pipeline


twhetzel:
Investigate why the NORD_LABEL is not being added into Mondo from the EMC pipeline. See related conversation here https://github.com/monarch-initiative/mondo/pull/8954#discussion_r2055379018

twhetzel:
From the Mondo side, one of the issues is that the synonymtypedef for `NORD_LABEL` does not exist in Mondo, this will be added. 

Other issues, based on some ChatGPT debugging, are that ROBOT handles annotation axioms at the OWL level and when converting back to OBO, ROBOT does not expose or allow granular editing of OBO synonym: line components (like inserting a NORD_LABEL) before serialization. Based on this and Nico's [previous efforts](https://github.com/monarch-initiative/mondo-ingest/issues/831) to debug why the `NORD_LABEL` is not being added I choose the text editing route with sed. 

However, the sed option also has an issue not because of sed, but because of how the notion of a string being a preferred label for a community is currently modeled in Mondo. In the OBO spec, only one synonymtypedef is allowed per synonym, therefore a synonym can not be both an `ABBREVIATION` and a `NORD_LABEL` or any other synonymtypedef value, e.g. a `CLINGEN_LABEL`. Therefore, while this current solution does add a `NORD_LABEL` annotation if there is already an existing synonymtypedef when `NORM` runs the previous synonymtypedef will be lost. This will be reported out to a file and added into the PR comment when the GH Action runs.

matentzn:
> , therefore a synonym can not be both an ABBREVIATION and a NORD_LABEL or any other synonymtypedef value, e.g. a CLINGEN_LABEL.

THANK YOU SO MUCH for tracking that down! 

That makes total sense - I think we need to change the modelling back to "xref: MONDO:NORD_PREFERRED" or some such. Same with CLINGEN_LABEL, before the pattern proliferates. Given your finding we should not use synonym types for modelling this information.

twhetzel:
Yes, I agree both of these should be modeled differently and had submitted this issue https://github.com/monarch-initiative/mondo/issues/9122. The "real" longtime fix will need to be a modeling solution that works in _both_ OBO format and OWL and will need some updates to the file format we collect from the community providers and some (minor) code changes.

PR #9120, opened 2025-05-24
Refactor adding the NORD_LABEL to synonyms

twhetzel:
closes https://github.com/monarch-initiative/mondo/issues/8984

Refactor the `update-nord` make goal to add the `NORD_LABEL` with sed since the ROBOT merge does not work. Also, report cases where the synonym looses an existing synonymtypedef since only one is allowed per synonym in the OBO spec.

This was tested locally by running `sh run.sh make update-nord -B` in this branch. Once this PR merged, I will update the GH Action to include the report of conflicts with synonymtypedef into the PR comment created by the "Update Externally Provided Content" GH Action.

- [x] Submit issue for alternative modeling proposal for community labels https://github.com/monarch-initiative/mondo/issues/9122

### DIFF
--- a/src/sparql/update/delete_nord_label_synonym_type.sparql
+++ b/src/sparql/update/delete_nord_label_synonym_type.sparql
@@ -1,0 +1,21 @@
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+
+# Delete NORD_LABEL annotation represented as synonymtypedef. OBO spec supports only one synonymtypedef per synonym.
+# This query is run after "insert_nord_label.sparql".
+
+DELETE {
+  ?axiom oboInOwl:hasSynonymType ?syn_type .
+}
+WHERE {
+  ?class a owl:Class ;
+         oboInOwl:hasExactSynonym ?syn_value .
+
+  ?axiom a owl:Axiom ;
+         owl:annotatedSource ?class ;
+         owl:annotatedProperty oboInOwl:hasExactSynonym ;
+         owl:annotatedTarget ?syn_value ;
+         oboInOwl:hasSynonymType ?syn_type .
+
+  FILTER(STR(?syn_type) = "http://purl.obolibrary.org/obo/mondo#NORD_LABEL")
+}
--- a/src/sparql/update/insert_nord_label.sparql
+++ b/src/sparql/update/insert_nord_label.sparql
@@ -17,7 +17,4 @@ WHERE {
          oboInOwl:hasSynonymType ?syn_type .
 
   FILTER(STR(?syn_type) = "http://purl.obolibrary.org/obo/mondo#NORD_LABEL")
-  FILTER NOT EXISTS {
-    ?axiom oboInOwl:source "MONDO:NORD_LABEL"
-  }
 }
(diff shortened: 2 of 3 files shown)


twhetzel:
@matentzn do you have time today to review this?

matentzn:
> 
@matentzn do you have time today to review this?


Yes but this is not a complete solution - only adding NORD_LABEL when there is no CLINGEN_LABEL. 

I would prefer we use sparql update to rewrite the nord template.owl file to use synonym source=MONDO:NORD_LABEL syntax instead so at least the metadata is completely represented.. We can do the same for Clingen in a second round

twhetzel:
Can these changes be handled with updates to the community templates?

twhetzel:
@matentzn feedback needed on the questions above

matentzn:
@twhetzel I am also happy to do it if you have enough other Mondo tech things that need your attention

twhetzel:
I have time this week to do this. It would be great to have your Mondo time on the ODK 1.6 issue for mondo-ingest so we can move forward with things needed from that update.

twhetzel:
@matentzn I updated this here in the mondo repo to change modeling of NORD_LABEL as you suggested, ie use synonym source=MONDO:NORD_LABEL

src/sparql/update/insert_nord_label.sparql
```diff
@@ -0,0 +1,23 @@
...
+         oboInOwl:hasExactSynonym ?syn_value .
+
+  ?axiom a owl:Axiom ;
+         owl:annotatedSource ?class ;
+         owl:annotatedProperty oboInOwl:hasExactSynonym ;
+         owl:annotatedTarget ?syn_value ;
+         oboInOwl:hasSynonymType ?syn_type .
+
+  FILTER(STR(?syn_type) = "http://purl.obolibrary.org/obo/mondo#NORD_LABEL")
+  FILTER NOT EXISTS {
```
matentzn: This filter seems redundant (timewise a bit expensive) as normalisation `make NORM` will kill the redundant axiom?

src/sparql/update/delete_nord_label_synonym_type.sparql
```diff
@@ -0,0 +1,21 @@
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+
+# Delete NORD_LABEL annotation represented as synonymtypedef. OBO spec supports only one synonymtypedef per synonym.
+# This query is run after "insert_nord_label.sparql".
+
+DELETE {
+  ?axiom oboInOwl:hasSynonymType ?syn_type .
+}
```
matentzn, line 9: Why not just do both at the same time:

```suggestion
}
INSERT {
  ?axiom oboInOwl:source "MONDO:NORD_LABEL" .
}
```

matentzn (approved):
This looks great. I make two small suggestions to increase performance, but I am happy even if you don't want to apply them - feel free to update and merge, or just merge, no need to ask me again! good one!

src/sparql/update/insert_nord_label.sparql
twhetzel: Good point that the `make NORM` step will remove any duplicates. The FILTER step in the query is removed.

src/sparql/update/delete_nord_label_synonym_type.sparql
twhetzel: It took awhile to get the insert query working to sort out this line `FILTER(STR(?syn_type) = "http://purl.obolibrary.org/obo/mondo#NORD_LABEL")`. I tried one combination of the insert/delete, but it did not work so leaving as is.

PR #9135, opened 2025-05-28
Add synonym for intellectual disability, autosomal dominant 5

StephanieMarsh:
close#9042


src/ontology/mondo-edit.obo
```diff
@@ -307290,6 +307290,7 @@ synonym: "autosomal dominant intellectual disability 5" EXACT [https://orcid.org
 synonym: "autosomal dominant mental retardation 5" EXACT DEPRECATED [DOID:0070035]
 synonym: "autosomal dominant non-syndromic intellectual disability 5" RELATED [DOID:0070035]
 synonym: "autosomal dominant non-syndromic intellectual disability caused by mutation in SYNGAP1" EXACT [MONDO:design_pattern]
+synonym: "Epilepsy due to SYNGAP mutations" EXACT [PMID:29588993] {IAO:0000233="https://github.com/monarch-initiative/mondo/issues/9042", http://purl.org/dc/elements/1.1/creator="ORCID:0000-0002-1532-8725"}
```
nicolevasilevsky, line 307293: synonyms in Mondo should all be lower case

the term tracker item should be a separate annotation and should be the datatype anyURI
image

the creator should be not in the synonym annotations

nicolevasilevsky (changes requested):
see comments inline - changes needed

nicolevasilevsky (approved):

PR #9166, opened 2025-06-05
add new general stats query to count xrefs for selected sources

twhetzel:
A new general stats query is added to count the total number of xrefs (MONDO:equivalentTo) for selected sources. This query is in addition to the query to count the distinct number of Mondo classes that have at least one xref (MONDO:equivalentTo) to selected sources.

### DIFF
--- a/src/sparql/reports/COUNT-selected_xrefs.sparql
+++ b/src/sparql/reports/COUNT-selected_xrefs.sparql
@@ -1,0 +1,26 @@
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+PREFIX MONDO: <http://purl.obolibrary.org/obo/MONDO_>
+
+# Get the total count of xrefs for the list of selected sources.
+
+SELECT ?xref_source (COUNT(*) AS ?xref_count)
+WHERE {
+  ?class a owl:Class ;
+         oboInOwl:hasDbXref ?xref .
+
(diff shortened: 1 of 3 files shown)


src/sparql/reports/COUNT-selected_xrefs.sparql
```diff
@@ -0,0 +1,26 @@
...
+PREFIX MONDO: <http://purl.obolibrary.org/obo/MONDO_>
+
+# Get the total count of xrefs for the list of selected sources.
+
+SELECT ?xref_source (COUNT(*) AS ?xref_count)
+WHERE {
+  ?class a owl:Class ;
+         oboInOwl:hasDbXref ?xref .
+
+  FILTER(REGEX(STR(?xref), "OMIM:|OMIMPS:|Orphanet:|DOID:|ICD10CM:|icd11.foundation:|OMIA:|SCTID:"))
```
matentzn, line 12: I wonder if we need this filter or can just generate stats for all prefixes?

matentzn (approved):
Small optional comment, looks good

src/sparql/reports/COUNT-selected_xrefs.sparql
twhetzel: I think it could be this "selected" list Sabrina has mentioned or all prefixes. If all prefixes however, if automation is added to move the results of this query content into the [Mondo Community web site](https://mondo.monarchinitiative.org/#stats), that process will need to parse to sources of interest. 

Do you have a preference based on that comment to keep the selected sources in the query?

PR #9181, opened 2025-06-10
Fix some source references in SSSOM

matentzn:


twhetzel:
I am checking as well, but do you know if/what are the downstream consequences of this change?

twhetzel:
@matentzn how does the timeline look to get mondo-ingest updated to ODK 1.6?

matentzn:
This is just metadata that is not actively used in the pipeline, just copied into the sssom files

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -97,51 +98,51 @@ relations:
...
     metadata:
-      object_source: http://purl.obolibrary.org/obo/doid.owl
+      object_source: infores:disease-ontology
       object_source_version: http://purl.obolibrary.org/obo/doid/releases/2018-07-05/doid.owl
       mapping_date: "2018-07-05"
   - from: MONDO
     to: icd11.foundation
     metadata:
-      object_source: http://purl.obolibrary.org/obo/mondo/sources/icd11foundation.owl
+      object_source: infores:icd11-foundation
```
twhetzel, line 148: should this be `infores:icd11-foundation` or `infores:icd11.foundation`?

matentzn: The resource does not technically exist yet 😜 https://github.com/biolink/information-resource-registry/pull/59

But the convention is in that resource is using a dash

PR #9183, opened 2025-06-10
Remove OGMS from import modules

matentzn:
Since we do not use it in logical definitions or similar, there is not obvious reason keeping OGMS identifiers in the imports module.


matentzn:
@twhetzel I tested this and it works as expected (`sh run.sh make imports/merged_import.owl`)

twhetzel (approved):
Looks great! Thanks for testing and adding the test command.

PR #9192, opened 2025-06-12
Autoimmune progesterone dermatitis/urticaria

### LINKED ISSUE #8069: Autoimmune progesterone dermatitis/urticaria
Hi! You have this condition in SNOMED:400009001.  I wanted to let you know that we have a NORD report known as "Progestogen Hypersensitivity". According to the medical literature it appears to be rare, with fewer than 200 reported cases, although it may be under-recognized, and there are no published estimates of incidence or prevalence [https://www.uptodate.com/contents/progestogen-hypersensitivity] Also, while "autoimmune progesterone dermatitis" is commonly used, "dermatitis is just one component in the range of reactions that have been reported, thus, the term "progestogen hypersensitivity" was subsequently proposed as a more accurate description". [https://www.jaci-inpractice.org/article/S2213-2198(23)00916-9/abstract] 

Therefore, I am requesting this term to be added to MONDO as a rare NORD condition and change the main name you currently have to  
 "progestogen hypersensitivity".

Thank you, Gioconda Alyea

katiermullen:
Closes #8069

- updated name
- updated definition and added comment
- updated superclass to 'hypersensitivity reaction disease'
- added excluded subclass of 'autoimmune disease' because PMID: [37579875](https://pubmed.ncbi.nlm.nih.gov/37579875/) says "“Autoimmune progesterone dermatitis” is a misnomer as, in most cases, there is no evidence of autoimmunity."
- added synonyms


src/ontology/mondo-edit.obo
```diff
@@ -499831,20 +499831,23 @@ property_value: seeAlso "https://rarediseases.info.nih.gov/diseases/9461/autoimm
...
-xref: MESH:C535299 {source="MONDO:equivalentTo"}
-xref: SCTID:400009001 {source="MONDO:equivalentTo"}
-xref: UMLS:C1260879 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source="MEDGEN:220359"}
-is_a: MONDO:0002406 {source="https://orcid.org/0000-0001-5208-3432"} ! dermatitis
-is_a: MONDO:0007179 {source="MESH:C535299", source="MONDO:Redundant"} ! autoimmune disease
+xref: MEDGEN:220359 {source="MONDO:relatedTo", source="MONDO:MEDGEN"}
+xref: MESH:C535299 {source="MONDO:relatedTo"}
+xref: SCTID:400009001 {source="MONDO:relatedTo"}
+xref: UMLS:C1260879 {source="MONDO:relatedTo", source="MONDO:MEDGEN", source="MEDGEN:220359"}
+is_a: MONDO:0000605 {source="PMID:37579875"} ! hypersensitivity reaction disease
```
sabrinatoro, line 499848: @katiermullen Why did you make these x-ref "MONDO:relatedTo"?

katiermullen: @sabrinatoro - I made these related to because they are for synonyms like "autoimmune progesterone dermatitis", which are technically incorrect since this is not an autoimmune condition and not always a dermatitis. I can be convinced that the synonyms and x-refs should be exact since they represent the same concept and now we just know more about what the concept actually is, but on the other hand, they are kind of historical as I put in the rdfs:comment. 

Please let me know your thoughts.

sabrinatoro: My understanding with this issue is that "everyone" got it sort of not quite right. I think keeping these are equivalent would make more sense. Also, MedGen and UMLS will be automatically updated. 
I suggest that we keep them equivalent.

katiermullen: Thanks, @sabrinatoro! I'll update these!

Issue #5687, opened 2022-11-30
[Revise subclass]     Fanconi anemia


pnrobinson:
**Mondo term (ID and Label)**
[Fanconi anemia](https://www.ebi.ac.uk/ols/ontologies/mondo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMONDO_0019391#)

**Suggested revision and reasons**
Fanconi does not always have all of the associaed symptoms, so it is not always true that it is a  [congenital limb malformation](https://www.ebi.ac.uk/ols/ontologies/mondo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMONDO_0019391#)

I do not like the term "congenital hematological disorder" because most of the diseases are genetic but not all necessarily manifest at birth. Can we rename this to Mendelian hematological disorder?
thougts @cmungall @ahamosh  ?

pnrobinson:
Same ting applies to "congenital myopathy"

cmungall:
I agree on both - I made an issue to tackle all congenital-X en masse #5695

ahamosh:
Can we talk about this tomorrow?  I am confused.  Are you saying that you want to rename Fanconi Anemia? I totally agree that the anemia part of Fanconi Anemia is not often or even usually congenital.  I feel differently about congenital myopathy because the congenital myopathies do present at birth (or they aren't congenital myopathies).  Let's discuss tomorrow.

PR #9196, opened 2025-06-12
remove SC for Fanconi Anemia

sabrinatoro:
Addresses #5687

I remove the SC since it was not correct. However, since there was no sources for this SC, I only deleted it instead of adding it as a "excluded subclass of" annotationa


katiermullen (approved):
@sabrinatoro I approve this PR, but wonder if a source is needed for hematologic disorder? For example, https://fanconi.org/clinical-care-guidelines/clinical-care-of-fanconi-anemia-hematologic-issues/. 

If you think this appropriate, I can add this source. Thank you!

sabrinatoro:
Thank you! This is a great comment. I will add it.

PR #9197, opened 2025-06-12
[Revise text def] Caroli syndrome

### LINKED ISSUE #7941: [Revise text def] Caroli syndrome
**Mondo term (ID and Label)**
https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_0018808?lang=en

It is unclear what the purpose of the current definition is. Do we want to obsolete this term? (We shouldnt)
Here is a source for definitions
Yonem O.
Bayraktar Y.
Clinical characteristics of Caroli’s disease.
World J Gastroenterol. 2007; 13: 1930-1933

katiermullen:
closes #7941


sabrinatoro (approved):
looks good! 
Approving and merging

PR #9204, opened 2025-06-13
Consolidate "content stats" with "general stats"

twhetzel:
Consolidate queries originally developed to provide content statistics as needed for presentations, etc. into the set of queries run as the "general stats" for Mondo release statistics. Remove obsolete content stats documentation.


matentzn (approved):
lgtm

PR #9217, opened 2025-06-13
New Term REEP6-related retinopathy

yshwetar:
- New term
- Defintion
- Term Tracker
- ClinGen Synonym
- Logical Definition
- One OMIM child terms


src/ontology/mondo-edit.obo
```diff
@@ -617788,6 +617790,15 @@ is_a: MONDO:0019118 {source="https://orcid.org/0000-0002-0587-4693"} ! inherited
 intersection_of: MONDO:0019118 {source="https://orcid.org/0000-0002-0587-4693"} ! inherited retinal dystrophy
 intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/9454 {source="https://orcid.org/0000-0002-0587-4693"} ! PROM1
 
+[Term]
+id: MONDO:1040066
+name: REEP6-related retinopathy
+def: "Any retinopathy caused by variants in the REEP6 gene." [https://clinicalgenome.org/working-groups/clinical-domain/ocular-cdwg/, https://orcid.org/0000-0001-5935-2391, https://orcid.org/0000-0002-0587-4693]
+synonym: "REEP6-related retinopathy" EXACT CLINGEN_LABEL [https://clinicalgenome.org/working-groups/clinical-domain/ocular-cdwg/, https://orcid.org/0000-0001-5935-2391]
+is_a: MONDO:0019118 {source="https://clinicalgenome.org/working-groups/clinical-domain/ocular-cdwg/", source="https://orcid.org/0000-0001-5935-2391"} ! inherited retinal dystrophy
```
sabrinatoro: The relation to the gene (probably via the logical definition) is missing.

src/ontology/mondo-edit.obo
```diff
@@ -362104,8 +362104,10 @@ xref: MEDGEN:934593 {source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 xref: OMIM:617304 {source="DOID:0080350", source="MONDO:equivalentTo"}
 xref: UMLS:C4310626 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source="MEDGEN:934593"}
 is_a: MONDO:0019200 {source="DC-OMIM:617304", source="DOID:0080350", source="MONDO:Redundant", source="OMIM:617304"} ! retinitis pigmentosa
+is_a: MONDO:1040066 {source="https://clinicalgenome.org/working-groups/clinical-domain/ocular-cdwg/", source="https://orcid.org/0000-0001-5935-2391"} ! REEP6-related retinopathy
 intersection_of: MONDO:0019200 ! retinitis pigmentosa
 intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/30078 ! REEP6
+relationship: excluded_from_qc_check http://purl.obolibrary.org/obo/mondo/sparql/qc/general/qc-single-child.sparql
```
sabrinatoro: Did you confirm that this "excluded from QC check" (ie having only one child) makes sense? Please add a comment on this term and/or the parent term to document how the terms are different.

sabrinatoro (changes requested):
Please see comments in line

src/ontology/mondo-edit.obo
yshwetar: Addressed!

src/ontology/mondo-edit.obo
yshwetar: I am confused by this - retinitis pigmentosa is the phenotypic presentation of this genetic mutation (which happens to be the 77th gene associated with retinitis pigmentosa). We always have OMIM terms as children of ClinGen Genotype-Phenotype terms, even as single children

sabrinatoro: @yshwetar We don't "always" have OMIM terms as children of ClinGen term. 
If there is one single child, we must have evidence that the OMIM term is different from the ClinGen term. Because it they are not different, then there should be only 1 Mondo term, and the ClinGen name should be added as a synonym

PR #9218, opened 2025-06-13
New Term PRPF4-related retinopathy

yshwetar:
- New term
- Defintion
- Term Tracker
- ClinGen Synonym
- Logical Definition
- One OMIM child terms


src/ontology/mondo-edit.obo
```diff
@@ -346350,8 +346350,10 @@ xref: MEDGEN:863118 {source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 xref: OMIM:615922 {source="DOID:0110392", source="MONDO:equivalentTo"}
 xref: UMLS:C4014681 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source="MEDGEN:863118"}
 is_a: MONDO:0019200 {source="DC-OMIM:615922", source="DOID:0110392", source="MONDO:Redundant", source="OMIM:615922"} ! retinitis pigmentosa
+is_a: MONDO:1040067 {source="https://clinicalgenome.org/working-groups/clinical-domain/ocular-cdwg/", source="https://orcid.org/0000-0001-5935-2391"} ! PRPF4-related retinopathy
 intersection_of: MONDO:0019200 ! retinitis pigmentosa
 intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/17349 ! PRPF4
+relationship: excluded_from_qc_check http://purl.obolibrary.org/obo/mondo/sparql/qc/general/qc-single-child.sparql
```
sabrinatoro, line 346356: Do you have evidence that this "excluded QC-single child" is correct? Please add a comment on this term (and/or on the parent term) to indicate how this term is different from the parent term.

sabrinatoro (commented):
@yshwetar 
This PR is perfect. Please add a comment about how the 1 child term is different from the parent term (since it is not obvious, and we might end up merging these terms later if we don't document why these are different.

sabrinatoro (changes requested):
see comments

PR #9230, opened 2025-06-17
Update language translations

matentzn:


.github/workflows/translations.yaml
```diff
@@ -0,0 +1,33 @@
+name: Update Mondo translations and test them
+
+on:
+  schedule:
+    - cron: '0 14 * * 1'
```
twhetzel, line 5: is this `cron` needed?

.github/workflows/translations.yaml
```diff
@@ -0,0 +1,33 @@
...
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.5.4
```
twhetzel, line 11: consider updating to 1.6

matentzn: We should move all of mondo (all actions and the build pipeline) to ODK 1.6 at the same time.

.github/workflows/translations.yaml
matentzn: Yes, this way no one needs to think about language translations; the cron will just run from time to time (its not at all time sensitive so I think this is fine).

.github/workflows/translations.yaml
twhetzel: Ok, I have a question for you in https://github.com/monarch-initiative/mondo/pull/9231 the ODK1.6 update PR

PR #9234, opened 2025-06-18
Removed conflicting prefixes

matentzn:
This PR removes one conflicting prefix which we do not use in Mondo (ICD11) and changes one expansion. 

It also updates the infores prefix URI which has been set up a few days ago.

I have run a complete release and can confirm all works as expected.


twhetzel (approved):
Look good.

PR #9264, opened 2025-07-01
refresh imports

twhetzel:
Refresh imports


sabrinatoro (commented):
I wonder whether the following will be a problem:
 
Screenshot 2025-07-01 at 9 14 48 AM

twhetzel:
That one (OGMS:0000031) also caught my eye. I am not sure why the name was removed, I do not see a change in OGMS https://www.ebi.ac.uk/ols4/ontologies/ogms/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FOGMS_0000031.

PR #9277, opened 2025-07-02
merge MONDO:0020071 infantile epilepsy syndrome' into MONDO:0100022 n…

nicolevasilevsky:
…eonatal/infantile epilepsy syndrome'

- merge MONDO:0100207 infantile-onset epilepsy' into MONDO:0100022 neonatal/infantile epilepsy syndrome'

address #8668


src/ontology/mondo-edit.obo
```diff
@@ -557273,17 +557254,31 @@ property_value: seeAlso "https://www.epilepsydiagnosis.org/syndrome/idiophatic-p
 id: MONDO:0100022
 name: neonatal/infantile epilepsy syndrome
 def: "An epilepsy sydrome that has an onset during the neonatal or infantile stage of life." []
```
sabrinatoro, line 557256: @nicolevasilevsky we need a x-ref for this definition

sabrinatoro (changes requested):
@nicolevasilevsky one one small change requested: the x-ref for the definition. Unfortunately I don't have time at this point to add it myself (because I don't know which x-ref would be appropriate).

PR #9288, opened 2025-07-04
adding review instructions for copilot

cmungall:


.github/copilot-instructions.md
```diff
@@ -0,0 +1,347 @@
...
+
+```
+property_value: http://purl.org/dc/terms/creator doi:10.1186/s13326-024-00320-3
+```
+
+You don't need to add a date, this will be handled for you
+
+## Relationships
+
+All terms should have at least one `is_a` (this can be implicit by a logical definition, see below).
```
twhetzel, line 128: I don't believe this is true as some terms only have a parent specified by a logical definition

cmungall: that's what the parenthetical comment is about - anyway you can make an edit or we can see how it does, we can always tweak later!

twhetzel:
FWIW - the Chrome Extension can have access to APIs if those are the kinds of tools you are thinking of.

cmungall:
@twhetzel 

perhaps the workflow can be

- curator makes issue
- curator or ontology editor refines issue using chrome extension
- dragon-ai makes PR
- copilot does first pass review [auto-triggered]
- ontology editor does final review, with the help of the chrome extension, and modifies instructions as necessary

PR #9304, opened 2025-07-10
adding review instructions for copilot

cmungall:


.github/copilot-instructions.md
```diff
@@ -0,0 +1,347 @@
...
+relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/23386 {source="PMID:36200388"} ! MYCBP2
+```
+
+Multiple sources can be cited for the same relationship.
+
+## 6. Gene Identifier Verification
+
+## Gene Identifiers
+
+**CRITICAL**: Always verify gene identifiers using the official HGNC database:
```
twhetzel, line 243: this should be updated to account for non-human animal diseases with a gene association since these gene identifiers are pulled from NCBI Gene. 

@katiermullen or @twhetzel to update

PR #9306, opened 2025-07-11
copilot enabling

### LINKED ISSUE #9305: Set up github copilot for this repo
@matentzn @twhetzel @kevinschaper @caufieldjh @justaddcoffee maybe you could investigate this while I am away. 

https://docs.github.com/en/copilot/how-tos/agents/copilot-coding-agent/best-practices-for-using-copilot-to-work-on-tasks#pre-installing-dependencies-in-github-copilots-environment

This operates in pretty much the same way as dragon-ai-agent, we'd need to customize it with MCPs, see here for pros/cons:

See also https://github.com/ai4curation/aidocs/issues/3

cmungall:
- **removed**
- **symlink**
- **set up copilot, fixes #9305**


.github/copilot-instructions.md
twhetzel, line 1: I see this is symlinked to CLAUDE.md, however a lot of what was in this file is not in the CLAUDE.md file and these examples look useful. Should more content from the copilot instructions file be moved into the CLAUDE.md file?

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Add tools to PATH
+      run: |
+        echo "${{ github.workspace }}/tools" >> $GITHUB_PATH
+        ls -alt ${{ github.workspace }}
+        ls -alt ${{ github.workspace }}/tools
+      shell: bash
+
+    - name: Add obo-scripts to PATH
+      run: |
+        git clone https://github.com/cmungall/obo-scripts.git ${{ github.workspace }}/tools/obo-scripts
```
twhetzel, line 73: do the obo-scripts exist in any other project directory vs. a personal repo? for initial testing these seems ok

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,86 @@
...
+      with:
+        path: ~/.jar-cache
+        key: ${{ runner.os }}-robot-v1.9.7
+        restore-keys: ${{ runner.os }}-robot-
+
+    - name: Download ROBOT JAR if not cached
+      run: |
+        mkdir -p ~/.jar-cache
+        if [ ! -f ~/.jar-cache/robot.jar ]; then
+          curl -L https://github.com/ontodev/robot/releases/download/v1.9.7/robot.jar -o ~/.jar-cache/robot.jar
```
twhetzel, line 52: ODK version 1.5.4 uses ROBOT version 1.9.8. is something incompatible with this ROBOT version?

matentzn: Or to ask it another way - why not just use ODK instead of downloading the JAR file?

caufieldjh:
Would it be enough to include this under `copilot-setup-steps`?
```yaml
container: obolibrary/odkfull:latest
```

.github/copilot-instructions.md
cmungall: I definitely encourage trying adding more examples!

.github/workflows/copilot-setup-steps.yml
cmungall: If you can figure out how to do that!

https://github.com/ai4curation/aidocs/pull/5

.github/workflows/copilot-setup-steps.yml
cmungall: not sure I follow the Q....

twhetzel: This is cloning a repo in your personal GitHub space, which is ok for now. I was asking if this repo with obo-scripts existed anywhere else within an existing project.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,86 @@
...
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+
+jobs:
+  # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
```
matentzn, line 23:
```suggestion
    runs-on: ubuntu-latest
    container: obolibrary/odkfull:1.6
```

Lets just try this?

twhetzel: Looks good!

caufieldjh:
I think the instructions will need to include an introduction to the ODK toolset and some examples.
In a perfect world without the need for trial and error, this would mean including the entire OBOOK text.
I don't think we're quite at that point yet, but there is also the chance that the underlying model may have some ODK exposure already.
Only one way to find out.

PR #9324, opened 2025-07-15

twhetzel:


twhetzel:
I spot checked the new NORD content and things look correct. The NORD QC failures are [here](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/external/nord-qc-failures.md) and the original NORD download file is from https://rdbdev.wpengine.com/wp-content/uploads/mondo-export/rare_export.tsv.

sabrinatoro (approved):
Looks good to me.

PR #9331, opened 2025-07-16
added 'PRSS56-related nanophthalmos' as exact synonym

### LINKED ISSUE #9126: PRSS56-related nanophthalmos
PRSS56-related nanophthalmos

**Synonyms**
isolated microphthalmia 6 (MONDO:0013293)

microphthalmia (MONDO:0021129)

**Definition**
Any nanophthalmos in which the cause of the disease is variants in the PRSS56 gene. Please note that this term can include cases diagnosed as microphthalmos (specifically as posterior microphthalmos) or nanophthalmos.

**Definition source (Please give PubMed ID, if applicable, in format PMID:#######)**
This term has not yet been described in a publication, but has been recommended by the ClinGen Glaucoma / Neuro-Ophthalmology GCEP as a more specific alternative to the existing term "isolated microphthalmia 6"

**Children terms (if applicable)**

https://clinicalgenome.org/affiliation/40077/

MeeSiing:
close #9126

added 'PRSS56-related nanophthalmos' as an exact synonym to isolated microphthalmia 6 (MONDO:0013293) because according to the requester, this new term encompasses cases diagnosed as microphthalmos/microphthlamia and nanophthalmos/nanophthalmia
- added CG working group as reference
- added Clingen label
- added term tracker


sabrinatoro (commented):
@MeeSiing I am not sure that this is what the requester asked. 
They said that the requested term is more specific than "isolated microphthalmia 6"(MONDO:0013293)
I think we need to reach out to the requester and ask whether the concept of the requested term is the same or different from MONDO:0013293 (ie, it is different,maybe more specific, than "an isolated microphthalmia in which the cause of the disease is a mutation in the PRSS56 gene"). If it is, we need to know how. It it is the same thing, than I agree with your adding a synonym to this term.

MeeSiing:
@sabrinatoro The requester mentioned that 'PRSS56-related nanophthalmos' is more specific than 'isolated microphthalmia 6' because 'PRSS56-related nanophthalmos' encompasses both microphthalmia and nanophthalmia associated with PRSS56 gene, therefore the new term is a lumping term. Will updating the synonym scope to "related" be acceptable?

sabrinatoro:
@MeeSiing 

> The requester mentioned that 'PRSS56-related nanophthalmos' is more specific than 'isolated microphthalmia 6' because 'PRSS56-related nanophthalmos' encompasses both microphthalmia and nanophthalmia associated with PRSS56 gene, therefore the new term is a lumping term. 

The new term is not a lumping term. 
A lumped term is a term that includes several more specific disease concept. 
In this case, 'isolated microphthalmia 6' is defined (I am assuming) as a disease with microphtalmia.
The requester is asking for a more specific term, ie a term that includes microphtalmia AND nanophtamia. 
Therefore this requested term should be a more specific type of the existing term, and therefore a child


> Will updating the synonym scope to "related" be acceptable?
It is my understanding that the users are requesting a more specific term than the one we currently have. (if this is a misunderstanding, please let me know) They are requesting a new term because they probably need this more specific term. If they do need this term, we should create a new one instead of adding the term as a synonym. 
The users need to be able to look for their term and find the corresponding disease concept. By adding a non-exact synonym, they will not know what term corresponds to their concept.

MeeSiing:
@sabrinatoro I get what you mean. I am happy to create it as a new term. Will classify it under 'isolated anophthalmia-microphthalmia syndrome' be fine then? I don't think it can be classified under 'nanophthalmia' knowing that it includes microphthalmia too (though it is not reflected in the definition and term label).
Screenshot 2025-07-21 at 2 51 18 PM

MeeSiing:
@sabrinatoro, Clingen got back to us and agreed that the requested concept, PRSS56-related nanophthalmos matches the existing term MONDO:0013293 isolated microphthalmia 6 and therefore should be added as an exact synonym. I had updated the definition of MONDO:0013293 based on our discussion and previous reply to the Clingen group. 
They shared their concerns regarding using ‘PRSS56-related nanophthalmos and/or microphthalmos’, and suggested 'PRSS56-related nanophthalmos or posterior microphthalmos' or just 'PRSS56-related nanophthalmos'. I think we can use 'PRSS56-related nanophthalmos' because they explained that cases diagnosed as posterior microphthalmos, "usually there are some effects on anterior dimensions especially with age", so that "nanophthalmos pretty much encompasses all of the PRSS56 cases". Let me know if you think otherwise.

wchankey:
Thank you both for your patience with us and your help to create the new synonym!

PR #9338, opened 2025-07-18

twhetzel:


twhetzel:
I see the [Mondo QC issue](https://github.com/monarch-initiative/mondo/actions/runs/16376653241/job/46278152823?pr=9338#step:4:37), but I don't recall how these are fixed. This is the validation query that is failing: https://github.com/monarch-initiative/mondo/blob/master/src/sparql/qc/mondo/qc-ordo-subset-exact-mapping.sparql so there are Mondo terms in the `ordo_disorder` subset that have xrefs to Orphanet that have a source `MONDO:equivalentObsolete`.

Is the `ordo_disorder` subset intended to be removed for these terms?

@sabrinatoro if you have time to fix these instead, please feel free to re-assign the PR.

twhetzel:
Given the Mondo QC error and that we ran the pipelines a bit out of the usual order (mini build of EMC content, EMC pipeline run, then full data build), let's close this PR and re-run the EMC pipeline again rather than manually fixing these issues.

PR #9349, opened 2025-07-21
Delete all import files no longer needed

matentzn:
@sabrinatoro asked me to remove all import related files that are no longer needed. I have tested this by running the imports pipeline. Not sure if something else broke though.


twhetzel:
The last issue we saw with removing imports was when running the refresh imports, e.g. see [here](https://github.com/monarch-initiative/mondo/pull/9264/commits/58304f2f8598f916e7876f5d78e69a8a80147db6#diff-889e0a9ba62d62cc6ee7c5c9bead0075444d506253590ef0dcf4786904ac6768R575018), so as long as this type of case was not seen in your refresh imports data then that should be good.

Are there other Mondo pipelines you are concerned might be affected?

Approving in the meantime.

matentzn:
@Copilot, can you check if any of the deleted files are used / needed in src/ontology/Makefile or src/ontology/mondo.Makefile?

matentzn:
> Are there other Mondo pipelines you are concerned might be affected?

TBH I don't know. As far as I could tell when doing my usual screening, no!

matentzn:
Lets just risk it, I have talked to ChatGPT and I feel good about this PR.

PR #9350, opened 2025-07-21

twhetzel:


sabrinatoro (approved):
This looks good. Approving and merging. 

Note to selves: we should think about a workflow when the x-ref are obsoleted at the source due to merge.

PR #9353, opened 2025-07-21

twhetzel:

- May need to revisit https://github.com/monarch-initiative/mondo/pull/8432 and linked items


twhetzel:
I reviewed the QC failures, but for 3 of the 4 `../sparql/qc/general/qc-deprecated-class-reference.sparql` failures I could not see what obsolete term was being referenced. 

For `../sparql/qc/general/qc-single-child.sparql`, [this reference](https://github.com/monarch-initiative/mondo/actions/runs/16430894140/job/46431910003?pr=9353#step:4:78) does not look like it should fail because it has two subclasses. 

@matentzn can you work with the curators to sort out the changes needed?

twhetzel:
Following up from Slack DM, @matentzn can you work with Sabrina to sort out the work to get this PR merged?

matentzn (changes requested):
I am a bit concerned that there are 4K NCIT references lost, 1.5K DOID confirmations lost, 4K Orphanet lost. These are so many lost classifications - I think we need to figure out why before we can proceed with this PR.

twhetzel:
@matentzn we have not merged in results from a subclass sync since your updates to the subclass sync https://github.com/monarch-initiative/mondo/pull/8432. We looked through some of these (NCIT and DOID) previously, but I think for this one more review is needed by you and/or curators.

matentzn:
This is a bit much of a curation task. Either I would have expected to have MORE confirmations, or I would have expected to have roughly the same but not so many less.. before we launch into a long review can you spot check 10 doids / ncits and see if they are correct according to these ontologies?

I already have a lot else to fix on this pr so it won't be merged anyways

twhetzel:
> can you spot check 10 doids / ncits and see if they are correct according to these ontologies?

yes, i can do that

twhetzel:
@matentzn with the work this past week on the Mondo/OMIM gene associations for susceptibility terms I have not been able to get back to check whether the DOID/NCIT information is correct. When do you need this information by?

twhetzel:
From 1:1 conversation with Nico --> close this branch and let’s come up with a plan to keep the the NCIT provenance for terms not in the neoplasm branch and use a different property for these NCIT sources of curated provenance. And consider using to fix the OMIMPS issue and eventually the Orphanet issue.

@twhetzel - still check ~10 DOIDs though from this PR

PR #9354, opened 2025-07-22
Remove mapping_date from sssom mappings sets

matentzn:


twhetzel:
I see the `mapping_date` is only commented out. Are there longer term plans to generate the date dynamically and include back in the file?

matentzn:
Yes we should really upgrade the metadata as a whole, including the mapping date.. when there is money

PR #9366, opened 2025-07-25

twhetzel:


twhetzel:
In reviewing the Mondo QC errors, there appears to be a bug in the pipeline that generates the synonym content to add into Mondo. 

In this Mondo QC error from https://github.com/monarch-initiative/mondo/actions/runs/16514613197/job/46703001318?pr=9366:
```
MONDO:0010159-MONDO:0031219,oboInOwl:hasExactSynonym-oboInOwl:hasExactSynonym,constitutional mismatch repair deficiency syndrome
MONDO:0031219-MONDO:0010159,oboInOwl:hasExactSynonym-oboInOwl:hasExactSynonym,constitutional mismatch repair deficiency syndrome
```

`MONDO:0010159` is the original term with this synonym based on review in OLS, however currently in the PR `MONDO:0031219` has a ClinGen label annotation. I checked the raw ClinGen file we ingest and `MONDO:0031219` is not listed in that file, only `MONDO:0010159`. Therefore the clingen label appears to be added back to the incorrect synonym.

PR #9293, opened 2025-07-07
merge DPH5-related diphthamide-deficiency syndrome

MeeSiing:
close #8326

merged DPH5-related diphthamide-deficiency syndrome with MONDO:0859295 'neurodevelopmental disorder with short stature, prominent forehead, and feeding difficulties'
- added definition for MONDO:0859295
- removed old ID and duplicate SC of 
- removed annotations "in subset obsoletion candidate" and "scheduled for obsoletion on or after"

added term tracker and obsolescence reason to the obsoleted term

PR #9369, opened 2025-07-25
added lost information

sabrinatoro:
Adding some information (source and definition related information) that were lost during https://github.com/monarch-initiative/mondo/pull/9293


katiermullen (approved):
approve and merging.

PR #9371, opened 2025-07-25
Alagille syndrome-JAG1

### LINKED ISSUE #8915: Alagille syndrome-JAG1
Alagille syndrome-JAG1 

**Synonyms**

**Parent term** MONDO:0007318

**Definition**
An autosomal dominant disorder that affects multiple organ systems including the liver, heart, skeleton, eyes and kidneys. Organ system involvement and severity may vary widely among patients and within families. The term lumps the disease assertions "Alagille syndrome 1", "Tetralogy of Fallot", and "Deafness, congenital heart defects, and posterior embryotoxon".

**Definition source 

**Children terms (if applicable)** 

https://clinicalgenome.org/affiliation/40066/

katiermullen:
Closes #8915 

Updated the clin gen preferred label on term MONDO:0016862 to Alagille syndrome-JAG1.

Also added textual and logical definitions, comment, and characteristic, "autosomal dominant inheritance."


src/ontology/mondo-edit.obo
```diff
@@ -402073,8 +402076,13 @@ xref: OMIM:118450 {source="Orphanet:261619", source="Orphanet:261619/ntbt", sour
 xref: Orphanet:261619 {source="MONDO:equivalentTo"}
 xref: UMLS:C1956125 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source="MEDGEN:365434"}
 is_a: MONDO:0007318 {source="OMIM:118450", source="Orphanet:261619"} ! Alagille syndrome
-relationship: has_characteristic SO:1000008 ! point_mutation
+intersection_of: MONDO:0007318 ! Alagille syndrome
+intersection_of: has_characteristic HP:0000006 ! Autosomal dominant inheritance
```
sabrinatoro, line 402080: @katiermullen should the autosomal dominance inheritance be part of the definition (ie is it a required condition, because if it is recessive, it is another disease)? 
If it is, then this should be added to the textual definition. 
If it is not (and it is a "feature" of the disease), then it should be removed from the logical definition and added as a SC (which is already is)

sabrinatoro (changes requested):
@katiermullen I have a comment in line. Thank you! (I know, you really like the 3 parts logical definitions... 😄 )

src/ontology/mondo-edit.obo
katiermullen: @sabrinatoro there is no autosomal recessive version of this disease to my knowledge, but OMIM and clin gen both assert that this disease is an autosomal dominant disorder. I think I should add this to the textual definition. Do you agree?

sabrinatoro: @katiermullen I think it should be removed from the logical definition

sabrinatoro (approved):
Approving this can be merged after QC checks have passed.

PR #9377, opened 2025-07-28
Refresh imports

twhetzel:
Refresh imports for August Mondo release


src/ontology/imports/merged_import.owl
```diff
@@ -26729,11 +26742,6 @@ Declaration(AnnotationProperty(<http://xmlns.com/foaf/0.1/homepage>))
 AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000115> <http://purl.obolibrary.org/obo/IAO_0000116> "An administrative note intended for its editor. It may not be included in the publication version of the ontology, so it should contain nothing necessary for end users to understand the ontology."@en)
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000116> "editor note"@en)
 
-# Annotation Property: <http://purl.obolibrary.org/obo/RO_0002175> (present in taxon)
```
sabrinatoro, line 26732: What does it mean for us that "present in taxon" is removed here?

src/ontology/imports/merged_import.owl
```diff
@@ -27237,15 +27226,6 @@ SubObjectPropertyOf(<http://purl.obolibrary.org/obo/RO_0002159> <http://purl.obo
 SymmetricObjectProperty(<http://purl.obolibrary.org/obo/RO_0002159>)
 TransitiveObjectProperty(<http://purl.obolibrary.org/obo/RO_0002159>)
 
-# Object Property: <http://purl.obolibrary.org/obo/RO_0002160> (only in taxon)
```
sabrinatoro, line 27240: What does it mean for us that this is removed here?

src/ontology/imports/merged_import.owl
twhetzel: @matentzn ^^ 
do you have any insight into why this is removed?

src/ontology/imports/merged_import.owl
twhetzel: @matentzn ^^
do you have any insight into why this is removed?

src/ontology/imports/merged_import.owl
sabrinatoro: Note that we use "present in taxon" for annotations (though no SC annotation). For example: [dog disease](http://purl.obolibrary.org/obo/MONDO_0024919)

src/ontology/imports/merged_import.owl
sabrinatoro: It doesn't look like we use "only in taxon" in Mondo

src/ontology/imports/merged_import.owl
twhetzel: I still see `RO:0002175` in Mondo in this git feature branch (and RO) and used with MONDO:0024919 'dog disease'. 

Of the 50 classes I reviewed of a total of 1326 classes where `RO:0002175` is removed, it is removed from CL classes.

UPDATE: It is removed from Uberon classes, e.g.  [here](https://github.com/monarch-initiative/mondo/pull/9377/files#diff-9f31de5fce8559e8d8b1e3b701ffd4c4191f8356342b63e0ef1ba10755de5f4fL149457)

matentzn: Very good catch, careful review. 

seems that used to be a bug not discovered before odk 16. fixed now https://github.com/monarch-initiative/mondo/commit/9194a352120ad66f81fff2704e0f17d9f44d9af7

i updated the import too

sabrinatoro (approved):
Approving this PR. Thank you @twhetzel and @matentzn

PR #9382, opened 2025-07-29
Update MONDO:Redundant status

matentzn:
This PR adds a workflow to update MONDO:Redundant annotations in mondo-edit.obo.

The workflows removes all such annotations (from subclass and excluded subclass), and then adds them back (only to subclass) if a term is _asserted to be a subclass of a non-parent ancestor_. Note that while this is certainly correct, its highly incomplete, as it does not catch cases where a reasoner is needed. Still, better then what it currently is, and still useful. If we wanted to use a reasoner, we would have to assert all inferred subclass axioms which is not ideal. You can test the pipeine locally by running:

```
sh run.sh make remove_redundant_subclass_annotations
sh run.sh make add_redundant_subclass_annotations
sh run.sh make NORM && mv NORM mondo-edit.obo
```


matentzn:
@twhetzel the diff in mondo-edit.obo is to large to push an actual run of the pipeline through, but the pipeline can be merged and added to the week 3 workflows docs

twhetzel:
Did you have a day/pipeline slot you were thinking when this should be run?

Does it need to be run before the August Mondo release?

twhetzel:
@matentzn - I ran the commands in the first comment, but I do not have any changes in `mondo-edit.obo`.

matentzn:
There was a typo in the make commands.. Please try again!

Issue #8302, opened 2024-11-04
Revise MONDO_0010945 (RP17)


laraholtes:
MONDO_0010945
retinitis pigmentosa 17

In 2020, complex structural variants were identified at a locus on chromosome 17 in 22 autosomal dominant retinitis pigmentosa families (PMID: 33022222). These structural variants segregated with disease and are thought to disrupt the three dimensional chromatin architecture of the locus and lead to ectopic gene expression in the retina. A variant in the carbonic anhydrase IV (CA4) gene had been previously implicated as the cause of retinitis pigmentosa type 17 in families of South African origin, however pathogenicity of the reported variant has been questioned because it has a population frequency of 4% in healthy controls in northern Sweden (PMID: 20238024). Furthermore, the families of South African origin were also found to harbour a complex structural variant in the RP17 locus (PMID: 33022222). For more information on retinitis pigmentosa type 17, see the OMIM entry (600852).

Lara K. Holtes
ORCID: 0009-0000-6337-4403

sabrinatoro:
@yshwetar I think this issue was fully addressed with PR https://github.com/monarch-initiative/mondo/pull/9385. Please confirm and if you think everything has been addressed, please close this issue. Thank you!

PR #9385, opened 2025-07-29
Update RP17

yshwetar:
Addresses #8302

- Remove logical def
- Update text def
- Add PMIDs + ORCID


src/ontology/mondo-edit.obo
```diff
@@ -255989,8 +255989,6 @@ xref: MESH:C563437 {source="MONDO:equivalentTo"}
 xref: OMIM:600852 {source="DOID:0110404", source="MONDO:equivalentTo"}
 xref: UMLS:C1833245 {source="MEDGEN:322153", source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 is_a: MONDO:0019200 {source="DC-OMIM:600852", source="DOID:0110404", source="MESH:C563437", source="MONDO:Redundant", source="OMIM:600852"} ! retinitis pigmentosa
-intersection_of: MONDO:0019200 ! retinitis pigmentosa
-intersection_of: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/1375 ! CA4
 relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/1375 {source="MONDO:mim2gene_medgen"} ! CA4
```
sabrinatoro, line 255992: This annotation should also be removed too.

sabrinatoro (changes requested):
@yshwetar 
One small change requested. 
(I am not updating myself so you are aware of the things your missed)

src/ontology/mondo-edit.obo
yshwetar: Addressed!

sabrinatoro:
@yshwetar I think this PR did address everything in the issue, and therefore instead of "addresses", we could have used the word "closes".

If you think that everything requested has been addresses, please close the original issue (it is still open). Thank you

PR #9411, opened 2025-08-02
Update OMIM gene references

twhetzel:
Update OMIM gene references.

PR #9415, opened 2025-08-04
refresh imports

twhetzel:
refresh imports after adding back Mondo/OMIM susceptibility gene associations and OMIM provenance in #9411


sabrinatoro (approved):
This looks good. 
Approving and merging

Issue #9418, opened 2025-08-05
Remove "MONDO:mim2gene_medgen" evidence from gene links


matentzn:
We now import and update the latest gene evidence directly from OMIM. I would suggest we drop the `"MONDO:mim2gene_medgen"` evidence as it is simply stale. This is a massive PR with 4300 lines changes, so we will have to do it fast to avoid conflicts.

@sabrinatoro I will deal with all this, I just need your ok.

sabrinatoro:
I approve of this move and give you my OK.
We will (potentially) end up with gene association without evidence. 
I suggest we coordinate so that, when you create the PR, I (or another curator) can review asap and ensure that we have a source for the gene association (or remove that association) as part of this clean-up. 
Thank you, Nico!

sabrinatoro:
I believe this is done and this issue can be closed

PR #9419, opened 2025-08-05
Create qc-subclass-relation-no-source.sparql

matentzn:
This QC check ensures that subclass axioms with relationships (eg gene relations) must have evidence associated with it.

@sabrinatoro there are 47 gene associations without evidence. There are three options for you to have the QC pass:

1. Manually remove them
2. Let me remove tham all in bulk now
3. Have a cleanup process that automatically removes them when they arise, for example after an OMIM update

#9418 might result in another number of relations that need to be cleaned up, but that will be a close second.

### DIFF
--- a/src/sparql/qc/general/qc-subclass-relation-no-source.sparql
+++ b/src/sparql/qc/general/qc-subclass-relation-no-source.sparql
@@ -1,0 +1,30 @@
+PREFIX owl:        <http://www.w3.org/2002/07/owl#>
+PREFIX rdfs:       <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX oboInOwl:   <http://www.geneontology.org/formats/oboInOwl#>
+PREFIX eco:        <http://purl.obolibrary.org/obo/eco#>
+PREFIX RO:         <http://purl.obolibrary.org/obo/RO_>
+
+SELECT ?entity ?property ?value
+WHERE {
+
+  VALUES ?property { 
+    RO:0004003 
+  }
+  ?entity rdfs:subClassOf ?restr .
+
+  ?restr a owl:Restriction ;
+         owl:onProperty   ?property ;
+         owl:someValuesFrom ?value .
+
+  FILTER isIRI(?entity)
+  FILTER STRSTARTS(STR(?entity), "http://purl.obolibrary.org/obo/MONDO_")
+
+  FILTER NOT EXISTS {
+    ?ax a owl:Axiom ;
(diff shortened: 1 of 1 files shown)


src/sparql/qc/general/qc-subclass-relation-no-source.sparql
```diff
@@ -0,0 +1,30 @@
...
+  FILTER isIRI(?entity)
+  FILTER STRSTARTS(STR(?entity), "http://purl.obolibrary.org/obo/MONDO_")
+
+  FILTER NOT EXISTS {
+    ?ax a owl:Axiom ;
+        owl:annotatedSource   ?entity ;
+        owl:annotatedProperty rdfs:subClassOf ;
+        owl:annotatedTarget   ?restr ;
+        # any recognised evidence predicate
+        (oboInOwl:hasDbXref | oboInOwl:source | eco:hasEvidence) ?evidence .
```
twhetzel: @sabrinatoro are any of these properties acceptable to gene association provenance or is this limited to `oboInOwl:source` in Mondo?  If all of these properties are acceptable, then other gene association QC queries should be reviewed and expanded for these properties.

sabrinatoro: For evidence of **SubClassOf** annotation, **at this point**, we are only using `oboInOwl:source`.

We are talking about adding another property (similar to "source") to make the distinction between when the external source said a SC existed, and when a curator used information from that source to create this SC. But we do not have this property yet. 

To my knowledge, oboInOwl:hasDbXref and eco:hasEvidence have not been in use for Mondo SC.

sabrinatoro:
@matentzn Taking a quick look, i think we need to make a manual review of these and add evidence. 
I suspect (based on the few examples I looked at) that most of them will have an OMIM source, and will have to be added to not be updated automatically. 

Should we do the review and add the sources in this PR?

matentzn:
> Should we do the review and add the sources in this PR?

There are not too many, so yes! :) Thanks!!

src/sparql/qc/general/qc-subclass-relation-no-source.sparql
matentzn:
```suggestion
        oboInOwl:source ?evidence .
```

matentzn: You are right, this was redundant.

sabrinatoro (approved):

PR #9420, opened 2025-08-05
Exclude OMIMPS from subclass sync process

matentzn:
@sabrinatoro this is what we discussed, this will ensure that OMIMPS related evidence is not removed during the subclass sync.


twhetzel (approved):
LGTM - I ran the GH Action for this branch with results [here](https://github.com/monarch-initiative/mondo/pull/9430) and did not see any OMIMPS sources removed.

PR #9424, opened 2025-08-05
Create subclass-axioms-only-supported-by-ordo.sparql

matentzn:


twhetzel (approved):
Looks good. I made some cosmetic changes to remove unused prefixed and add docstrings. I tested queries as: `sh run.sh make report-query-%` where `%` is replaced with the query name.

PR #9439, opened 2025-08-08
Add OAK pronto QC check

### LINKED ISSUE #9435: Re-opening #7579
I'll duplicate here my latest comments. I apologize that it took me so long to respond.

I've just checked the latest versions of Pronto (2.7.0) and Mondo (2025-08-05), and there is an error:
```
graphdata.lineage[superentity].sub.add(subentity)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'BFO:0000031'
```

I think this is the same error as before. Dangling class that breaks Pronto's lineage construction.

This is the only problematic term. There are other BFO terms in mondo.obo but they are declared in the file while BFO:0000031 is referenced in is_a clause but not declared anywhere.

My understanding is that this behavior is allowed by the OBO format specification but it's strange that all the other terms are declared in mondo.obo except this one, so I wonder if this may be an issue in this Mondo release? It seems that the intention was to avoid such behavior.

matentzn:
This is adding a QC check that ensures that mondo.obo is still parseable with OAK. Users have been complaining multiple times.

Fixes #9435 for good.


twhetzel:
@matentzn can you review the [Mondo QC error](https://github.com/monarch-initiative/mondo/actions/runs/16831726828/job/47680855868?pr=9439#step:4:421):
```
runoak -i pronto:mondo.obo ontology-metadata > tmp/mondo-metadata.txt
/usr/local/lib/python3.12/dist-packages/oaklib/implementations/pronto/pronto_implementation.py:177: UnicodeWarning: unsound encoding, assuming utf-8 (99% confidence)
  ontology = Ontology(str(resource.local_path), **kwargs)
KeyError: 'BFO:0000031'
```

matentzn:
> @matentzn can you review the [Mondo QC error](https://github.com/monarch-initiative/mondo/actions/runs/16831726828/job/47680855868?pr=9439#step:4:421):

Thats exactly the correct error! Yes. It fails because pronto fails to parse it. With #9442 merged this will now pass.

PR #9442, opened 2025-08-08
Refactor and refresh imports

matentzn:
1. Removing ensemble genes from Mondo imports which are only used in CL types
2. Adding a BFO import - since mondo.owl has _most_ of BFO in it through BFO imports, this just fills the gap. A user noticed one class being used but not declared, which is now imported correctly. This may also result in some beneficial QC
3. Removing src/sparql/construct/construct-ncbigene.sparql from git as the file is dynamically generated


sabrinatoro (approved):
Things look ok, I am not seeing any red flags here. 
I wonder how that will impact the Mondo ontology though (in terms of new classification and potentially - though hopefully no - QC failures). 
I am approving.

matentzn:
if QC passes this wont have any implications on classification.. it is too high level.

PR #9457, opened 2025-08-12
update xrefs for exanthem

### LINKED ISSUE #8744: Update xrefs on MONDO:0006547 exanthem
MONDO:0006547 exanthem

MONDO:0006547 has two xrefs to NCIT as seen in the screenshot below

Image

The xref to NCIT:C39594 'Skin Rash' does not look correct, ie equivalentTo vs. MONDO:mondoIsNarrowerThanSource, and the xref to NCIT:C111884 'Exanthem' looks like it should be updated to MONDO:equivalentTo. 

Here's a definition from DermNet https://dermnetnz.org/topics/exanthems (apols I could not quickly find a good PubMed reference):
```
Exanthem is the medical name given to a widespread rash that is usually accompanied by systemic symptoms such as fever, malaise and headache. It is usually caused by an infectious condition such as a virus, and represents either a reaction to a toxin produced by the organism, damage to the skin by the organism, or an immune response.
```
While all exanthems are considered to be skin rashes, not every skin rash can be classified as an exanthem.

https://orcid.org/0000-0002-3458-4839

**Other comments:**

MeeSiing:
close #8744 

updated NCIT xrefs for exanthem
edited the synonym scope and the definition for exanthem based on the NCIT xrefs


sabrinatoro (changes requested):
@MeeSiing I don't think any of these x-ref should be equivalent. 
Both are in "otherhierarchy", which means that they are not in the "disease/disorder" branch of NCIT (it is in the "finding" branch, which represents "phenotypes" or "lab results".

I think the to-do is: 

- [ ] make both these NCIT "MONDO:otherHierarchy". 
- [ ] update the definition to make sure it refers to a disease

MeeSiing:
> [ ] make both these NCIT "MONDO:otherHierarchy".

@sabrinatoro do I remove the "MONDO:equivalentTo" from NCIT:C111884 and "MONDO:mondoIsNarrowerThanSource" from  NCIT:C39594?

sabrinatoro:
@MeeSiing: yes, you should remove all the source that are "MONDO:equivalent" and "MONDO:narrowerthansource", and keep the "other hierarchy"

sabrinatoro (approved):

PR #9472, opened 2025-08-15
Community preferred labels as language tags

matentzn:
This is a bit criminal, but hey. 

Lets trial this with the next release and see if there are complaints.

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -1657,6 +1657,7 @@ $(TRANSLATIONSDIR)/%.babelon.json: $(TRANSLATIONSDIR)/%.babelon.tsv
 
 $(ONT)-international.owl: $(ONT).owl $(TRANSLATIONS_OWL)
 	$(ROBOT) merge $(patsubst %, -i %, $^) \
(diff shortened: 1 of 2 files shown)


twhetzel:
I think this is clever, but as you also say a bit criminal. Is the goal of doing this so that the language tag in OLS can be used to switch the display label? We already have various subsets of Mondo where the label is switched for the specific community label and did a lot of work to get a new OMO property added and incorporated into Mondo. 

I am also testing out the code and will approve after reviewing the results.

twhetzel (approved):
I tested the updated code and it's working as expected and Approving. 

For the community specific labels tagged with `language: en-clingen` or `language: en-nord` as compared to the actual language translated label for `language: jp` there are no other annotations on the clingen and nord labels. I'm not sure that any are needed, but it looks bare and possibly as a bug without any other annotations that might explain or hint why `en-clingen` and `en-nord` are language tags.

matentzn:
After some back and forth with @twhetzel on slack I feel less confident about this PR. Anyways, the work is done, so if we have to we can always revive it and deliver it within a minute, but between the fact that we publish a community-specific custom subset already and the abuse of a standardised mechanism for language translations my resolve is a little shaky. Thanks for the feedback @twhetzel!

PR #9485, opened 2025-08-19

twhetzel:

### DIFF
--- a/src/ontology/debug_inference_check.owl
+++ b/src/ontology/debug_inference_check.owl
@@ -75,10 +75,31 @@
     
 
 
+    <!-- http://purl.obolibrary.org/obo/BFO_0000001 -->
+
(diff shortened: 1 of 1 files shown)


src/ontology/debug_inference_check.owl
twhetzel, line 1: I am not familiar with this file or why this was created as a result of running the EMC pipeline.

sabrinatoro: @matentzn what is this file? Should it be removed or is it ok to keep?

matentzn: There is no concern for this file, its ancient and does not seem very useful to me.. You can merge this safely, I can account for the change it shows.

twhetzel: Thanks Nico - I will merge now.

PR #9500, opened 2025-08-25
Add omim susceptibility subset and update gene pipeline

matentzn:
This PR 

1. adds all the 393 omim susceptibility tags we got from the OMIM pipeline (hope that ballpark number sounds right @sabrinatoro).
2. Ensures that all axioms related to OMIM susceptibilities (NOT any of their descendants!!!) get preserved during the mean OMIM gene update purges (they do not get deleted, are considered manually curated)

Needs 
- [x] a technical review by @twhetzel
- [ ] a curator review by @sabrinatoro, especially on the design assumptions above

@twhetzel I tested 

```
sh run.sh make tmp/keep-susceptibility-gene-assocation-axioms.owl
```

To make sure that we keep the "right" amount of omim susceptibility classes.

I tested:

```
sh run.sh make update-omim-subsets
```

To tested the subset pipeline whose result I am posting as part of the PR (updates to mondo-edit.obo).


src/ontology/mondo.Makefile
```diff
@@ -611,6 +611,7 @@ update-external-content-incl-rare:
 	$(MAKE) update-efo-subset -B
 	$(MAKE) update-clingen -B
 	$(MAKE) update-ordo-subsets -B
+	$(MAKE) update-omim-subsets -B
```
twhetzel, line 615: @matentzn are you ok with also adding this line to the `update-external-content` goal? I know we use the combined goal where you added this line here, but to keep things consistent I think it would also be good to add to `update-external-content`.

sabrinatoro:
@matentzn can you please point me towards a document (or query or something) explaining how "susceptibilities" are identified? Thank you!

src/sparql/reports/omim-susceptibilities.sparql
```diff
@@ -0,0 +1,16 @@
...
+WHERE {
+  ?cls inSubset: <http://purl.obolibrary.org/obo/mondo#omim_susceptibility> .
+  
+  # Extract MONDO ID from the IRI
+  BIND(REPLACE(STR(?cls), "http://purl.obolibrary.org/obo/MONDO_", "MONDO:") AS ?mondo_id)
+  
+  # Filter to only include MONDO terms
+  FILTER(STRSTARTS(STR(?cls), "http://purl.obolibrary.org/obo/MONDO_"))
+}
+ORDER BY ?mondo_id
```
twhetzel:
```suggestion
ORDER BY ?mondo_id

```

src/ontology/mondo-edit.obo
twhetzel, line 1: Mondo is missing the new subset property as a subproperty of `http://www.geneontology.org/formats/oboInOwl#SubsetProperty`

src/ontology/mondo.Makefile
twhetzel: I decided to add it for consistency between the two goals.

src/ontology/mondo.Makefile
```diff
@@ -706,10 +708,14 @@ update-rare-subset:
...
 $(TMPDIR)/keep-susceptibility-gene-assocation-axioms.owl: $(SRC)
+	$(MAKE) report-query-omim-susceptibilities
+	sed -i 's/"//g' reports/report-omim-susceptibilities.tsv
 	$(ROBOT) filter --input $(SRC) \
-		--term MONDO:0042489 \
-		--select "self descendants" \
+		-T reports/report-omim-susceptibilities.tsv \
+		--trim false \
+		filter \
 		--term RO:0004003 \
```
twhetzel, line 719: Are there any cases where both subclasses of MONDO:0042489 'disease susceptibility' and the OMIM susceptibility file need to be used to "keep" the correct subset of gene associations? For example, if a gene association is manually added for a subclass of MONDO:0042489 'disease susceptibility' would it be assumed that it would be maintained without checking that it also existed in OMIM?

twhetzel:
@sabrinatoro susceptibilities here are based on OMIM where the phenotype name is enclosed in braces, e.g. `{}`, based on https://omim.org/help/faq#1_6

src/ontology/mondo-edit.obo
twhetzel: I added this into the file.

twhetzel (approved):
This looks good to me. However, for the updated goal `$(TMPDIR)/keep-susceptibility-gene-assocation-axioms.owl` I added a comment because I'm not sure if there are any cases where both filters (the one removed and the one added) would be needed.

src/ontology/mondo.Makefile
matentzn: Looks good thank you!

src/ontology/mondo.Makefile
matentzn: @sabrinatoro?

Can we assume that your original instruction: "if and only if it is a gene related to an OMIM disease susceptibility, we always keep it"? In particular, this does _not_ relate to susceptibilities which are NOT reflected in OMIM: these gene associations will NOT have OMIM evidence no matter what.

sabrinatoro: What Nico said. 
If it is a manually created susceptibility term (we don't create them, but let's suspend this for now) and it doesn't have an OMIM x-ref, then the gene association will not have an omim evidence.

twhetzel: @matentzn is the goal ok as-is or need updates?

Issue #9501, opened 2025-08-26
Add MONDO:0019317 follicular atrophoderma-basal cell carcinoma to be a child of MONDO:0004993 carcinoma


elliottsharp:
MONDO:0019317 follicular atrophoderma-basal cell carcinoma
does not have a parent of 
MONDO:0004993 carcinoma

despite the disease resulting in carcinomas

I would recommend MONDO:0019317 become a child of MONDO:0004993

matentzn:
Thank you @elliottsharp for keeping up with making issues, much appreciated. I am running this by the curation team and will come back to you!

matentzn:
# AI-based analysis

Warning: The content of this AI-generated recommendation is being reviewed by a human reviewer at the moment.

**Status: Unreviewed by human expert**
**Model: Claude with Sonnet 4.**

## ✅ Why the user request is valid:

```mermaid
graph BT
    B[MONDO:0005341 skin basal cell carcinoma] --> A[MONDO:0004993 carcinoma]
    C[MONDO:0019317 follicular atrophoderma-basal cell carcinoma] -.->|PROPOSED<br/>NEW PARENT| A
    C --> D[MONDO:0010535 Bazex-Dupre-Christol syndrome]
    
    style C fill:#ffcccc
    style A fill:#ccffcc
    style B fill:#ccffcc
```

### Medical/Clinical Justification:
- **MONDO:0019317** represents a specific manifestation that results in basal cell carcinomas
- The term name explicitly includes "basal cell carcinoma" indicating it IS a type of carcinoma
- **MONDO:0010535** (Bazex-Dupre-Christol syndrome) is defined as "a rare genodermatosis with a predisposition to early-onset basal cell carcinomas"
- The synonym "follicular atrophoderma and basal cell carcinomas" clearly indicates carcinomatous nature

### Terminological Correctness:
- All other basal cell carcinoma terms in MONDO (MONDO:0002935, 0002936, 0002937, etc.) are properly classified under carcinoma hierarchy
- The term should follow the same classification pattern as other basal cell carcinomas
- Currently missing the carcinoma classification creates an inconsistency

### Ontological Gap:
- **MONDO:0019317** is currently only classified under the syndrome (MONDO:0010535) but not under the disease it actually represents (carcinoma)
- This creates a classification gap where a carcinoma is not recognized as such ontologically

## 💡 Suggested improvements:

```mermaid
graph BT
[...]

PR #9535, opened 2025-09-04
DO NOT MERGE (only for discussion) Claude GitHub issues

matentzn:


CLAUDE.local.md
matentzn, line 1: If I want to use claude locally and I dont have these scripts on my path, I just want to tell claude to run them locally.

CLAUDE.md
```diff
@@ -18,6 +18,7 @@ This includes instructions for editing the npso ontology.
 - Only search over `src/ontology/mondo-edit.obo`
 - DO NOT bother doing your own greps over the file, or looking for other files, unless otherwise asked, you will just waste time.
 - ONLY use the methods above for searching the ontology
+- Check first if `obo-grep.pl` is on your path. If it is not, use `perl ./obo-grep.pl` instead of `obo-grep.pl`
```
matentzn, line 21: This is a little bit of a conflict for what local claude needs to know and what dragon ai needs to know..

CLAUDE.md
```diff
@@ -417,6 +419,62 @@ property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/
 
 These improvements would help AI systems better understand MONDO's specific conventions and produce changes that more closely align with human expert curation practices.
 
+## GitHub Issue Handling
+
+When handling GitHub issues:
+
+1. **View the issue**: Use `gh issue view [number]` to read the issue content
```
matentzn, line 430: This is a bit of a conflict for what local claude needs to know when its running locally and what dragon ai / claude needs to know when it is running in github actions..

I dont know what the best strategy is here for keeping the two instructions separate..

CLAUDE.md
```diff
@@ -417,6 +419,62 @@ property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/
...
+2. **Analyze validity**: Assess if the request is medically and terminologically valid
+3. **Search for improvements**: Look, for example, for more specific parents/terms that might be better
+4. **Present findings** in the following format. Print the findings on the terminal, but also write them to a file @src/ontology/tmp/issue_x_analysis.md, where "x" should be the issue number:
```
matentzn, line 433: See [here](https://github.com/monarch-initiative/mondo/issues/9501#issuecomment-3252036750) for an example.

CLAUDE.md
cmungall: I'm not following. Keep it simple, just give it all the tools it needs on its path?

CLAUDE.md
cmungall: I'm missing context - why aren't these the same?

(there may be a separate issue, and that the official claude *action* (https://github.com/anthropics/claude-code-action/) has its own way of interacting with the issue, but I don't think that's what is happening here?

I don't get why every human and agent doesn't have `gh` on their path?

CLAUDE.md
```diff
@@ -417,6 +419,62 @@ property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/
...
+
+### Model information
+
+- Write a clearly visible warning that the report is AI generated
+- Write which model you are using for generating the report
+
+### ✅ Why the user request is valid:
+
+- Mermaid diagram that describes SPECIFICALLY THE USER REQUESTED change
```
cmungall, line 442: LOVE THIS

cmungall: might want to consider a subagent for this

cmungall:
Not sure why the perl scripts need to be checked in? Maybe nbd as these should ideally remain stable but I'm worried about forking this across multiple repos...

CLAUDE.md
cmungall: of course, I missed the giant "DO NOT MERGE (only for discussion)" at the top of my screen so you can likely ignore my comments!

matentzn: I am mostly using local claude not github actions.. I would prefer not to create configuration hell and have everyone download and add those scripts to their paths.. This PR here is not for merging though :P Dont worry I would not drive such as change before talking it through with you.

matentzn:
This is no longer needed in favour of #9691

PR #9539, opened 2025-09-05
Use curies implementation for extracting prefixes

cthoyt:
I'm doing a GitHub-wide cleanup of operations on dataframes with CURIEs in them, replacing with functionality that's tested and hardened in the upstream `curies` package

cc @matentzn


matentzn:
Nice, is this available in 

```
curies                        0.10.19
```

which is what we depend on?

cthoyt:
nope, i just added it today, so we need curies 0.10.21 (https://github.com/biopragmatics/curies/releases/tag/v0.10.21)

src/scripts/split_sssom_by_source.py
matentzn, line 1: Blocked by 0.10.21 requirement, which requires

- an ODK minor release
- migration of Mondo to ODK minor release

cthoyt: es lohnt sich!

actually i realize there was a tiny backwards compatible import change so we need 0.10.21, which I just released

twhetzel:
@matentzn are you managing this PR?

matentzn:
Yes. Assigned to myself now.

PR #9543, opened 2025-09-08
Update workshop.md to update Mondo outreach call schedule

sagehrke:
I removed any items from the outreach call schedule for the rest of 2025. We will re-add presentation dates if we end up scheduling any ad hoc outreach in the fall 2025.


twhetzel (approved):
lgtm

PR #9552, opened 2025-09-10
Update OMIM gene references

twhetzel:
Update OMIM gene references.


src/ontology/mondo-edit.obo
```diff
@@ -213433,7 +213433,7 @@ xref: UMLS:C0282512 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source=
 is_a: MONDO:0002254 {source="NCIT:C84806"} ! syndromic disease
 is_a: MONDO:0020072 {source="Orphanet:98818", source="https://orcid.org/0000-0001-8486-0558"} ! childhood-onset epilepsy syndrome
 is_a: MONDO:1060139 {source="PMID:27683935", source="https://orcid.org/0000-0001-9310-0163"} ! GRIN2A-related complex neurodevelopmental disorder
-relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/4585 {source="MONDO:mim2gene_medgen", source="OMIM:245570"} ! GRIN2A
+relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/4585 {source="MONDO:mim2gene_medgen"} ! GRIN2A
```
sabrinatoro, line 213436: I don't know understand why this omim source was removed. It should be kept.

sabrinatoro: I take it back: it is because it is a "included". All good

sabrinatoro (approved):
Everything looks good. I am approving and merging this PR

src/ontology/mondo-edit.obo
twhetzel: If there is no OMIM source and if the next step Nico will do is to remove `MONDO:mim2gene_medgen` source information, then in the future this gene association will not have any source information. Or is there a plan to keep this `MONDO:mim2gene_medgen` source?

twhetzel: This missing source is now part of 82 other gene associations missing sources following removal of the mim2gene source information in this [PR](https://github.com/monarch-initiative/mondo/actions/runs/17681589241/job/50256777281?pr=9558) so this comment here no longer needs to be addressed individually.

PR #9562, opened 2025-09-15

twhetzel:


twhetzel:
The changes look good to me, but this also needs someone else to officially approve the PR once the Mondo QC checks finish.

twhetzel:
The data from the mondo-ingest repo used with the pipeline was generated on Sep. 3 following the Mondo release. However, this issue with `MONDO:0800029 ` highlights there should be a data build at the end of Week 2 regardless of whether there is any alignment work done (which would have resulted in a data re-build) to account for additional Mondo changes in Week 2. 

I will generate a new data build in mondo-ingest tonight. Do you prefer we re-run the EMC pipeline after that new mondo-ingest data is available or is it too much effort to review another EMC pipeline PR again with the newer data? 

Update: This [removed line](https://github.com/monarch-initiative/mondo-ingest/pull/876/files#diff-a19e7f31129062736841fdccad2f46e42bd8959281ee0ffe7ba0c82b6d94a08cL9095) in the ordo subsets file in the new mondo-ingest data build shows `MONDO:0800029` will not be annotated as an ordo_disorder subset.

sabrinatoro:
I think it would be better to rerun the EMC after the data rebuild. We want to make sure we reduce non-synchronization as much as possible. Thanks!

twhetzel:
I agree. Once Nico reviews the [data build PR](https://github.com/monarch-initiative/mondo-ingest/pull/876), I will re-run the EMC pipeline.

PR #9564, opened 2025-09-15
Add scripts and ontology terms for disease reclassification

sabrinatoro:
Introduces Python and shell scripts to automate reclassification of human disease subclasses in the MONDO ontology. Adds new grouping terms to support genetic, anatomical, environmental, functional, and life stage related disease categories, with definitions and metadata for each.

### DIFF
--- a/complete_reclassification.py
+++ b/complete_reclassification.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+
+def reclassify_terms():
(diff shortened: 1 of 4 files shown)


twhetzel:
These files can be added under `src/scripts` and in their own folder may be helpful along with a README.

sabrinatoro:
Sure, however, this was not what I had asked it to do... :-(

PR #9571, opened 2025-09-16

twhetzel:


src/ontology/mondo-edit.obo
```diff
@@ -620078,11 +620135,59 @@ property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/
...
 name: INTU-related skeletal ciliopathy
 def: "A skeletal ciliopathy caused by a mutation in INTU gene and is characterized by facial dysmorphism, tongue nodules, developmental delay, and polydactyly. Some individuals may also present with short stature, or other variable syndromic findings." [PMID:34623732, PMID:38702430]
+subset: inferred_rare
+subset: rare
 is_a: MONDO:0005308 {source="PMID:34623732", source="PMID:38702430"} ! ciliopathy
 relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/29239 {source="PMID:34623732", source="PMID:38702430"}
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-7638-4659
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/9248" xsd:anyURI
 
+[Term]
```
sabrinatoro: @twhetzel Do these terms come from? 
It looks like they came from the test I was doing with Claude yesterday, but nothing was pushed to master (at least to the best of my knowledge). 
These should not be there.

sabrinatoro (changes requested):
@twhetzel there are new terms that should not be there. Where did they come from? 
They were created in a PR when I tested Claude. But to my knowledge, the test PR were in draft mode and were not merged.

src/ontology/mondo-edit.obo
```diff
@@ -572151,6 +572152,16 @@ property_value: curated_content_resource "https://www.malacards.org/card/dyskine
 id: MONDO:0800029
 name: obsolete interstitial lung disease 2
 comment: This term represented multiple concepts. This ID was obsoleted, and the following 2 terms were created as a replacement: MONDO:0800497 - interstitial lung disease 2 , and MONDO:0800498 - idiopathic pulmonary fibrosis
+subset: clingen {source="MONDO:CLINGEN"}
+subset: gard_rare {source="GARD:8609", source="MONDO:GARD"}
+subset: nord_rare {source="MONDO:NORD", source="NORD:1296"}
+subset: omim_susceptibility {source="OMIM:178500"}
```
sabrinatoro, line 572158: should not be here. This term has no x-ref to omim

src/ontology/mondo-edit.obo
twhetzel: To follow-up from the Slack comment, they were pushed into `master` https://github.com/monarch-initiative/mondo/commits/master/

sabrinatoro:
@trish this PR looks ok now. I have approved it and I am merging it now.

PR #9572, opened 2025-09-16
removed Claude added terms

sabrinatoro:
These terms were added from Claude directly into Mondo master, but they were never meant to be added.


sabrinatoro:
closing in favor of Revert "Claude testing - high level reclassification of human disease branch" #9573

PR #9599, opened 2025-09-23
removed incorrect gene associations

### LINKED ISSUE #9598: Remove incorrect gene associations
In [this issue](https://github.com/monarch-initiative/mondo/issues/9418), some gene annotations were kept because they were part of logical definition and needed more review time 

Here, I reviewed the terms that have logical definition including a gene association, but does not have a gene association SubClassOf annotation. (this report was created by goose).

The goal is to determine whether these genes annotations are correct. 

Summary of this work is in [this spreadsheet](https://docs.google.com/spreadsheets/d/15VEjbHen25_EAlRd2gbFxjJpmFnxWAMdYQ2vPPb4nVQ/edit?gid=0#gid=0)
Note that for the terms that were updated, I removed the definitions that were based on the gene-disease pattern (note that the evidence for these were only the gene-disease pattern). Because of a lack of time, I did not add a new definition.

sabrinatoro:
Closes #9598

Spreadsheet: https://github.com/monarch-initiative/mondo/issues/9598

Please see issue for explanation


MeeSiing (approved):
@sabrinatoro I think the PR looks good to me but I am not sure if you want to add the term tracker especially to those that had the gene association removed.

Issue #9301, opened 2025-07-10
[Revise subclass] Perceptual disorders are not all psychiatric?


kanems:
**Mondo term (ID and Label)**
"perceptual disorders" MONDO:0024417 
(and many of its children
  - Hearing disorder MONDO:0021945
  - vision disorder MONDO:0021084
  - ..... and others)

**Suggested revision and reasons**
Based on the definition of "psychiatric disorders" MONDO:0002025 (and the common use of this concept/term) I do not know if it is correct to have all Hearing and Vision disorders under this branch.  They certainly fit under Perceptual disorders, but not all perceptual disorders are psychiatric in nature/root cause. 

There is a similar question/concern about the terms under Intellectual Disability ending up under "psychiatric diseases" as well. But this is a point that could be argued one way or the other, I suspect.

<img width="634" height="351" alt="Image" src="https://github.com/user-attachments/assets/a33954dc-6e7d-4c62-aaa4-2d67bb9dd56d" />

**Your nano-attribution (ORCID)**
If you don't have an ORCID, you can sign up for one [here](https://orcid.org/)

PR #9608, opened 2025-09-26
excluded SCof Psychiatric disorder for Perceptual disorder and Intellectual disability

MeeSiing:
address #9301


src/ontology/mondo-edit.obo
```diff
@@ -559672,7 +559675,8 @@ synonym: "hallucinogen abuse with hallucinogen persisting perception disorder" N
 xref: ICD10CM:F16.183 {source="MONDO:relatedTo"}
 xref: MEDGEN:536590 {source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 xref: UMLS:C0236720 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source="MEDGEN:536590"}
-is_a: MONDO:0024417 {source="https://orcid.org/0000-0001-5208-3432"} ! perceptual disorders
+is_a: MONDO:0001423 {source="PMID:29547576"} ! drug-induced mental disorder
+relationship: excluded_subClassOf MONDO:0024417 {source="https://orcid.org/0000-0001-5208-3432", source="https://orcid.org/0000-0002-7638-4659"} ! perceptual disorders
```
sabrinatoro: I am not 100% sure this is correct.
According to the definition, it is "A perceptual disorder caused by intoxication with hallucinogen drugs, especially LSD."
Is it both a perceptual disorder AND a drug-induced mental disorder? 
If if it not a perceptual disorder, the definition should be updated.

MeeSiing:
@sabrinatoro Thank you for reviewing. Yes, it is both perceptual disorder and drug-induced mental disorder.

sabrinatoro (approved):
Great job @MeeSiing 
Approving and merging

PR #9629, opened 2025-10-02
merge benign glioma and low grade glioma

### LINKED ISSUE #4371: review benign glioma (MONDO:0000638)
!! requires a more detailed investigation to figure out what we should do!!!

MONDO:0000638
benign glioma 

the only xref for this term was DOID:0060101 (now obsoleted) 

DOID:0060101 is now an alternative ID for DOID:0080829 (low grade glioma) that is not a xref in Mondo

We might need to obsolete 'benign glioma' and add the new DOid to 'low grade glioma' (MONDO:0021637).

katiermullen:
closes #4371


sabrinatoro (approved):
Looks great. Thank you @katiermullen

PR #9642, opened 2025-10-09
Adds 440 icd11 foundation mappings from Medkorrekt

matentzn:
@sabrinatoro these mappings have been curated by @Silvia113-SM and reviewed by claude code.

Claude flagged 29 mappings as questionable so I passed them back for a second round of reviews. I think this is a good pattern:

1. Trusted curator provides mapping
2. We ask Claude for a review of the mapping
3. Where they agree, we just wave it through (and see what QC is flagging up)
4. Where they disagree, we arbitrate

How does that sound?


matentzn:
Remove Claude as the reviewer (or play with sssom:reviewer_id)

sabrinatoro:
@matentzn Shouldn't all of these xref have MONDO:equivalentTo source? 
(or other is this is not an exact match?)

matentzn:
> source: MONDO:EquivalentTo is missing

@sabrinatoro oops.

matentzn:
uiiii sorry - newwwwbieee mistake! Should be fixed now!

sabrinatoro (approved):
Looks good! Approving and merging. 
Thank you @matentzn

PR #9658, opened 2025-10-15
Update synonyms and references for littoral cell angioma

### LINKED ISSUE #9546: Request for new synonym littoral cell angioma of the spleen
### Mondo term

littoral cell angioma of the spleen

### Synonyms

littoral cell angioma

### Synonym type

exact

### Additional comments

for the term
MONDO:0023650	littoral cell angioma of the spleen

add the synonym "littoral cell angioma" 

I would also suggest, based on the frequency of usage in the literature, changing the primary name to "littoral cell angioma"

https://en.wikipedia.org/wiki/Littoral_cell_angioma
https://www.pathologyoutlines.com/topic/spleenlittoralcellangioma.html
https://www.nejm.org/doi/full/10.1056/NEJM200109063451016

matentzn:
Fixes #9546 

Made "littoral cell angioma" exact synonym. Added 'LCA' as an exact abbreviation and updated the synonym for 'littoral cell angioma' with additional sources.

@sabrinatoro I have

1. Asked Claude to analyse the issue: https://github.com/monarch-initiative/mondo/issues/9546#issuecomment-3404548861
2. Checked a few papers and referenced one
3. Found that there is an abbreviation that is commonly used and added that alongside the paper reference


sabrinatoro:
@matentzn I will approve this PR, but how about the suggestion to change the primary name to "littoral cell angioma", based on the frequency of usage in the literature?

sabrinatoro (approved):

matentzn:
> but how about the suggestion to change the primary name to "littoral cell angioma", based on the frequency of usage in the literature?

@sabrinatoro I didnt want to do that, as the naming seems to be all over the place, and I would suggest we keep the more descriptive name which is _widely used_: https://pubmed.ncbi.nlm.nih.gov/?term=littoral+cell+angioma

PR #9664, opened 2025-10-15
Updated definition

sabrinatoro:
related to #9633

The requester confirmed that these terms can be defined as "neurodevelopmental disorder"


katiermullen (approved):
@sabrinatoro Thank you for checking on this. I think this is great! I approve and will merge this PR.

PR #9678, opened 2025-10-21
Update text for susceptibility definition

MeeSiing:


src/patterns/dosdp-patterns/susceptibility_by_gene.yaml
```diff
@@ -45,7 +45,7 @@ annotations:
   - gene
 
 def:
-  text: A susceptibility or predisposition to %s in which the cause of the disease
+  text: A susceptibility or predisposition to %s in which the cause of the predisposition
```
sabrinatoro, line 48: hmmm.... I am not sure this is correct. 
I think that the cause of the **disease** is the variation in the gene, but it is a predisposition because the penetrance is not 100% (ie if you have a variation in the gene, you might not have the disease, but there is a good chance that you will). 

@katiermullen can you share your opinion on this?

katiermullen: I think that it needs to be added that this is a genetic susceptibility or predisposition because there are also environmental predispositions to diseases (e.g., exposure to asbestos and predisposition to mesothelioma).

The definition from NCI https://www.cancer.gov/publications/dictionaries/genetics-dictionary/def/hereditary-predisposition 

"An increased chance or likelihood of developing a particular disease based on the presence of one or more genetic variants and/or a family history suggestive of an increased risk of the disease. Having a hereditary predisposition does not mean an individual will develop the disease. Lifestyle and environmental factors can also affect an individual's risk of disease. Also called genetic predisposition, genetic susceptibility, and inherited predisposition."

PR #9688, opened 2025-10-24
Add a GitHub action that automatically normalised Mondo edits

matentzn:
@twhetzel can you review?

This has been discussed with @sabrinatoro.

Basically, instead of people remembering to normalise their branches locally with ODK, this action does it for them: whenever there is content on a branch that needs to be normalised, it creates a commit.

I am not concerned about the race condiation between git pull and git push - that has never happened in Mondo (what Copilot is complaining about).

### DIFF
--- a/.github/workflows/normalize-mondo-edit.yaml
+++ b/.github/workflows/normalize-mondo-edit.yaml
@@ -31,8 +31,7 @@ jobs:
 
       - name: Run normalization
         run: |
-          cd src/ontology
-          make NORM && mv NORM mondo-edit.obo
+          cd src/ontology && make NORM && mv NORM mondo-edit.obo
 
       - name: Check for changes
         id: check_changes


.github/workflows/normalize-mondo-edit.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Configure git
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Run normalization
+        run: |
+          cd src/ontology
+          make NORM && mv NORM mondo-edit.obo
```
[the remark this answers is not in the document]
matentzn: Nah, too cautious.

.github/workflows/normalize-mondo-edit.yaml
```diff
@@ -0,0 +1,56 @@
...
+jobs:
+  normalize:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.6
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
twhetzel, line 23:
```suggestion
          fetch-depth: 1
```

twhetzel: Does this work with a shallow clone?

matentzn: Good point, but no, git diff and git pull with --rebase cannoth work with a shallow clone. Unfortunately!

twhetzel (approved):

PR #9776, opened 2025-11-19
Update OMIM gene references

twhetzel:
Update OMIM gene references.


twhetzel:
I am reviewing the QC errors, but will post some questions shortly.

twhetzel:
There are a few entries that failed the QC because the source was missing due to a bug in the OMIM processing. However, there are remaining entries that failed this QC and need a source. In some cases, the OMIM entry was updated since we last processed this data and the gene association no longer meets our criteria for being disease defining and is also not in the "protected" file.

PR #9819, opened 2025-12-04
Rename MONDO:0004762 to pelvic congestion syndrome and add syns

### LINKED ISSUE #9820: Rename Taylor syndrome to pelvic congestion syndrome
A community member (Sophie DeLuca from Every Cure) has brought to my attention the confusing nature of this disease. It seems real after digging around in a few review papers, so I created #9819 accordingly (with lots of PMIDs to back me up). I did this manually, without AI etc, so..

1. Changed the primary name from Taylor syndrome to pelvic congestion syndrome
2. Added 3 or 4 synonyms from review papers
3. Added a comment about the confusing and incomplete definition of the disease
4. Added a fuzzy definition which you will probably want to delete.

matentzn:
Fixes #9820

Renamed MONDO:0004762 from 'Taylor syndrome' to 'pelvic congestion syndrome', added definition, comment, and updated synonyms and references to reflect current medical understanding and literature.


src/ontology/mondo-edit.obo
```diff
@@ -93442,10 +93442,15 @@ replaced_by: HP:0008722
 
 [Term]
 id: MONDO:0004762
-name: Taylor syndrome
-synonym: "congestion-fibrosis syndrome" EXACT [DOID:9346]
-synonym: "pelvic congestion syndrome" EXACT [DOID:9346, ICD9CM:625.5]
-synonym: "Taylor's syndrome" EXACT [DOID:9346]
+name: pelvic congestion syndrome
+def: "Pelvic congestion syndrome is a form of chronic pelvic pain caused by reflux or obstruction of the ovarian or pelvic veins, often associated with pelvic or vulvar varicosities." [PMID:38981444]
```
matentzn: Due to the known vagueness of the definition we should probably NOT include this!


```suggestion
```

src/ontology/mondo-edit.obo
```diff
@@ -93442,10 +93442,15 @@ replaced_by: HP:0008722
...
+comment: "PCS is real and studied, but its definition is vague, diagnostic criteria are not standardized, and pathophysiology remains multifactorial and incompletely understood, see for example PMID:26404449."
+synonym: "congestion-fibrosis syndrome" EXACT [DOID:9346, PMID:38981444]
+synonym: "female varicocele" EXACT [PMID:31345010, PMID:38981444]
+synonym: "PCS" EXACT ABBREVIATION [PMID:38981444]
+synonym: "pelvic congestion syndrome" EXACT [DOID:9346, ICD9CM:625.5, PMID:38981444]
+synonym: "pelvic vein incompetence" EXACT [PMID:31345010, PMID:38981444]
+synonym: "Taylor's syndrome" EXACT [DOID:9346, PMID:38981444]
```
sabrinatoro, line 93454: We also need to add the current label (ie "Taylor syndrome") as a synonym.

src/ontology/mondo-edit.obo
sabrinatoro: We do not repeat the name of the term in the definition. It should say "A form of chronic pelvic pain..."

sabrinatoro: Since the parent term is 'uterine disorder', maybe a better definition would be 
"a uterine disorder characterized by chronic pelvic pain...."

sabrinatoro (commented):
@matentzn Please see my comments in line. 
I can take care of the changes for you later today or tomorrow.

src/ontology/mondo-edit.obo
matentzn:
```suggestion
synonym: "Taylor's syndrome" EXACT [DOID:9346, PMID:38981444]
synonym: "Taylor syndrome" EXACT [PMID:9074246]
```

Was actually not so easy to find a usage of that name, it does not seem to be widely used without the apostrophe "s".

src/ontology/mondo-edit.obo
matentzn:
```suggestion
def: "A uterine disorder characterised by chronic pelvic pain, caused by reflux or obstruction of the ovarian or pelvic veins, often associated with pelvic or vulvar varicosities." [PMID:38981444]
```

Is it ok to use that PMID as "evidence" even if the definition is not exactly supported in that paper?

sabrinatoro: yes it is ok.

sabrinatoro (approved):

PR #9821, opened 2025-12-05

twhetzel:


sabrinatoro (approved):
This looks good to me. 
Question before merging: how are we handling synonyms that exact at the source, but we have determined that they are not exact, we consider them "narrow" or "broad".

It looks like the synonym x-ref was removed for these synonyms (see example below). Is this the expected functionality? If it is, then this PR can be merged, but we should discuss how to handle these cases. 
@matentzn @twhetzel 
Screenshot 2025-12-11 at 3 39 55 PM

matentzn:
> It looks like the synonym x-ref was removed for these synonyms (see example below). Is this the expected functionality? If it is, then this PR can be merged, but we should discuss how to handle these cases.

I think this is correct. We should not misrepresent the source and say "they said its broad" and we can have an independent analysis checking for scope changes (we actually have one already!). So I think this is correct behaviour.

twhetzel:
Yes, what's run here are only confirmed synonyms where the synonym scope and the synonym string match between Mondo and the source.

Issue #7693, opened 2024-05-07
review terms that are obsolete in Mondo but "orphanet disorders"


sabrinatoro:
Based on https://github.com/monarch-initiative/mondo/pull/7681#issuecomment-2099034928

Some obsolete classes with exact mappings to the external resource which are not added to the RD subset. 

| Mondo ID      | label                                                                                                                   |
|---------------|-------------------------------------------------------------------------------------------------------------------------|
| MONDO:0017049 | obsolete hypomyelination neuropathy-arthrogryposis syndrome                                                             |
| MONDO:0034979 | obsolete peripapillary staphyloma                                                                                       |
| MONDO:0029051 | obsolete autosomal recessive nail dysplasia                                                                             |
| MONDO:0035314 | obsolete congenital tricuspid valve dysplasia                                                                           |
| MONDO:0021690 | obsolete congenital left ventricular aneurysm                                                                           |
| MONDO:0018888 | obsolete congenital cornea plana                                                                                        |
| MONDO:8000032 | obsolete malformation syndrome                                                                                          |
| MONDO:0016250 | obsolete rare adenocarcinoma of the breast                                                                              |
| MONDO:0035250 | obsolete anomalous aortic origin of the right coronary artery                                                           |
| MONDO:0009930 | obsolete pulmonary arteriovenous malformation                                                                           |
| MONDO:0018344 | obsolete periodic paralysis with transient compartment-like syndrome                                                    |
| MONDO:0017173 | obsolete non-syndromic male infertility due to sperm motility disorder                                                  |
| MONDO:0018393 | obsolete male infertility with azoospermia or oligozoospermia due to single gene mutation                               |
| MONDO:0020256 | obsolete congenital trochlear nerve palsy                                                                               |
| MONDO:0032011 | obsolete biological anomaly                                                                                             |
| MONDO:0015976 | obsolete hyper-IgM syndrome without susceptibility to opportunistic infections                                          |
| MONDO:0034028 | obsolete symptomatic form of hemochromatosis type 1                                                                     |
| MONDO:0014753 | obsolete autosomal recessive optic atrophy                                                                              |
| MONDO:0019138 | obsolete bleeding diathesis due to a collagen receptor defect                                                           |
| MONDO:0016599 | obsolete autosomal dominant secondary polycythemia                                                                      |
| MONDO:0016665 | obsolete unclassified vasculitis                                                                                        |
[...]

PR #9834, opened 2025-12-18
update orphanet x-ref

sabrinatoro:
Addresses #7693

This PR is in preparation for the work done to reinstate or keep obsolete terms equivalent to orphanet "disorder" or "subtype of disorder"

Information from [this spreadsheet](https://docs.google.com/spreadsheets/d/1qysaj-uSDkVDCgjLjSo-TNpXuYGGsYQKXaNT-laCrI8/edit?gid=157848082#gid=157848082) - "Sabrina" as curator


src/ontology/mondo-edit.obo
```diff
@@ -277489,7 +277505,6 @@ xref: UMLS:C0086795 {source="MONDO:equivalentTo", source="MONDO:MEDGEN", source=
 is_a: MONDO:0001586 {source="DOID:0111390", source="NCIT:C61261", source="Orphanet:93473"} ! mucopolysaccharidosis type 1
 is_a: MONDO:0002254 {source="NCIT:C61261"} ! syndromic disease
 is_a: MONDO:0002561 {source="NCIT:C61261/inferred", source="Orphanet:93473/inferred", source="PMID:21723623"} ! lysosomal storage disease
-is_a: MONDO:0016340 ! familial restrictive cardiomyopathy
 is_a: MONDO:0800088 {source="PMID:31633310"} ! lysosomal storage disease with skeletal involvement
```
katiermullen, line 277508: @sabrinatoro does this need to have an excluded subclass annotation?

sabrinatoro: no, because we don't know where this annotation came from. 
I removed it because it didn't have a source/evidence. It is possible it might come back with some evidence.

PR #9865, opened 2026-01-09
Fix translation pipeline

matentzn:
The translation pipeline was ALWAYS WRONG (my bad). Terue informed me about this.. This is fixing it finally!


twhetzel:
I gather my assignment on this is due to legacy CODEOWNERS status?

matentzn:
@twhetzel - yes; this will be fixed here #9866 - sorry for the noise.

src/ontology/mondo.Makefile
```diff
@@ -1630,7 +1630,7 @@ TRANSLATE_PREDICATES=rdfs:label
 .PHONY: update-mondo-japanese-translation
 update-mondo-japanese-translation:
 	@echo "DOWNLOADING JAPANESE TRANSLATION IS SKIPPED"
-	wget "https://raw.githubusercontent.com/dbcls/mondo-japanese/refs/heads/main/babelon/mondo-jp.babelon.tsv" -O $@
+	wget "https://raw.githubusercontent.com/dbcls/mondo-japanese/refs/heads/main/babelon/mondo-jp.babelon.tsv" -O ../translations/mondo-jp.babelon.tsv
```
ptgolden, line 1636: Is there any particular reason to have this be a `.PHONY` target? `$(TRANSLATIONSDIR)/mondo-jp.babelon.tsv` is a dependency on [line 1648](https://github.com/monarch-initiative/mondo/blob/464e73cc4a41f63cd6644a614828b9f490edcbfb/src/ontology/mondo.Makefile#L1648)-- it seems better to me to make that file the target here, and then add a secondary `.PHONY` target as an alias.

(And in keeping with the rest of the file, `$(TRANSLATIONSDIR)` should be used here rather than hardcoding `../translations`).

ptgolden (changes requested):
Nice find! Small changes requested to fit in with the conventions of the rest of the Makefile.

src/ontology/mondo.Makefile
matentzn: The point here is that we did not want the pipeline in 1648 to actually try and download the file (it caused to many errors - so we have a separate action to download and QC the file so the rest of the release pipeline is not affected...

PR #9881, opened 2026-01-15
associate infected hydrocele with infectious disease 

### LINKED ISSUE #9502: Add parent term to MONDO:0004919 infected hydrocele
the term
MONDO:0004919	infected hydrocele

does not have a parent which is
MONDO:0005550 infectious disease

MeeSiing:
close #9502

added definition to infected hydrocele, with PMIDs as xref
added bacterial infectious disease as SCof, with PMIDs as source
added logical definition of "hydrocele and ('disease has infectious agent' some Bacteria)"
added term tracker


src/ontology/mondo-edit.obo
```diff
@@ -96459,7 +96460,12 @@ xref: MEDGEN:510235 {source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 xref: SCTID:11666007 {source="DOID:9911", source="MONDO:equivalentTo"}
 xref: UMLS:C0156300 {source="MEDGEN:510235", source="MONDO:equivalentTo", source="MONDO:MEDGEN"}
 is_a: MONDO:0004920 {source="DOID:9911"} ! hydrocele
+is_a: MONDO:0005113 {source="PMID:30029966", source="PMID:4588074"} ! bacterial infectious disease
```
matentzn: Can you explain how the classification as "infectious disease" follows from PMID:30029966?

MeeSiing: It was mentioned that "Infected hydroceles, however, are usually a surgical finding and their infectious aetiology may be reactive to a contagious genitourinary and anorectal infection. The microorganisms involved may be responsible for sexually transmitted diseases, gram-positive cocci, Enterobacteriaceae and anaerobes in cases of gangrene. Viruses (varicella-zoster), Brucella, filariasis and bacillus Calmette-Guérin are less common in case of bladder cancer treatment. Finally, emerging pathogens such as A. urinale."

matentzn: In my view: The excerpt shows that hydroceles can become infected, but that does not make hydrocele itself an infectious disease— infection is a secondary complication with its own etiology? You disagree?

MeeSiing: I think we are defining 'infected hydrocele' not hydrocele. From my discussion with Sabrina and Katie, infected hydrocele is always associated with infections, and that should align with the definition of an infectious disease.

matentzn: Yes, agreed that it is _associated with_ infectious disease but I am still not convinced it is one. I am still looking for the a sentence like "an infectious disease is defined as an X; IH corresponds to that definition according to PMID 2". Maybe I am not following what we define as infectious disease exactly?

MeeSiing: Maybe i did not understand the discussion correctly. Let me ask for Katie's review for more of her input.

katiermullen: Maybe we should distinguish these as primary and secondary as in https://www.ncbi.nlm.nih.gov/books/NBK559125/. 

I do think it is important to call out filarial (ie parasitic infection) hydrocele as a type of secondary hydrocoele. This condition is endemic to tropical and subtropical regions around the world, particularly in Asia and in areas of poor socioeconomic status. This type of chronic hydrocele due to blockage or dysfunction of the scrotal lymphatic vessels that have become clogged, scarred, or damaged by nematodes.

https://www.ncbi.nlm.nih.gov/books/NBK560776/

sabrinatoro:
- I don't think adding "disease_has_infectious_agent some Bacteria" is correct. Do we know that the infection is always bacterial? 
- I agree with Katie that according to [here](https://www.ncbi.nlm.nih.gov/books/NBK559125/), we should make the distinction between primary and secondary. But this could be done when we have more time (haha)
- In the meantime, I think that "infected hydrocele" should be defined as a 'disease has basis in feature' some 'infectious disease' (similar examples would be MONDO:0018837,  MONDO:0002812)

I can do the changes, but first I would like to make sure we agree.

katiermullen:
I agree with @sabrinatoro. Thank you for making these changes.

MeeSiing:
@sabrinatoro, thank you for looking into this. Infected hydrocele is commonly caused by bacteria but not always bacteria. 
I agree with you and thank you for making the changes.

sabrinatoro (approved):

PR #9888, opened 2026-01-20
Overhaul mapping pipeline to use sssom-java

ptgolden:
This mostly removes our use of sssom-py in the mapping pipeline in favor of using the CLI of sssom-java. Before merging, I need to change a couple things, mostly related to embedding metadata in output SSSOM files.

On my machine, the end effect is:

Before:
Elapsed time: 30:34.81
Peak memory: 6160944 kb

After:
Elapsed time:
7:13.24
Peak memory: 3572144 kb

The big time save comes from 1) not building `mondo-ingest.db` using semsql and 2) not splitting the SSSOM dataframe in sssom-py. There are also numerous other timesaves which stem from sssom-java being faster to parse files than sssom-py. There is also the benefit of just using a CLI rather than maintaining external scripts.

There is one new script added: `extract_all_labels.py`. This *may* not be necessary, but I included it to make one-to-one parity with the output of the previous pipeline. What this script does is make an n-triples file containing every URI in an OWL XML ontology that has an `rdfs:label`. A triple is emitted with the *first* `rdfs:label` that is encountered. The resulting n-triples file is passed to the `--update-from-ontology` flag in sssom-java.

Why is this even necessary, and why can't the base `mondo-ingest.owl` file be passed to that flag? Again, it is for absolute parity with the sssom-py pipeline. There were some object labels being populated there that were *not* OWL classes, but rather rdf:Description classes. (I think these were NCIT, but I don't quite recall). I would gladly delete this script and take out this logic if including those is not important. Or maybe something in the mondo-ingest pipeline needs to be fixed.

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -1181,21 +1181,43 @@ sssom:
 oaklib:
 	python3 -m pip install --upgrade pip setuptools && python3 -m pip install --upgrade --force-reinstall oaklib
 
-tmp/%.sssom.tsv: tmp/mirror-%.json
-	sssom parse tmp/mirror-$*.json --no-strict-clean-prefixes -I obographs-json -m $(METADATADIR)/mondo.sssom.config.yml -C merged -o $@
+$(TMPDIR)/mondo-extracted.sssom.tsv: $(TMPDIR)/mirror-mondo.json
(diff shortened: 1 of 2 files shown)


ptgolden:
While we may decide that we want to use `sssom-java` in the future, that will be future work. Closing in favor of #9907, which was merged last week.

PR #9898, opened 2026-01-25

matentzn:


src/ontology/mondo-edit.obo
```diff
@@ -22983,6 +22983,8 @@ property_value: IAO:0000589 "myopia (disease)" xsd:string
 id: MONDO:0001385
 name: cortical blindness
 def: "Visual impairment due to visual cortex dysfunction." [NCIT:P378]
+subset: inferred_rare
+subset: rare
```
sabrinatoro, line 22987: "cortical blindness" being inferred_rare is weird... 
It is considered rare_inferred because "visual cortex"  is considered rare as per orphanet ([orphanet:447788](https://www.orpha.net/en/disease/detail/447788)). According to my research, "visual cortex" is not rare, so this also should not be "rare".  According to [here](https://www.perkins.org/rare-diseases-and-cvi-explained/#why-consider-rare) cerebral visual impairment should be considered as it is often missed in patients with rare diseases. 
However, Orphanet considers it a rare disease. So I guess it is ok?

matentzn: I would flag this somewhere for review, but not block this PR based on this

PR #9905, opened 2026-01-27
NTR CASK-related disorder

### LINKED ISSUE #9747: FG syndrome 4 and new gene-related disease addition
Hi, I have two different requests but they are related, so I decided to  use a blank issue:

1.  I am requesting to add the following synonyms to FG syndrome 4 (MONDO:0010318):  X-linked intellectual disability (XLID) with or without nystagmus and FG syndrome caused by CASK variants (or CASK-related FG syndrome) 
2. I am requesting to add  a new term **CASK-related disorders** which will be the parent term of: 

- "syndromic X-linked intellectual disability Najm type" (synonym: Intellectual disability and microcephaly with pontine and cerebellar hypoplasia (MICPCH))
- FG syndrome 4 (the one I am asking to add the synonyms)

In addition, it may be better to use this synonym (X-linked intellectual disability (XLID) with or without nystagmus) as the main term instead of FG syndrome 4 (see GR: "An FG syndrome (FGS)-like phenotype has been suggested as a distinct CASK-related phenotype based on findings in affected males from two families [Piluso et al 2009, Dunn et al 2017]. However, with the exception of FGS1 caused by a recurrent MED12 pathogenic variant (see MED12-Related Disorders), FGS is not clearly defined and FGS4 is not discernible as phenotype. Thus, it seems more appropriate to subsume the phenotype described in these families under XLID with or without nystagmus".  

Sorry for grouping the requests in one issue.  Please let me know your thoughts, Thank you!   Gioconda Alyea

MeeSiing:
close #9747

1. created a new term for 'CASK-related disorder' to be a parent term to 'syndromic X-linked intellectual disability Najm type' and 'FG syndrome 4' MONDO:0010318 with user's ORCID and GR as reference
2. updated synonyms of MONDO:0010318 and 'syndromic X-linked intellectual disability Najm type' with user's ORCID and GR or OMIM as reference
3. added term tracker

Question: user requests to relabel MONDO:0010318 to 'X linked intellectual disability with or without nystagmus' based on [GR](https://www.ncbi.nlm.nih.gov/books/NBK169825/) (see GR: "An FG syndrome (FGS)-like phenotype has been suggested as a distinct CASK-related phenotype based on findings in affected males from two families [Piluso et al 2009, Dunn et al 2017]. However, with the exception of FGS1 caused by a recurrent MED12 pathogenic variant (see MED12-Related Disorders), FGS is not clearly defined and FGS4 is not discernible as phenotype. Thus, it seems more appropriate to subsume the phenotype described in these families under XLID with or without nystagmus".)
My understanding is that 'X linked intellectual disability with or without nystagmus' is not exactly the same as MONDO:0010318 because OMIM has 'X linked intellectual disability with or without nystagmus' as included and therefore relabel should not happen. Do we create new term for 'X linked intellectual disability with or without nystagmus'? What is the best way to approach this?

**Update:** we talked to Ada on 2/17/2026 and she will look into FG syndrome 4 and XLID. At the mean time, we create XLID as a new term (because it's an included entry) and have CASK-related intellectual disability as a parent over FG syndrome 4, XLID and syndromic X-linked intellectual disability Najm type.


src/ontology/mondo-edit.obo
```diff
@@ -622580,6 +622579,16 @@ relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-7638-4659
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/9732" xsd:string
 
+[Term]
+id: MONDO:1060192
+name: CASK-related disorder
+def: "A syndromic disorder in which the cause of the disease is a mutation in the CASK gene. It is associated with a wide phenotypic spectrum ranging from mild-to-severe intellectual disability with or without nystagmus to moderate-to-profound intellectual disability and progressive microcephaly with pontine and cerebellar hypoplasia, often associated with seizures." [MONDO:patterns/disease_series_by_gene, PMID:24278995]
```
sabrinatoro: Please add Gioconda's orcid

src/ontology/mondo-edit.obo
```diff
@@ -622580,6 +622579,16 @@ relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/
 property_value: http://purl.org/dc/terms/creator https://orcid.org/0000-0002-7638-4659
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/9732" xsd:string
 
+[Term]
+id: MONDO:1060192
+name: CASK-related disorder
```
sabrinatoro: This name is too broad. I know, we have that name "everywhere", but it is a problem. 
Can we use the name "CASK-related syndrome" instead? (based on the definition) It would be more specific. We could add this name as a synonym (exact until we have example(s) of a disease caused by CASK but not syndromic). Let's mention this to Gioconda.

katiermullen: How do you feel about the name, CASK-related intellectual disability?
https://medlineplus.gov/genetics/condition/cask-related-intellectual-disability/#references

It is not quite what the user requested, but is more descriptive.

src/ontology/mondo-edit.obo
katiermullen: Maybe "variation" instead of "mutation"

src/ontology/mondo-edit.obo
```diff
@@ -622580,6 +622579,16 @@ relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/
...
+def: "A syndromic disorder in which the cause of the disease is a mutation in the CASK gene. It is associated with a wide phenotypic spectrum ranging from mild-to-severe intellectual disability with or without nystagmus to moderate-to-profound intellectual disability and progressive microcephaly with pontine and cerebellar hypoplasia, often associated with seizures." [MONDO:patterns/disease_series_by_gene, PMID:24278995]
+is_a: MONDO:0002254 {source="PMID:24278995"} ! syndromic disease
+intersection_of: MONDO:0002254 ! syndromic disease
```
katiermullen: Can this new term be a child of X-linked syndromic intellectual disability, MONDO:0020119?

katiermullen (commented):
@MeeSiing - a few questions for you. 

Note: OMIM:300422 includes both FG Syndrome 4 and Intellectual developmental disorder, with or without nystagmus, thus I think this is indication that they are exact synonyms? I would be okay with changing the label to "intellectual developmental disorder, with or without nystagmus" if that is the name commonly used as long as @sabrinatoro agrees.

MeeSiing:
> @MeeSiing - a few questions for you.
> 
> Note: OMIM:300422 includes both FG Syndrome 4 and Intellectual developmental disorder, with or without nystagmus, thus I think this is indication that they are exact synonyms? I would be okay with changing the label to "intellectual developmental disorder, with or without nystagmus" if that is the name commonly used as long as @sabrinatoro agrees.

@katiermullen @sabrinatoro Intellectual developmental disorder, with or without nystagmus is listed as **included** under FSG4 in OMIM. Is it still an exact synonym to FGS4?

sabrinatoro:
@katiermullen and @MeeSiing 
According to [OMIM:300422](https://www.omim.org/entry/300422?search=300422&highlight=300422), "INTELLECTUAL DEVELOPMENTAL DISORDER, X-LINKED, WITH OR WITHOUT NYSTAGMUS" is an included entry, and therefore should be its own Mondo entry with the OMIM:300422 annotated as "included". 

"included" in OMIM means that it is a different though related disease.

src/ontology/mondo-edit.obo
MeeSiing: I will suggest this name to Gioconda. Thanks.

MeeSiing: I will also suggest this name to Gioconda. Thanks.

src/ontology/mondo-edit.obo
MeeSiing: added

sabrinatoro: Should the logical definition be "syndromic intellectual disability" and not "any syndrome" ?

PR #9955, opened 2026-02-18
Add clingen-uuids.yml with replacement mappings

matentzn:
This is a fake file so @setiadha-unc can develop his pipeline.


ptgolden (approved):
Great! I'll look forward to hearing what the pipeline is about.

PR #9958, opened 2026-02-19
Add ClinGen gene-disease NTR issue template

sabrinatoro:
@matentzn 
This is a draft of a template for ClinGen to use to request a new term. 
I don't know whether the semantics are 100% correct, but the information requested is

### DIFF
--- a/src/ontology/issue_templates/monogenic_ntr.md.j2
+++ b/src/ontology/issue_templates/monogenic_ntr.md.j2
@@ -13,7 +13,10 @@
      term_label        - preferred label for the new term
 
    Optional variables:
-     synonyms          - list of synonym strings
+     synonyms          - list of synonym objects with keys:
+                           value (required) - the synonym string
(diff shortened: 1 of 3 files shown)


matentzn (changes requested):
The content is correct, but the form is not - I will take this over to create the correct form.

sabrinatoro:
@matentzn can you tell me the high level explanation of what is incorrect with the form? I know it is a .yml file, and the other templates we have are markdown (I thought yml was better for ClinGen). 

I am making claude-code do the work (under supervision of course), so I can tell it to do the correct form (I have been working on the other required templates)

matentzn:
Will tell you in our meeting l8er - its confusing but we dont need a _github_ issue template, but a jinja template for an agent to _instatiate_ an issue comment.. I will show you. This here is correct if the goal was to have clingen submit the issues manually.

PR #9939, opened 2026-02-11
cystic fibrosis

sabrinatoro:
addresses #9936

This PR includes:
- files reporting the Claude analysis
- changes in the ontology made by Claude and reviewed by me (Sabrina). The summary of the changes will be added in the comment below

PR #9965, opened 2026-02-20
2602 cf

sabrinatoro:
Addresses #9939 

Note that this PR includes the changes made in mondo-edit. 
[this other PR](https://github.com/monarch-initiative/mondo/pull/9939/changes) includes files that have been created to prompt Claude-Code and created by Claude. This analysis has prompted the changes in this PR (ie the changes in the ontology), and has guided other projects and discussions.


src/ontology/mondo-edit.obo
```diff
@@ -200727,7 +200705,8 @@ property_value: curated_content_resource "https://www.malacards.org/card/polycys
...
+comment: Diagnosis of CF requires **one clinical indicator** (positive newborn screen, symptoms consistent with CF, or family history) **plus one functional/genetic indicator**:\n1. Sweat chloride concentration ≥60 mmol/L, or\n2. Identification of two CF-causing CFTR pathogenic variants, or\n3. Abnormal nasal potential difference measurement\n\nSee also CFTR-related disorders (MONDO:7770004), which involve CFTR dysfunction limited to a single organ system and do not meet CF diagnostic criteria.
 subset: gard_rare {source="GARD:6233", source="MONDO:GARD"}
 subset: nord_rare {source="MONDO:NORD", source="NORD:1026"}
 subset: omim_susceptibility {source="OMIM:219700"}
```
katiermullen, line 203235: @sabrinatoro - Just a question - do you think Mondo needs a grouping class for all channelopathies? PMID: [24578711](https://pubmed.ncbi.nlm.nih.gov/24578711/)

src/ontology/mondo-edit.obo
```diff
@@ -622866,6 +622754,40 @@ is_a: MONDO:0005451 {source="PMID:38849953"} ! eating disorder
...
+id: MONDO:7770003
+name: cystic fibrosis-related diabetes
+def: "A form of diabetes mellitus unique to individuals with cystic fibrosis (CF), arising primarily from progressive fibrotic destruction of the pancreas leading to insulin deficiency, with additional contributions from insulin resistance. CFRD affects approximately 20% of adolescents and 40-50% of adults with CF. It shares features with both type 1 and type 2 diabetes but is a distinct clinical entity with its own diagnostic criteria and management guidelines." [PMID:21115772, PMID:29515516]
+synonym: "CF-related diabetes" EXACT [PMID:21115772]
+synonym: "CFRD" EXACT ABBREVIATION [PMID:21115772]
+xref: ICD10CM:E13 {source="MONDO:equivalentTo"}
+is_a: MONDO:0005015 {source="PMID:21115772"} ! diabetes mellitus
+intersection_of: MONDO:0005015 ! diabetes mellitus
+intersection_of: disease_arises_from_feature MONDO:0009061 ! cystic fibrosis
+relationship: disease_arises_from_feature MONDO:0009061 {source="PMID:21115772"} ! cystic fibrosis
```
katiermullen, line 629289: How come not a SC of cystic fibrosis?

src/ontology/mondo-edit.obo
```diff
@@ -622866,6 +622754,40 @@ is_a: MONDO:0005451 {source="PMID:38849953"} ! eating disorder
...
+id: MONDO:7770005
+name: cystic fibrosis-related liver disease
+def: "A liver disorder that arises as a consequence of cystic fibrosis. Cystic fibrosis-related liver disease encompasses a spectrum of hepatobiliary abnormalities caused by CFTR dysfunction in cholangiocytes, including focal biliary cirrhosis, multilobular biliary cirrhosis, hepatic steatosis, and cholangiopathy. It is a significant cause of non-pulmonary morbidity and mortality in CF, affecting approximately 5-10% of CF patients." [PMID:28753176, PMID:37278811]
+synonym: "CF liver disease" EXACT [PMID:28753176]
+synonym: "CF-related liver disease" EXACT [PMID:28753176]
+synonym: "CFLD" EXACT ABBREVIATION [PMID:28753176]
+is_a: MONDO:0005154 {source="PMID:28753176"} ! liver disorder
+intersection_of: MONDO:0005154 ! liver disorder
+intersection_of: disease_arises_from_feature MONDO:0009061 ! cystic fibrosis
+relationship: disease_arises_from_feature MONDO:0009061 {source="PMID:28753176"} ! cystic fibrosis
```
katiermullen, line 629311: How come not a SC of cystic fibrosis?

src/ontology/mondo-edit.obo
sabrinatoro: We had a term (MONDO:0021016), but it was obsoleted. 
We should review this decision though

src/ontology/mondo-edit.obo
sabrinatoro: Because the disease is "diabetes" and it is a consequence of CF. Therefore, it is not a SC of cystic fibrosis. (Note that all the resources I looked at says that it is NOT a cystic fibrosis). 
I concede that a patient would need to have cystic fibrosis first to have CF-related diabetes. BUT I guess the CF could be treated while the diabetes is not.

src/ontology/mondo-edit.obo
sabrinatoro: same as above. 
All sources said this disease is a consequence of CF, but not CF.

sabrinatoro:
@katiermullen please see my answers. IF you disagree, we can discuss.

sabrinatoro:
CF-related diseases should absolutely have  SC "'has material basis in germline mutation in' some CFTR"  (the variation is what "decides" whether a patient will have CF or CF-related diseases (it is also in the definition) 


I went back and forth about whether the diabetes and liver disease from CF should or should not get the gene association. And here is where I ended: 
The cause of these diseases is CF, which itself is caused by a variation in CFTR. So the variation in the gene is "one hop" above. Therefore, I think it is better to not add it to these terms, though the relation will be indirect. 
Please let me know if you disagree. 
Thanks!

PR #9988, opened 2026-02-27
Update RD subset

matentzn:
@sabrinatoro note this commit I need to manually make:

- https://github.com/monarch-initiative/mondo/commit/2f5ae57fc773a5066654c3252088f0ec237409fa


sabrinatoro:
@matentzn 2 questions: 
1) why did you remove the synonym? what is conflicting with another term?

2) All the new Orphanet terms are shown as "inferred rare". Probably because the pipeline adding the "disorder/subtype of disorder" have run (I thought it was part of this, but not)
Do you think it is ok for this upcoming release? 
Screenshot 2026-02-27 at 7 12 35 AM

PR #9989, opened 2026-02-27
Update OMIM gene references

matentzn:
Update OMIM gene references.


sabrinatoro (approved):
@matentzn as mentioned in the other commit, there were 8 omim sources that were removed, but according to the omim website, they are still correct and should have been retained. We should review why they were removed, but for now, I re-added them manually. 

I am approving this PR, and will merge this issue after QCs have passed.

PR #9991, opened 2026-02-28
Add NCIT RD list

matentzn:
Everything added by NCIT rare is just confirmatory- so safe.

A single class has been added to RD list _independent_ of the NCIT rare list:

[Term]
id: MONDO:1060202
name: toxic thyroid adenoma

Through inference that must have been a new thing picked up from the main branch.


sabrinatoro (commented):
@matentzn is this PR adding ncit-rare to the "mondo rare disease subset"?

Issue #9190, opened 2025-06-12
Mondo high-level reclassification proposal


sabrinatoro:
Jonathan Berg suggested a main change in the high-level classification (see attached document). 
In summary, there would be 3 major classifications under human disease 
- disease of body system or component
- disease of developmental or physiological process
- disease of etiologic mechanism

This new reclassification would be more intuitive to users. 

[JSB thoughts about Mondo.docx](https://github.com/user-attachments/files/20699654/JSB.thoughts.about.Mondo.docx)

sabrinatoro:
I tried to rely on patterns and on the reasoner to create this classification. 

TLTR: 
- the classification created by relying on the reasoner is not user-intuitive (there are a lot of additional children, as expected from the reasoner). 
- some patterns are missing (e.g. for etiology) or are not consistent.

One of the test is in [PR 9189 ](https://github.com/monarch-initiative/mondo/pull/9189)
See questions and discussion points in separate comments (note all “test terms” start with AAA)

sabrinatoro:
## Issue 1: lack of “common” pattern allowing grouping via inference
The suggested term “disease of etiologic mechanism” would group diseases based on their etiology. Each suggested subclasses follow a different pattern as shown in the table below
<google-sheets-html-origin><style type="text/css"></style>
grouping term | current pattern(s)
-- | --
disease of etiologic mechanism |  
--genetic or genomic mechanism |  
----chromosomal disorder (581) | disease arises from alteration in structure' some chromosome
----hereditary disease (11,467) | has characteristic' some inheritedhas material basis in germline mutation in' some GENE
--primarily extrinsic mechanism |  
----iatrogenic disease (4) | has characteristic' some iatrogenic
----infectious disease (1,073) | disease has infectious agent' some NCBITaxon
----disease due to injury or trauma |  
----disease due to exposure |  
----------toxin |  
----------poisoning (70) | realized in re'sponse to stimulus' some CHEBI
----------radiation-induced disorder (20) | realized in response to' some 'exposure to electromagnetic radiation'
----nutritional disorder (110) |  
----occupational disorder (6) | realized in response to' some 'exposure to occupation'




In https://github.com/monarch-initiative/mondo/pull/9189, I asserted the classes (see ‘AAAofknown etiology' term)
Even though most of the  extrinsic mechanism terms (except for infectious diseases) could be defined using ECTO), I cannot think of a pattern that we currently use that could work for all these terms. 

That worked ok to assert them though

[...]

PR #10004, opened 2026-03-05
Restructure top-level classification of human disease (MONDO:0700096)

sabrinatoro:
Addresses #9190
changes were supported by Claude-code, but supervised and reviewed by Sabrina.

Note that previous version of this restructure was based on axioms. However, using the reasoner, many specific diseases would be found directly under the main grouping class, defeating the purpose of this work

Summary of Claude:

## Overview

This PR introduces a new top-level organizational structure for the direct and near-direct children of **MONDO:0700096 (human disease)**, creating four orthogonal classification axes. These axes are **not mutually exclusive** — a disease can appear in multiple branches via multiple inheritance. All existing `is_a` relationships were preserved; only new parents were added.

## Motivation

The direct children of `human disease` were a heterogeneous mix of body-system terms, etiological terms, and specific conditions. This restructuring brings coherent organization by introducing grouping terms that reflect distinct classification perspectives: anatomy, process, etiology, and molecular mechanism.

## New grouping terms created

All new terms include a comment: *"This term is intended solely for classification purposes and should not be used for direct data annotation."*

| ID | Name | Parent |
|---|---|---|
| MONDO:7770006 | disease of body system or component | MONDO:0700096 (human disease) | |MONDO:7770007 | disease of developmental or physiological process | MONDO:0700096 (human disease) | 
| MONDO:7770008 | disease of etiologic mechanism | MONDO:0700096 (human disease) |
| MONDO:7770009 | disease of genetic or genomic mechanism | MONDO:7770008 (disease of etiologic mechanism) | 
| MONDO:7770010 | disease of primarily extrinsic mechanism | MONDO:7770008 (disease of etiologic mechanism) | 
| MONDO:7770011 | disease of molecular mechanism | MONDO:0700096 (human disease) |

### Under disease of body system or component (MONDO:7770006) — 20 terms

| ID | Name |
|---|---|
| MONDO:0002022 | disorder of orbital region |
| MONDO:0002051 | integumentary system disorder |
| MONDO:0002081 | musculoskeletal system disorder | | MONDO:0002118 | urinary system disorder |
| MONDO:0002254 | syndromic disease |
| MONDO:0002409 | auditory system disorder |
| MONDO:0002657 | breast disorder |
| MONDO:0003900 | connective tissue disorder |
| MONDO:0004335 | digestive system disorder |
| MONDO:0004995 | cardiovascular disorder |
| MONDO:0005039 | reproductive system disorder |
| MONDO:0005046 | immune system disorder |
| MONDO:0005071 | nervous system disorder |
| MONDO:0005087 | respiratory system disorder |
| MONDO:0005151 | endocrine system disorder |
| MONDO:0005570 | hematologic disorder |
| MONDO:0006858 | mouth disorder |
| MONDO:0024458 | disorder of visual system |
| MONDO:0024623 | otorhinolaryngologic disease |
| MONDO:0044991 | upper digestive tract disorder |

### Under disease of developmental or physiological process (MONDO:7770007) — 10 terms

| ID | Name |
|---|---|
| MONDO:0002025 | psychiatric disorder |
| MONDO:0005066 | metabolic disease |
| MONDO:0019303 | premature aging syndrome |
| MONDO:0021147 | disorder of development or morphogenesis | | MONDO:0021166 | inflammatory disease |
| MONDO:0024322 | disorder of glycosylation |
| MONDO:0044970 | mitochondrial disease |
| MONDO:0100081 | sleep disorder |
| MONDO:0100086 | perinatal disease |
| MONDO:0700003 | obstetric disorder |

### Under disease of etiologic mechanism (MONDO:7770008) — 2 terms (direct children, excluding sub-groups)

| ID | Name |
|---|---|
| MONDO:0045024 | cancer or benign tumor |
| MONDO:0700007 | idiopathic disease |

### Under disease of genetic or genomic mechanism (MONDO:7770009) — 2 terms

| ID | Name |
|---|---|
| MONDO:0003847 | hereditary disease |
| MONDO:0019040 | chromosomal disorder |

### Under disease of primarily extrinsic mechanism (MONDO:7770010) — 8 terms

| ID | Name |
|---|---|
| MONDO:0005137 | nutritional disorder |
| MONDO:0005550 | infectious disease |
| MONDO:0021669 | post-infectious disorder |
| MONDO:0029000 | poisoning |
| MONDO:0043459 | radiation-induced disorder |
| MONDO:0043543 | iatrogenic disease |
| MONDO:0100366 | occupational disorder |
| MONDO:0700220 | disease related to transplantation |

### Under disease of molecular mechanism (MONDO:7770011) — 6 terms

| ID | Name |
|---|---|
| MONDO:0000510 | synucleinopathy |
| MONDO:0005308 | ciliopathy |
| MONDO:0005574 | tauopathy |
| MONDO:0019119 | muscular channelopathy |
| MONDO:0021060 | RASopathy |
| MONDO:0021179 | proteostasis deficiencies |

## Terms not reclassified (deferred)

- **MONDO:0043839 (ulcer disease)** — pathological finding that doesn't fit any of the four axes cleanly; left as direct child of human disease for future curator discussion.
- **MONDO:0020573 (inherited disease susceptibility)** — intentionally kept in its own branch, not reclassified under etiologic mechanism.
- Other special cases (acute disease, breast fibrocystic disease) deferred for future PRs.

## Key design decisions

1. **Only additive changes**: No existing `is_a` parents were removed. New grouping parents were added alongside existing relationships.
2. **Psychiatric disorder, sleep disorder, and obstetric disorder** were classified under *developmental or physiological process* rather than *body system*, as they are defined by behavioural/functional/temporal criteria rather than anatomy.
3. **Disorder of visual system** (MONDO:0024458) was classified under *body system* — note that the original plan referenced the wrong ID (MONDO:0004207, which is pulmonary artery leiomyosarcoma); this was caught and corrected.
4. **Mitochondrial disease** was classified using MONDO:0044970 (the broader grouping term), not MONDO:0004069 (inborn mitochondrial metabolism disorder, which is a narrower inborn-errors subclass).

## Validation

- OBO normalization (`make NORM`) passed
- ROBOT syntax check (`robot convert`) passed with no errors


katiermullen:
- Can mouth disorder and upper digestive disorder be children of digestive disorder?
- Can disorder of visual system and disorder of orbital region be children of eye disorder or something similar?
- otorhinolaryngologic disease is a medical specialty, but I am not sure if I would consider it a body system. Terms in this branch include respiratory disorders, auditory/ear disorders, and disorders of development or morphogenesis
- Disease of molecular mechanism is a bit unclear to me. Don't all diseases have a molecular mechanism? Same with etiological mechanism. I would consider removing the latter and then classifying cancer as a new meta category called disease of neoplastic transformation or something of the sort?
- idiopathic disease could be classified under a meta category of disease of unknown etiology.

Just some initial thoughts. Thank you for all of your work on this. I really like it!

sabrinatoro:
>Can mouth disorder and upper digestive disorder be children of digestive disorder?

I don't think "mouth disorder" is, just because there are part of the mouth that are not part of the digestive system (eg jaw)
"upper digestive disorder" should. I don't know why it hasn't (probably uberon problem?). I will add this. 

> Can disorder of visual system and disorder of orbital region be children of eye disorder or something similar?

Good question. I think they are related but not the same (ie we might want a different class to group both): the visual system includes more than the eye, so it cannot be a child of eye (otherwise the visual cortex would also be in this branch). 
"Eye" includes other part that are not part of the visual system (eg eyelid). So eye (as a whole) cannot be part of the visual system. 
I agree that we should somehow bring these together, but this problem has been discussed (and ignored) for a long time. We should bite the bullet in a later PR. 

> otorhinolaryngologic disease is a medical specialty, but I am not sure if I would consider it a body system. Terms in this branch include respiratory disorders, auditory/ear disorders, and disorders of development or morphogenesis

I agree. Though the exact synonym is "ear, nose and throat disorder", so from that perspective, it is a grouping based on anatomical system. I think we need to review this class in a follow-up PR

> Disease of molecular mechanism is a bit unclear to me. Don't all diseases have a molecular mechanism? Same with etiological mechanism. I would consider removing the latter and then classifying cancer as a new meta category called disease of neoplastic transformation or something of the sort?

Yes, all diseases have mechanism, an etiology, or an anatomical system. Here, these grouping classes are defined as "A grouping class for human diseases classified by their underlying molecular or pathophysiological mechanism". In other words, these classes are grouping diseases based on how they are generally defined. 
Eg "metabolic disorder" is defined by the fact that a process -metabolism in this case- is affected; "channelopathy" is defined by the fact that channel process is disrupted, etc 

> idiopathic disease could be classified under a meta category of disease of unknown etiology.

True, but I would argue that idiopathic disease = disease of unknown etiology (also, I would argue whether we need this class in the first place). 
It is currently under "disease of etiologic mechanism" because it is defined based on the mechanism underlying the disease (ie an unknown mechanism)


@katiermullen please let me know if you agree/disagree.

sabrinatoro:
From discussion on slack: 

"nutritional disease" can be caused by both intrinsic and extrinsic causes. Therefore I have removed it from the 'disease of primarily extrinsic mechanism' metaclass

bpow:
Regarding @cmungall's comment about #3661, maybe this is a case of "everything old is new again", but I think is probably more along the lines of "was too hard to maintain before, but newer tools now make this feasible".

I'm in general in favor of this restructuring. One thought about the names of the new grouping terms below human disease. Instead of things like "disease *of* etiologic mechanism", the prior obsoleted terms would sometimes be "disease *by* etiologic mechanism" or "disease *with* etiologic mechanism'. The last might be weird for having idiopathic underneath it (although with idiopathic diseases there is often assumed to be a yet-to-be-discovered mechanism).

I get that "disease by etiologic mechanism" or "disease by genetic or genomic mechanism" feels different from a unitary concept that one could refer to/reason on in many ways, but I do think that such a name would convey in part that these terms are "intended solely for classification purposes and should not be used for direct data annotation."

sabrinatoro:
@jonathan-s-berg Thank you for your patience! 
Please see the suggested high-level reorganization of Mondo [here](https://ols.monarchinitiative.org/ontologies/mondo_reorder/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_0700096?lang=en). Please note that the reorganization shown is only for the highest level in the ontology. 
The changes you see in this link will not be published until you have had a chance to review, share your feedback, and potential updates are made. 

You can add your feedback in a comment here, or in [the original GitHub issue  where I am also tagging you for convenience](https://github.com/monarch-initiative/mondo/issues/9190). 
Thank you!

jonathan-s-berg:
Hi all and @sabrinatoro -

This looks great.  I think it will be very helpful to users and the organization is more consonant with the way that pathophysiology is typically defined (causes, pathogenic mechanisms, and phenotypic effects in organ systems). 

### Regarding naming and definitions:

I agree with comments by @bpow above that naming these categories is important but difficult...  this proposed trifurcation is really trying to align with modern pathophysiology and clinical medicine, where etiology, mechanism, and phenotype are all intertwined, which makes it sound a little strange to say "disease of ..." or "disease by ..."  etc.  

Could we describe them like this:

- Causes of human disease / Definition: "A grouping class for human diseases classified by their etiologic mechanism, encompassing primarily intrinsic (genetic, genomic), primarily extrinsic (e.g., infections, toxins, radiation), and complex multifactorial (e.g. nutritional) causes."
- Pathogenic mechanisms of human disease / Definition: "A grouping class for human diseases classified by the molecular, biochemical, cellular, anatomic, developmental or physiological process that is disrupted."  _Note that I broadened this to reflect that pathophysiological mechanisms can happen at all levels of biology ... and this implies that the current subset of "disease of molecular mechanism" would be moved underneath this category._
- Body systems or components impacted by human disease / Definition: "A grouping class for human diseases classified by the body system or anatomical component primarily affected."

### Regarding specific items:

My suggestion for "ulcer disease" would be to place it underneath the "disease of developmental or physiological process" branch.  This is reflected in the definition of this entity ("A lesion on the surface of the skin or a mucous surface, produced by the _sloughing of inflammatory necrotic tissue_.").  It is essentially a process by which the normal epithelium is disrupted and can have a number of different etiologies.

As indicated above, I suggest putting "disease of molecular mechanism" within a combined category that includes both molecular and developmental and physiologic mechanisms, given that it is just describing some very specific subgroups of pathophysiology.

Also, noted previously, I think that "acute disease" should NOT stand alone at this level of the hierarchy. Some of those terms are really just descriptive of the timing between **cause** and **effect** (e.g., acute radiation syndrome versus the other radiation-related terms).  Basically an adjective or modifier and not a fundamental quality that differentiates "human disease."  All diseases have a natural process in which some symptoms occur early ("acute") and then either resolve (to normal or to death) or turn into longer term symptoms ("chronic").  People have also used the word "acute" differently in the way they describe diseases -- therefore, computational assignment to this class is susceptible to the vagaries of how different humans have historically named things.  Randomly checking the list, it looks like a lot of them (perhaps all?) are already appropriately represented in other parts of the ontology.  It would be a lot easier to check computationally to see if any of these entities would no longer be represented if this category were removed, and we could easily place any such outliers manually.

### Regarding other improvements

Assuming that you are just looking for agreement with the top-level category (human disease and the immediate daughter branches) I will hold off on making more specific recommendations for changes within each branch -- though there will be a need to revisit some of them and really think about their definitions, such as how to organize the etiologies, what constitutes a pathophysiologic mechanism, and what constitutes a body system or component.

We may also want to look at the levels above "human disease" especially "disease susceptibility" which has 777 items, with just one subcategory of "inherited disease susceptibility" having 776 items.  Some of these are monogenic disease entities (e.g. the hereditary cancer predisposition syndromes) mixed in with common/complex/multifactorial disease susceptibility -- probably due to challenges of nomenclature and historical happenstance, such as the way that OMIM categorized some of the initial GWAS literature (but did not continue systematically recording all of the GWAS hits once it became so overwhelming).  My suggestion will be to move the monogenic conditions into the appropriate location within the intrinsic/genetic/genomic etiology category, and to move polygenic risk into a "complex multifactorial etiology" category.

sabrinatoro:
@bpow and @jonathan-s-berg. 
Thank you for the comments and feedback. I have made the following updates based on your feedback (they are viewable here, not in the ols version I shared - I can ask for the OLS version to be updated if you want to take a look before we push the changes)
[ ] **Rename the high-level grouping classes:** I used "by", for example: disease **_by_** etiologic mechanism (note that I didn’t use “disease with etiologic mechanism” to avoid confusion that some diseases with etiologic mechanisms will be found in other branches. We can review this decision later
[ ] **Update definition for 'disease by etiologic mechanism'** to "A grouping class for human diseases classified by their etiologic mechanism, encompassing primarily intrinsic (genetic, genomic), primarily extrinsic (e.g., infections, toxins, radiation), and complex multifactorial (e.g. nutritional) causes."
[ ] **Update definition for 'disease by body system or component'** to "A grouping class for human diseases classified by the body system or anatomical component primarily affected."
[ ] **Update definition for 'disease by developmental or physiological process'** to "A grouping class for human diseases classified by the molecular, biochemical, cellular, anatomic, developmental or physiological process that is disrupted.
[ ] **Reclassified "disease of molecular mechanism"** would be moved underneath Pathogenic mechanisms (ie 'disease by developmental or physiological process')
[ ] **Reclassified "ulcer disease"** would be to place it underneath the "disease of developmental or physiological process" branch. 





I created the following GH issues, to be followed-up : 
- [Obsolete: ‘acute disease’](https://github.com/monarch-initiative/mondo/issues/10052) - we need to go through the obsoletion process before we obsolete this term (note that I agree completely about the non-usefulness of this term. Another term that we might also consider obsoleting is "idiopathic diseases")
- [Review disease susceptibility terms](https://github.com/monarch-initiative/mondo/issues/10053)

Please let me know if you have any other concerns before I publish this update

cmungall:
I'm placing my original analysis here, this was for the original PR, some things may have changed. TL;DR - the local changes with this PR are great and much more intuitive, but when we look at the global implications there are numerous very odd classifications. Note this is not reflective of the logic of the proposed new classifications, it is reflective of the legacy of classifications at deeper levels in mondo.

## Summary

The proposed top-level restructuring introduces four classification axes under `human disease`. The axes are sound as an organizational principle, but encoding them as `is_a` (subClassOf) axioms creates a metaclass conflation problem that produces false inferences for a significant number of diseases. This is compounded by the number of classification issues at the mid-level grouping class level, which is arguably where we should be spending our energies, using agents.

Funamentally this proposal while less confusing for users coming top-down in a drill-down way into Mondo could fundamentally confuse them further when they look at leaf classes.

## The problem

Consider the assignment of `nutritional disorder` under `disease of primarily extrinsic mechanism`. As a *class-level characterization*, this is reasonable: `nutritional disorder` exists as a grouping concept that is prototypically about extrinsic causes (dietary deficiency, malnutrition, etc.). If someone asked "what kind of category is nutritional disorder?", the answer "it's primarily an extrinsic-mechanism grouping" would be fair.

But `is_a` in OWL/OBO is subClassOf -- a universal quantification over instances. The assertion:

```
nutritional disorder is_a disease of primarily extrinsic mechanism
```

means: *every individual that is a nutritional disorder is also a disease of primarily extrinsic mechanism*. By transitivity, this propagates to all subclasses.

In practice, many subclasses of `nutritional disorder` are monogenic hereditary conditions:

```
runoak -i sqlite:obo:mondo info .sub 'nutritional disorder' .and .sub 'hereditary disease'
```

This returns **58 diseases**, including:

- Vitamin D-dependent rickets types 1 and 2 (genetic defects in vitamin D metabolism)
- Inherited obesity syndromes (leptin deficiency, MC4R deficiency, POMC deficiency)
- Inborn disorders of cobalamin metabolism and transport
- Hereditary intrinsic factor deficiency

Under the proposed restructuring, all of these would be simultaneously classified as both `disease of primarily extrinsic mechanism` (via nutritional disorder) and `disease of genetic or genomic mechanism` (via hereditary disease). Vitamin D-dependent rickets type 1A is a monogenic autosomal recessive disorder -- calling it a disease of "primarily extrinsic mechanism" is incorrect.

A converse problem exists with the anatomic/process grouping. It's impossible to over-classify here (EVERY disease is process/anatomy related). But *under* classification here will be really confusing to users. Many classes are NOT found under these groupings in random and confusing ways. E.g. 2k of cancers ARE under here but 86 are NOT.

## The metaclass distinction

The root cause is a conflation of two different kinds of statement:

1. **Instance-of (class-level metadata):** "The *class* `nutritional disorder` is an instance of the metacategory 'extrinsic-mechanism grouping'." This is a statement about the class itself -- its intensional character, its prototypical members, the reason it exists as a category. It does not propagate to subclasses.

2. **SubClassOf (logical subsumption):** "Every *instance* of `nutritional disorder` is also an instance of `disease of primarily extrinsic mechanism`." This is a universal quantification that propagates transitively to all subclasses and their instances.

The four proposed axes are fundamentally a *library classification* of classes -- they organize the top-level grouping concepts by the perspective they represent (anatomy, process, etiology, molecular mechanism). This is an instance-of relationship between a class and a metaclass. But OBO format only gives us `is_a`, which is subClassOf, and that changes the semantics entirely.

When the instance-of becomes a subClassOf, the organizational intent leaks into the logical hierarchy and produces false entailments if the proposed new groupings are treated like other classes.

The proposed structure is not necessarily wrong if we are clear that the metaclasses should not be treated as normal classes. In the original Mondo these had labels like "disease *BY* extrinsic mechanism" to signify to the user that these are organizational units rather than assertions of causation. We could do a similar thing here or make it even clearer, e.g. "classification based on ..."

Alternatively we could bite the bullet and go through classes like `nutritional disease` and make them strictly about the external factor. This would be a good thing to do anyway! Arguably it should precede and be informed by the top level proposal.

## Empirical analysis of the extrinsic-mechanism axis

To quantify the problem, I checked all 8 terms assigned to `disease of primarily extrinsic mechanism` for overlap with `hereditary disease` using:

```
runoak -i sqlite:obo:mondo info .sub '<TERM>' .and .sub 'hereditary disease'
```

### Results

| Term | Overlap with hereditary disease | Key examples |
|---|---|---|
| nutritional disorder | **58** | vitamin D-dependent rickets (monogenic), inherited obesity (leptin/MC4R/POMC deficiency), inborn cobalamin disorders |
| infectious disease | **14** | inherited prion diseases: fatal familial insomnia (AD, PRNP), Gerstmann-Straussler-Scheinker (AD, PRNP), inherited CJD |
| post-infectious disorder | **13** | familial acute necrotizing encephalopathy (RANBP2 mutations), Burkitt/Hodgkin lymphoma via immune/viral pathway genes |
| radiation-induced disorder | **9** | photosensitive trichothiodystrophy (ERCC2/ERCC3 DNA repair mutations), photosensitive epilepsy |
| poisoning | **0** | |
| iatrogenic disease | **0** | |
| occupational disorder | **0** | |
| disease related to transplantation | **0** | |

### Pattern

The 8 terms divide cleanly into two groups:

**No leakage (4 terms):** poisoning, iatrogenic disease, occupational disorder, disease related to transplantation. These terms genuinely denote *extrinsic causation* -- the class name implies the cause is external, and this holds for all subclasses.

**Significant leakage (4 terms):** nutritional disorder, infectious disease, post-infectious disorder, radiation-induced disorder. These terms describe *what kind of insult the body fails to handle* rather than *what causes the failure*. Genetic subclasses exist where the underlying defect is endogenous:

- "Nutritional disorder" includes genetic diseases that *disrupt nutritional pathways* (not caused by diet)
- "Infectious disease" includes inherited prion diseases where *the protein misfolds due to a germline mutation* (not caused by extrinsic infection)
- "Radiation-induced disorder" includes diseases where *UV triggers symptoms due to a DNA repair defect* (not caused by abnormal radiation exposure)

In each case, the class name is ambiguous between "caused by X" and "involving the process of X." The proposed `is_a` axiom assumes the former reading, but the existing subclass hierarchy reflects the latter.

## Scope of the broader problem

The extrinsic-mechanism axis is the most severely affected, but the issue applies in principle to any axis where the organizing principle (what the class is *about*) differs from properties that hold universally across all subclasses:

| Term | Assigned axis | But has subclasses that are... |
|---|---|---|
| nutritional disorder | extrinsic mechanism | hereditary (58 terms) |
| infectious disease | extrinsic mechanism | hereditary (14 terms) |
| post-infectious disorder | extrinsic mechanism | hereditary (13 terms) |
| radiation-induced disorder | extrinsic mechanism | hereditary (9 terms) |
| inflammatory disease | developmental/physiological process | sometimes caused by extrinsic agents |
| metabolic disease | developmental/physiological process | sometimes acquired/extrinsic |

The body-system axis is less affected, because "affects body system X" is generally true of subclasses if it's true of the parent. The process and molecular-mechanism axes are also relatively safe. The etiological axis is where the problem is most severe, precisely because etiology varies across the specificity spectrum of a disease hierarchy.

## Possible approaches

### 1. Use annotations instead of is_a

Tag the direct children of `human disease` with their axis membership using a non-logical property (annotation, subset, or shorthand relation that is not transitive). This preserves the organizational value for browsing and querying without creating false logical entailments.

```
property_value: mondo_classification_axis "body_system" xsd:string
```

Pro: Semantically honest. No false inferences.
Con: Less visible in standard hierarchy browsers. Requires tooling support.
Con: Arguably this is the job of the DP system, which we have historically failed to take full advantage of

### 2. Restrict to the body-system and process axes only

The body-system and process axes are less prone to this problem because they classify by *what is affected* rather than *why*. The etiological and molecular-mechanism axes could be deferred or handled differently.

Pro: Gets the most value with the least risk.
Con: Loses the etiological organization, which has genuine navigational value.

However, these metaclasses may come with their own issues. While it's impossible to *over-classify* to these (because EVERY disease affects anatomy and processes), the *converse* under-classification may be confusing to users.

For example, 86 cancers do NOT fall under `disease of developmental or physiological process`, and >2k DO fall under this. This is pretty confusing.

### 3. Accept the leakage with documentation

Treat the axes as "primarily about" rather than definitional, and document that subclass membership in these groupings is approximate. This might involve clearly signifying to the user that these are organizational units (metaclasses) and should not be read in the same way as other classes.

Pro: Simple to implement.
Con: Undermines the reliability of queries against the hierarchy. Users will reasonably expect that `is_a` means what it means.

### 4. Refine the axis assignments

Move terms like `nutritional disorder` from the etiological axis to the process axis (it's really about *what process is disrupted* -- nutrition -- not about *what causes it*). This reduces but doesn't eliminate the problem.

Pro: More accurate for specific cases.
Con: Doesn't solve the fundamental metaclass issue; just shifts which terms are misclassified.

### 5. Fully commit to the class interpretation and refactor

In this scenario we remove the hedge "primarily" from the grouping label and commit to the label. This means that if 'nutritional disorder' goes under 'extrinsic disorder' then we relabel it as 'disorder caused by nutritional factors' or similar, define it this way, and go through all subclasses (with the help of agents) and ruthlessly remove edges that don't fit this definition.

Arguably we should be focusing on this in the first place. The looseness of many intermediate groupings is a general source of confusion and maintenance headaches

Pro: results in a simpler, clearer, less lattice-like structure for users
Cons: a lot of work if done without the assistance of agents

## Recommendation

At minimum, the etiological axis assignments should be reviewed with this propagation problem in mind. A query like the one above should be run for each proposed assignment to quantify the leakage. The body-system axis is likely fine; the etiological axis needs the most scrutiny.

Longer term, this is a well-known problem in ontologies that mix organizational/navigational structure with logical/inferential structure (ICD has similar issues). The Mondo community may want to establish a convention for how to represent class-level metadata that should not propagate -- whether via annotations, subsets, or a dedicated relation with appropriate semantics.

Arguably this is attempting to overlay a UI/navigation system onto an OWL-based polyhierachy, which is a fundamental impedance mismatch.

It may be strategically informative to look at the dismech classification approach. It maps to Mondo for lattice-like classification, but also offers alternative means in that it takes established nosologies and offers these as classification choices individually rather than trying to merge into a single OWL lattice.

PR #10018, opened 2026-03-09
Add Orphanet to Xref must have source QC

matentzn:
This is just checking that all Orphanet xrefs must have SOME xref. This is arguable even weaker than what we _should_ be doing - checking that all Orphanet xrefs must have

1. A MONDO:equivalentTo etc precision qualifier
2. An orcid that is accountable for the review


ptgolden:
Is there an issue for that stronger form of validation?

matentzn:
@sabrinatoro here it is: merge it while its still hot.

I removed xrefs to core mapping namespaces that lack a recognized mapping
predicate annotation (equivalentTo, relatedTo, mondoIsNarrowerThanSource,
mondoIsBroaderThanSource, obsoleteEquivalent, etc.). These untyped xrefs
carry no semantic meaning and mislead consumers who treat all xrefs as
equivalent mappings.

Breakdown of removed xrefs:
- Orphanet: 3,151 (mostly provenance-only, e.g. source="OMIM:nnn")
- DOID: 162 (all tagged MONDO:obsolete, not a mapping predicate)
- OMIM: 4 (all tagged MONDO:obsolete)
- OMIMPS: 1 (tagged MONDO:nonExistent)

Also fixes a malformed annotation on MONDO:0035819 where
source="MONDO:equivalentTo\"" had a stray escaped quote; corrected
to source="MONDO:equivalentTo" and kept.

All correctly typed xrefs are preserved. The new QC SPARQL check
(qc-xref-without-precision.sparql) now covers Orphanet in addition
to the previously checked OMIM/OMIMPS/DOID/OMIA prefixes.

See https://github.com/monarch-initiative/mondo/issues/10086, @cmungall agrees but wants us to go further than that, which I will do in a subsequent PR.

PR #10047, opened 2026-03-20
Rename the mondo_rare subset to mondo_curated_rare

matentzn:


sabrinatoro:
@matentzn 
This looks good. PR approved and merged. 
Thank you for doing this so quickly !

PR #10120, opened 2026-04-03
Update normalize-mondo-edit.yaml

matentzn:
@ptgolden - I hate this kind of stuff. 

@sabrinatoro noticed a bug in the way the normalise workflow works.

1. PR created
2. QC started
3. Normalise runs and commits
4. QC interruped

I now made QC passing required but that is no fun unless the normalise action works correctly:

1. PR created
2. QC started
3. Normalise runs and commits
4. Old QC interruped
5. New QC (from inside the normalise action) runs

Please review carefully I am not sure at all about this. After networks and servers my third most behated subject is concurrency!

### DIFF
--- a/.github/workflows/normalize-mondo-edit.yaml
+++ b/.github/workflows/normalize-mondo-edit.yaml
@@ -1,26 +1,34 @@
 name: Normalize mondo-edit.obo
 
 on:
-  # Trigger on pushes that modify mondo-edit.obo
   push:
     paths:
       - 'src/ontology/mondo-edit.obo'
     branches:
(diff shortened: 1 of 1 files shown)


ptgolden (changes requested):
Per our call today-- while this works, it seems simpler to just make normalization part of the QC workflow so that we don't have to worry about coordination between actions.

ptgolden:
> Per our call today-- while this works, it seems simpler to just make normalization part of the QC workflow so that we don't have to worry about coordination between actions.

Done in #10178. If that's approved, I'll close this issue.

PR #10148, opened 2026-04-13
change preferred name of MCD to "minimal change disease"

matentzn:
See also: https://github.com/monarch-initiative/dismech/pull/1245
https://en.wikipedia.org/wiki/Minimal_change_disease


src/ontology/mondo-edit.obo
```diff
@@ -149430,7 +149430,7 @@ property_value: curated_content_resource "https://www.malacards.org/card/lip_can
 
 [Term]
 id: MONDO:0006835
-name: lipoid nephrosis
```
sabrinatoro, line 149433: This name is already a synonym for this term, and was therefore not added as a synonym in this PR

sabrinatoro (approved):
looks good. 
I am approving

PR #10166, opened 2026-04-16
merge epilepsy of infancy with migrating focal seizures into MONDO:0017385

### LINKED ISSUE #9829: [Merge]malignant migrating partial seizures of infancy & epilepsy of infancy with migrating focal seizures
**Mondo term (ID and Label)**
MONDO:0100025 epilepsy of infancy with migrating focal seizures

**Reason for deprecation**
UMLS has mapped many Mondo X-refs from these 2 terms onto a single Unified CUI - C4518639  (Orphanet 293181, NCIT C125387) and C3494976 is no longer a valid identifier. 
Also noting that SNOMED CT 733195008 is flagged as obsolete in UMLS; instead 784345005 is used on the current CUI.

**Term to be merged with**
MONDO:0017385 malignant migrating partial seizures of infancy

MeeSiing:
close #9829

confirmed both terms represent same concept

UMLS 4518639 is mapped to both terms


src/ontology/mondo-edit.obo
```diff
@@ -578290,27 +578302,11 @@ property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/
 
 [Term]
 id: MONDO:0100025
-name: epilepsy of infancy with migrating focal seizures
-def: "This syndrome is characterized by onset of refractory focal seizures in the first year of life, with associated severe encephalopathy. Focal seizures arise independently in both hemispheres and can migrate from one cortical region to another randomly but consecutively in the same seizure. Seizures are often prolonged with episodes of status epilepticus. Prognosis is poor with severe neurological disability and reduced life expectancy, although a milder evolution has been reported in a few children." [https://orcid.org/0000-0001-8486-0558, https://www.epilepsydiagnosis.org/syndrome/infancy-migrating-focal-overview.html]
```
sabrinatoro, line 578294: It looks like there are some good information in this definition. I suggest that we add this information in a comment in the remaining term, with provenance.

sabrinatoro (changes requested):
One comment to address. 
Everything is ok regarding the merge process, but there are some information which I think would be nice to maintain.

sabrinatoro (approved):

Issue #10159, opened 2026-04-15
OMIA-Mondo alignment


katiermullen:
Three part process:

- [x] Add gene relations for existing Mondo non-human animal terms that should have gene relations based on 1. being a species-specific term, 2. having an entry in the variant table in OMIA, and 3. not already having a gene relation in Mondo. Update term labels.
- [ ] Add disease-related terms from OMIA that are not already in Mondo. Exclude susceptibility/resistance terms for now.
- [ ] Review OMIA terms that have previously been excluded and see if any of those should be added to Mondo. E.g. are some of the OMIA terms that were excluded for reason phenotype actually diseases/diagnoses according to Venom and OMIA? This would support their inclusion in Mondo.

katiermullen:
In Step 1, I used the OMIA database from 2026-03-07 and used Claude code to search for terms that were in Mondo but had gene relations that were not in Mondo and met the criteria above. I added the gene relations and updated the name:

  1. Add gene relation and updated renamed 228 term labels in mondo-edit.obo                                                                                                        
                                                                                                                                                      
  Updated the name: field for 228 existing MONDO terms  (non-human animal disease terms) to include the causal gene in the label. The pattern   
  changed from:                                                                                                                                       
                                                                                                                                                      
  ▎ disease name, species → disease name, GENE-related, species                                                                                       
   
  For example:                                                                                                                                        
  - hemophilia A, dog → hemophilia A, F8-related, dog                                                                                               
  - Wilson disease, domestic cat → Wilson disease, ATP7B-related, domestic cat                                                                        
  - polydactyly, chicken → polydactyly, SHH-related, chicken                  
                                                                                                                                                      
  All 228 IDs were found and updated, with 0 missing. The file was then normalized using make NORM. 

In doing so, I realized that some OMIA term IDs referred to multiple genes. In Mondo, these terms should each be their own disease, so:                                                 
                                                                                                                                                      
  2. Created a ROBOT template for 13 new child terms                                                                                                  
                                                                                                                                                      
  Created src/ontology/tmp/issue-10159-new-terms.tsv with 13 new gene-specific child terms under existing species-specific parent terms:        
                                                                                                                                                    
  - 2 cerebellar abiotrophy subtypes in horse (MUTYH, TOE1)                                                                                           
  - 2 degenerative myelopathy subtypes in dog (SOD1, SP110)                                                                                         
[...]

PR #10171, opened 2026-04-17
OMIA-Mondo alignment

katiermullen:
addresses #10159 

Adds disease-gene related terms from OMIA that are not yet in Mondo.
Note: I included the coat color terms in horses because they are associated with disease and later need to be renamed to reflect the actual disease they are associated with, but I wanted to start by capturing the relationship as currently present in OMIA.
Also note that I reviewed the two terms that are digenic.
Also note that another PR will address adding the non-gene related terms from OMIA that are not yet in Mondo.


sabrinatoro (approved):
no red flag. 
I approve

PR #10178, opened 2026-04-20
Update mondo-edit.obo normalization workflow.

ptgolden:
This commit makes two major changes.

First, running the QC workflow depends on `mondo-edit.obo` being normalized. If normalization modifies the file, the workflow fails early and `make test` is not run at all.

Second, it adds a way to trigger `mondo-edit.obo` normalization via a comment on a pull request. If someone types `/normalize-mondo-edit`, then:

1. The commenter's permissions are checked to see if they have write access to the repo

2. The default branch is checked out as a trusted workspace

3. `mondo-edit.obo` is shallow-cloned from the PR branch and copied into the trusted workspace

4. `make NORM` is run using trusted code from the default branch

5. If normalization changed anything, the result is committed to the PR branch and the QC CI workflow is manually re-triggered.

6. A comment is added to the PR with the result.

This ensures normalization is enforced without requiring contributors to run ODK locally.

NOTE: This adds the normalization step to every PR, even ones that don't touch `mondo-edit.obo`. It's possible to change this, but it would add some complexity and `make NORM` is not particulary expensive.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -28,8 +28,24 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: Normalize mondo-edit.obo
+        run: |
+          cd src/ontology
+          make NORM && mv NORM mondo-edit.obo
+
+      - name: Fail if mondo-edit.obo is not normalized
+        run: |
+          cd src/ontology
+          if ! git diff --quiet -- mondo-edit.obo; then
+            echo "::error file=src/ontology/mondo-edit.obo::mondo-edit.obo is not normalized. Comment /normalize-mondo-edit to run normalization, or run make NORM && mv NORM mondo-edit.obo yourself and commit the changes."
+            git --no-pager diff -- mondo-edit.obo
+            exit 1
+          fi
+
       - name: Run ontology QC checks
         env:
           DEFAULT_BRANCH: master
-        run: cd src/ontology; make GITHUB_ACTION=true IMP=false PAT=false test -B
+        run: |
+          cd src/ontology
+          make GITHUB_ACTION=true IMP=false PAT=false test -B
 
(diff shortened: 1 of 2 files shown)


matentzn:
This looks coherent, but I want to know why you chose not to take this approach:

in `main.yaml`:

1. checkout
2. normalise
3. commit + push changes to mondo-edit.obo directly to branch
4. validate

ptgolden:
I didn't put normalization-and-push directly into `main.yaml` because I wanted QC to be a validation workflow, not a mutation workflow. I think this model is clearer and more predictable:

- Required QC checks remain deterministic and side-effect free
- Branch mutations happen only in response to an explicit action
- It's clear whether a commit came from a person or from the normalization helper

It's true that the addition of the "/normalize-mondo-edit" directive introduces something new for curators/maintainers to learn. But I think it's worth it in this case. I prefer the model where a PR is only modified in response to an explicit action. It also opens the door for similar directives in the future if we want them.

---

There's also a secondary safety reason not to do the normalize-and-push flow directly from `main.yaml`. If the `NORM:` recipe in `mondo.Makefile` were ever modified in a PR, the workflow would be running branch-mutating logic defined in PR-controlled code. Even if this repository is write-restricted, I think it's better to keep that boundary explicit and only allow branch mutation through a separate helper that runs trusted code from the default branch.

We could add logic that mitigates this in `main.yaml`, but that gets back to my initial reasoning: adding all of that would significantly increase the complexity of what should be a step that performs validation. Keeping QC simple and side-effect free feels like the cleaner separation.

ptgolden:
One other thing I just thought of. If the normalization were automatically committed in the QC step, `main.yaml` would become self-reinvoking (normalize -> push -> re-trigger). It should converge after one extra run, but it still introduces a feedback loop into what should be a simple validation workflow, which makes the behavior harder to reason about.

matentzn (approved):

PR #10184, opened 2026-04-21
Second stab at new QC/normalization coordination

ptgolden:
This commit automates and simplifies what was in 6f00a80.

After these changes, there are now two reusable workflows, which are coordinated by a new "orchestrator" workflow.

- `normalize-mondo-edit.yaml` checks if mondo-edit is normalized in a PR, and pushes a commit to it if it is not.
- `qc.yaml` runs `robot verify ...` on the repo

The orchestrator workflow, `pr-orchestrator.yaml`, is triggered on all PR events and coordinates these two steps.

It first runs the normalization workflow. If normalization produces a new commit, the orchestrator stops.

The new commit triggers another PR event, which re-runs the orchestrator. On this second run, normalization is a no-op and the QC step proceeds.


matentzn (changes requested):
The orchestrator pattern is clean and I like the direction — splitting normalize and QC into reusable `workflow_call` workflows is a good simplification over #10178. But IMO there's a problem with the re-trigger mechanism.

The PR body says "The new commit triggers another PR event, which re-runs the orchestrator." GitHub Actions explicitly prevents this — events triggered by `GITHUB_TOKEN` [do not create new workflow runs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). So after normalization pushes a commit, no new `pull_request` event fires, the orchestrator never re-runs, and QC never executes. The PR would sit with a green normalize check and no QC check at all.

**Other things I would like some clarity on:**

- **Fork PRs lost their guard.** #10178 had a check for `isCrossRepository` — this PR removed it. The normalize workflow will try to push to a fork branch and fail. Worth either re-adding the guard or at least skipping normalization for fork PRs.

- **Deleted `main.yaml` → check name change.** If branch protection requires the old "CI / ontology_qc" check, the new check name from the orchestrator will be different (something like "PR orchestrator / ontology_qc / ontology_qc"). Branch protection rules would need updating. Not a code issue, just a deployment detail to keep in mind.

ptgolden:
> The PR body says "The new commit triggers another PR event, which re-runs the orchestrator." GitHub Actions explicitly prevents this — events triggered by GITHUB_TOKEN [do not create new workflow runs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). So after normalization pushes a commit, no new pull_request event fires, the orchestrator never re-runs, and QC never executes. The PR would sit with a green normalize check and no QC check at all.

UGH. How could I forget. Let me do a little more thinking...

It would mess up the modularity a little bit. But maybe QC could be called from the normalization workflow itself if it ended up changing something. That way, things work like:

If normalization workflow results in a commit, QC is run from `normalize-mondo-edit.yaml` and skipped in `qc.yaml` (since `qc.yaml` would be running code from an out-of-date `HEAD`).

If normalization workflow does not result in a commit, QC is run from `qc.yaml`.

This would require not doing the safe normalization steps (pulling the repo from the `main` branch, and only pulling `mondo-edit.obo` from the PR), but I think that's okay. The same issue is in the QC workflow anyway-- any PR could edit the `test:` target and make it run whatever. We have write protections for PRs for a reason.

> Fork PRs lost their guard. https://github.com/monarch-initiative/mondo/pull/10178 had a check for isCrossRepository — this PR removed it. The normalize workflow will try to push to a fork branch and fail. Worth either re-adding the guard or at least skipping normalization for fork PRs.

I took it out to simplify things, but I'll add it back in.

> Deleted main.yaml → check name change. If branch protection requires the old "CI / ontology_qc" check, the new check name from the orchestrator will be different (something like "PR orchestrator / ontology_qc / ontology_qc"). Branch protection rules would need updating. Not a code issue, just a deployment detail to keep in mind.

Yep- I have that in mind.

ptgolden:
If I did what I suggested above, I think it's still worth it to keep the concurrency stuff in. No use on keeping a QC job running if a new commit comes in.

matentzn (approved):
Lets try this

ptgolden:
This is not going to work, for a familiar reason.

If the normalize step results in a commit, (let's call the original HEAD `A` and the normalize commit `B`), then the QC step that runs within `normalize-mondo-edit.yaml` will be run against `B`... but the action will be reported for commit `A`, since `A` was the action for which the workflow was triggered.

Because `B` is generated by a commit with `GITHUB_TOKEN`, it will *not* trigger workflows, and will stay waiting for the QC step forever.

Rather than experimenting here more, I created a repository to experiment in, and this is what I came up with: https://github.com/ptgolden/actions-testing

@matentzn let me know what you think about that approach.

matentzn:
Just for the record - if you can convince us both this is safe to do, I am fine with the GitHub App route

PR #10186, opened 2026-04-22

matentzn:


sabrinatoro (approved):
This looks good. 
Note that we lost a lot of x-ref for exact synonyms. We should think of how to handle synonyms without x-ref.

PR #10231, opened 2026-05-05
add infectious diseases from venom

katiermullen:
addresses #5726


sabrinatoro (approved):
no red flags. 
I am approving and merging

PR #10239, opened 2026-05-08
Add MONDO:7770747 hepatic fibrinogen storage disease (human)

### LINKED ISSUE #10238: Add term for hepatic fibrinogen storage disease (human)
MONDO has no dedicated term for the human disease "Hepatic Fibrinogen Storage Disease" (HFSD), also known as "Hereditary Hypofibrinogenemia with Hepatic Storage" (HHHS). The animal counterpart exists (`MONDO:1011782`) but no human sibling was created.

## Context

HFSD/HHHS is an autosomal dominant disorder caused by missense mutations in `FGG` (most commonly the Aguadilla variant, p.Arg375Trp / c.1201C>T). Variant fibrinogen is retained in the hepatocellular endoplasmic reticulum, producing hypofibrinogenemia and variable liver disease ranging from mild elevation of liver enzymes to cirrhosis. It is mechanistically distinct from generic hypofibrinogenemia (no hepatic storage) and from dysfibrinogenemia (functional anomaly of circulating fibrinogen).

Current MONDO state and why none of the existing terms can be reused:

- `MONDO:1011782` "hepatocellular fibrinogen storage disease, **non-human animal**" — the mechanistic match, but explicitly scoped to animals (parented under non-human animal disease, xref `OMIA:002582`). Has one cattle child `MONDO:1012569` (DGKG-related). Cannot be repurposed for human use.
- `MONDO:0015096` "familial hypofibrinogenemia" — **not reusable.** Already mapped equivalent to `Orphanet:101041`, `MEDGEN:1826143`, `UMLS:C5681803`, all of which describe a purely quantitative deficiency of circulating fibrinogen with mild post-traumatic bleeding — no liver pathology, no intracellular aggregation. HFSD is a mechanistically distinct subset (variant fibrinogen retained in hepatocyte ER, leading to liver disease up to cirrhosis). Renaming/redefining `MONDO:0015096` would silently change the meaning of the existing UMLS/MEDGEN/Orphanet equivalents. HFSD should be a child of this term, not a replacement.
- `MONDO:0018060` "congenital fibrinogen deficiency" — **not reusable.** This is a disorder grouping (`Orphanet:335`) that encompasses afibrinogenemia, hypofibrinogenemia, dysfibrinogenemia, and hypodysfibrinogenemia. Far too broad; HFSD is a specific subtype.
- `MONDO:0014452` "familial dysfibrinogenemia" carries `OMIM:616004`. OMIM 616004 = "DYSFIBRINOGENEMIA, CONGENITAL" — mapping is correct, but it's a different condition (functional anomaly of circulating fibrinogen, not ER storage), so not reusable for HFSD.

External reference grounding:

- No dedicated OMIM phenotype number — described under the `FGG` locus entry (`OMIM:134850`) via allelic variants (Aguadilla, Brescia, Angers, etc.).
- No dedicated Orphanet entry — folded into `Orphanet:101041` (familial hypofibrinogenemia).

## Expected behaviour

A new MONDO term for the human disease, e.g. "hepatic fibrinogen storage disease" with synonym "hereditary hypofibrinogenemia with hepatic storage" (HHHS).

## Suggested approach

- Name: `hepatic fibrinogen storage disease`
- Synonym (EXACT): `hereditary hypofibrinogenemia with hepatic storage`
- Synonym (EXACT, ABBREVIATION): `HFSD`, `HHHS`
- Parent: `is_a MONDO:0015096` (familial hypofibrinogenemia)
- Logical definition along the lines of `MONDO:0015096` ∩ `has_material_basis_in_germline_mutation_in FGG` (HGNC:3694), if the curator agrees the disease-by-gene pattern is appropriate — but note Aguadilla and other FGG variants drive the HFSD phenotype specifically via ER retention, so a plain gene-mutation logical def may overgeneralise. Probably better to leave this as a textual subtype of familial hypofibrinogenemia for now.
- Definition source: PMID:31965886, PMID:33105716
- xref `OMIM:134850` (FGG locus, closest external grounding) — flag as related, not equivalent.
- Cross-reference the existing animal term `MONDO:1011782` in a comment for curator awareness.

## Related

- `MONDO:1011782` — non-human animal counterpart
- `MONDO:0015096` — proposed parent
- PMID:31965886 — Aguadilla case report ([PMC7169362](https://pmc.ncbi.nlm.nih.gov/articles/PMC7169362/))
- PMID:33105716 — HHHS review ([PMC7659954](https://pmc.ncbi.nlm.nih.gov/articles/PMC7659954/))

matentzn:
Adds a human-scoped term for HFSD / hereditary hypofibrinogenemia with hepatic storage (HHHS), an FGG-mutation disorder with ER retention of variant fibrinogen. Parented under MONDO:0015096 (familial hypofibrinogenemia). Companion to the existing animal term MONDO:1011782; OMIM:134850 (FGG locus) added as MONDO:relatedTo. No logical definition (per issue: a plain gene-mutation def would over-generalise across afibrinogenemia/dysfibrinogenemia FGG variants).

Closes https://github.com/monarch-initiative/mondo/issues/10238


matentzn:
@katiermullen my main issue here is this commit: [`05803c7` (this PR)](https://github.com/monarch-initiative/mondo/pull/10239/commits/05803c76787c1af17d65f4592c1e5b113530c98f)

So my conflict is this: 

1. We added one subtype
2. There are likely more
3. Why did we add one and not the others

You need to teach me precisely how to mentally solve this conundrum.

katiermullen:
@matentzn,

Thank you for working on this issue. I think that it is complicated because there are different ways to classify the subtypes depending on the source. The mental process that I use is to see if 1. there are established classification schemes in the literature with clear definitions and then 2. to see how these relate to the classification of the diseases in OMIM and orphanet. 

Interestingly in this case, OMIM groups all of these disorders as one:

Screenshot 2026-05-08 at 4 33 29 PM

Mondo has a classifications that could be improved in my opinion. 
For example, I do not think familial hypofibrinogenemia should be a child of congenital afibrinogenemia because the part term refers to no fibrinogen and the child refers to decreased fibrinogen:
Screenshot 2026-05-08 at 4 38 03 PM

I actually think that this ORDO classification makes more sense:
Screenshot 2026-05-08 at 4 35 13 PM

In my opinion, with this classification, your term would still be a child of familial hypofibrinogenemia. 

Then the next issue to solve is related to genes. Your term clearly is related to variations in the FGG gene based on the PMIDs you provide. How granular you make the classification of the other terms is up for debate and probably will require the advice of an expert or more research. 

For example, this PMID: 30332696 suggests that "Afibrinogenemia was almost always caused by null mutations (98.6%), mainly in FGA (85%). Hypofibrinogenemia was mainly caused by missense mutations of FGB or FGG (54.2%). Dysfibrinogenemia was almost always caused by heterozygous missense mutations (99.3%) in FGA and FGG." Do we give each its own Mondo ID?

If I were your Claude, I would recommend reclassifying this branch in accordance with the ORDO, adding your term as a child of familial hypofibrinogenemia as you have, add the gene relation to the new term with the PMIDs as the sources, and seeking expert input (or deep research) on whether there is enough evidence to include to include the gene specific children of hypofib, afib, and dysfib. 

I hope this helps?

matentzn:
@katiermullen — yes, this helps a lot, thank you!! Massively appreciated.

1. **Adding `relationship: has_material_basis_in_germline_mutation_in HGNC:FGG` to HFSD with PMID:31965886 / PMID:33105716 as sources** — oh yes. I deliberately left out the `intersection_of` (logical definition) because it would over-generalise — FGG missense variants also drive hypo and dysfib subtypes, so `familial hypofibrinogenemia ∩ FGG-mutation` doesn't equal HFSD. But the bare relationship is a true claim that the PMIDs support, and it's local to the new term. Added that.

2. **`familial hypofibrinogenemia is_a congenital afibrinogenemia` is broken** — agree, "no fibrinogen" can't be a parent of "reduced fibrinogen". Made #10241 as follow up.

3. **Gene-specific subtypes (FGA-related afib, FGG-related hypofib, FGA-related dysfib, etc.)** — Opened issue for that: #10242.

On the workflow meta-point: I'd typically like to keep this (and any PR analogously) scoped to "add HFSD". Local additions to the new term itself (1) are in scope; restructuring the surrounding tree (2, 3) feels like it belongs in its own PRs — both changes get the review attention they deserve, and the equivalence mappings on `MONDO:0015096` don't silently shift inside an NTR diff. I think the healthy pattern when a reviewer spots an adjacent structural issue is "follow-up issue", not "expand the PR" — otherwise contributors who care enough to notice get punished, and you end up either holding NTRs hostage or letting the structural debt accumulate. Happy to be argued out of this if you see it differently — but my plan is to land (1) here and file (2) and (3) as issues today.

Thanks for the careful read, and especially for the screenshots — they made the argument click immediately. Is there

katiermullen:
@matentzn I agree with you on the workflow meta-point in principle, but I would say in practice that sometimes the "can of worms" gets addressed in a single issue with documentation in the issue of the changes that were made. Sometimes it just feels more efficient to get everything done in a single issue - but this is probably not best practice.

src/ontology/mondo-edit.obo
```diff
@@ -664353,6 +664353,20 @@ is_a: MONDO:1011302 {source="https://orcid.org/0000-0002-5002-8648"} ! branchial
 relationship: in_taxon NCBITaxon:9986 {source="OMIA:000618-9986"} ! Oryctolagus cuniculus
 property_value: IAO:0000233 "https://github.com/monarch-initiative/mondo/issues/10159" xsd:anyURI
 
+[Term]
+id: MONDO:7770747
+name: hepatic fibrinogen storage disease
+def: "An autosomal dominant disorder characterized by hypofibrinogenemia and variable liver disease (ranging from asymptomatic transaminase elevation to cirrhosis), caused by missense variants in FGG that lead to retention of variant fibrinogen as eosinophilic inclusions within the hepatocellular endoplasmic reticulum." [https://orcid.org/0000-0002-7356-1779, PMID:31965886, PMID:33105716]
```
katiermullen: I would consider changing the definition to "cased by variations in FGG" and leave out missense because in this PMID: [33105716](https://pubmed.ncbi.nlm.nih.gov/33105716/) Table 2 one person was reported to have a deletion + splicing mutation.

katiermullen (changes requested):
@matentzn just one change requested

src/ontology/mondo-edit.obo
matentzn: I updated this to caused by variants in FGG, seemed a bit more standard - or do you have a reason to prefer "variations in"?

katiermullen (approved):

PR #10244, opened 2026-05-11
Change normalization/QC workflow to use custom statuses

ptgolden:
This supersedes #10184 and replaces the workflow implemented in #10178.

This should simplify the normalization & QC workflows, and get around the problems discussed in https://github.com/monarch-initiative/mondo/pull/10184#issuecomment-4371807437 without resorting to registering a GitHub App.

The PR orchestrator workflow described in #10178 is abandoned in favor of having the `normalize-mondo-edit.yaml` workflow triggered on pull request events. This workflow checks if a file is normalized, and automatically pushes a normalization commit if normalization is needed. (This automatic normalization/push is not run on fork PRs. In that case, it simply raises an error and instructs the PR author to run normalization manually).

The difference from the previous workflow is that **the QC workflow is not triggered automatically**. Instead, it is dispatched manually **on a specific commit**:

```
gh api \
  --method POST \
  "repos/${GITHUB_REPOSITORY}/actions/workflows/qc.yaml/dispatches" \
  -f "ref=${PR_BRANCH}" \
  -f "inputs[sha]=${{ steps.push.outputs.sha }}" \
  -f "inputs[pr_number]=${PR_NUMBER}"
```

If normalization did not occur, it is triggered on the original commit. If normalization *did* occur, it is triggered on the normalization commit.

The issue so far is that, even if QC is dispatched in this way, it does not automatically show up as a status check for the commit SHA in the pull request. Instead of relying on that, this commit **adds a [custom commit status](https://docs.github.com/en/rest/commits/statuses) that is posted by the QC workflow**. This is allowed with the `statuses: write` permission and *does not* need a GitHub App token. The process by which the status is updated is in the `pending` and `report` jobs in `qc.yaml`. **This custom status should be the target for branch protection**.

A proof of concept for this workflow is in the repository <https://github.com/ptgolden/actions-testing-v2/>.

---

Benefits of this approach:

* Normalization is triggered automatically and will always run before QC (on non-fork PRs).
* No GitHub App token is needed to re-trigger workflows.
* There is no recursive calling of the normalization/QC workflow.

Potential risks:

* The QC status checks are now manually posted, and do not rely on GitHub's built-in action check mechanism.
* Workflows still run arbitrary code from `make NORM` and `make test`-- although it is worth noting that **this has always been the case with these workflows**. It may be worth hardening things by not running these steps if any file in `.github/workflows/.*`, `src/ontology/*Makefile`, `src/ontology/scripts/*`, and other code-executing files is edited.


.github/workflows/normalize-mondo-edit.yaml
```diff
@@ -1,162 +1,74 @@
 name: Normalize mondo-edit.obo
 
 on:
-  # Triggers on comments that start with /normalize-mondo-edit
-  issue_comment:
-    types: [created]
+  pull_request:
+    branches: [ main ]
```
matentzn, line 5: surely not, right (copy paste error.. wrong branch)

.github/workflows/normalize-mondo-edit.yaml
```diff
@@ -1,162 +1,74 @@
...
+          repository: ${{ env.PR_REPO }}
+          ref: ${{ env.PR_BRANCH }}
           fetch-depth: 1
-          token: ${{ github.token }}
-          sparse-checkout: |
-            src/ontology/mondo-edit.obo
-          sparse-checkout-cone-mode: false
 
       - name: Configure git
+        if: ${{ env.SAME_REPO == 'true' }}
```
matentzn: so no QC _ever_ on forks? I think its OK - if there is a way for us to trigger it manually.

matentzn: Esp given the fact that normalisation IS run on forks..

- normalize step runs `make NORM`
- if NORM changes the file → `exit 1` with the "run normalization locally" message (good)
- if NORM doesn't change the file → step succeeds, but the QC dispatch is skipped, so `required-qc-check` is never posted
- branch protection then blocks the merge indefinetely

.github/workflows/normalize-mondo-edit.yaml
```diff
@@ -1,162 +1,74 @@
...
-      # Run normalization using code from the default branch (trusted), not
-      # the PR branch, to avoid executing potentially modified arbitrary code
-      # in Makefiles or scripts
-      - name: Normalize PR file using trusted code
+      - name: Normalize PR file
         run: |
-          cp pr/src/ontology/mondo-edit.obo trusted/src/ontology/mondo-edit.obo
-          cd trusted/src/ontology
+          cd src/ontology
           make NORM && mv NORM mondo-edit.obo
```
matentzn, line 40: If `make NORM` blows up for any reason (ODK image change, OOM, transient infra), the workflow errors out before the QC dispatch step. Result: no `required-qc-check` status is posted at all, the PR sits indefinitely with branch protection blocking merge, and the only signal is buried in the normalize workflow logs. Could post a `failure` status with a useful description if normalize itself fails?

.github/workflows/qc.yaml
```diff
@@ -0,0 +1,93 @@
+name: Ontology QC
```
matentzn, line 1: branch protection needs to be repointed from whatever the old `main.yaml` check was to `required-qc-check` before this merges, otherwise we'll have a window where nothing is enforced.

ImportanT!

matentzn (changes requested):

ptgolden:
As of the latest commit (8a5231b), this is what the flow of PRs looks like:

```mermaid
flowchart TD
    Start([Pull request event])

    Start -->|same-repo PR| Norm
    Start -->|fork PR| ForkOrch

    subgraph Norm[normalize-mondo-edit.yaml — on: pull_request]
        N1[Checkout PR branch] --> N3{Is mondo-edit normalized?}
        N3 -->|No| N4[Commit + push<br/>normalization to PR branch]
        N4 --> N6[Dispatch qc.yaml<br/>with HEAD sha]
        N3 -->|Yes| N6
    end

    subgraph ForkOrch[fork-pull-orchestrator.yaml — on: pull_request_target]
        F1[Dispatch qc.yaml<br/>with fork sha, from main]
    end

    N6 --> QC
    F1 --> QC

    subgraph QC[qc.yaml — on: workflow_dispatch]
        Q1[Mark required-qc-check: pending]
        Q1 --> Q3[Checkout input sha]
        Q3 --> QN{mondo-edit is normalized?}
        QN -->|No| Q6[Status: failure]
        QN -->|Yes| QT[Run make test]
        QT --> Q4{Result?}
        Q4 -->|pass| Q5[Status: success]
        Q4 -->|fail| Q6
        Q4 -->|cancelled| Q7[Status: error]
    end

```

ptgolden:
To summarize:

* Same-repo PRs trigger the workflow `normalize-mondo-edit.yaml`. They auto-commit any necessary normalization, and then manually dispatch `qc.yaml` with the correct SHA.
* Fork PRs trigger the workflow `fork-pull-orchestrator.yaml`-- importantly with the `pull_request_target` event, which allows for a read/write `GITHUB_TOKEN` (as opposed to the `pull_request` event, which gets a read-only token from forks). This workflow *just* triggers `qc.yaml` and does not run any untrusted code that would have access to the read/write token. Read the commit message in 8a5231b to understand why I believe this is safe.

One consequence of this is that any workflow which might result in automatic commits-- e.g. EMC, synonym sync, and so on-- needs to be amended to dispatch `qc.yaml` manually. This will get around our current dependence on PATs as described in https://github.com/monarch-initiative/mondo/pull/10260#issuecomment-4547671281.

.github/workflows/normalize-mondo-edit.yaml
ptgolden: The `issue_comment` event was for a previous approach, when we said you could type `/normalize-mondo-edit` to trigger normalization. This triggers on pull request events.

.github/workflows/qc.yaml
ptgolden: Correct!

.github/workflows/normalize-mondo-edit.yaml
ptgolden: This is a good point. Let me add one more thing to my latest changes.

Issue #9979, opened 2026-02-25
Request to relabel 46 XY Disorder of Sex Development (MONDO:0020040)


GabrielleGCA:
46 XY Disorder of Sex Development (MONDO:0020040)

46 XY Differences of Sex Development 

[https://clinicalgenome.org/affiliation/40073/](https://clinicalgenome.org/affiliation/40073/)

Patient community request. Requested by ClinGen's Male Infertility GCEP.

@ErinRiggs

MeeSiing:
Hi @GabrielleGCA, thank you for the ticket. A PR has been created to relabel the term and it will be reviewed soon.

PR #10271, opened 2026-05-21
relabel MONDO:0020040 46 XY Disorder of Sex Development

MeeSiing:
address #9979 

relabeled MONDO:0020040 to 46 XY differences of sex development because all the database cross references (Orphanet, NCIT and GARD) have this term name.


sabrinatoro (approved):
approving

Issue #10270, opened 2026-05-21
Fix case-insensitive synonym/label duplicates surfaced by updated QC checks


matentzn:
I made three synonym QC queries case-insensitive #10269, which surfaced ~565 violations that the case-sensitive versions were missing. We need to work through these and clean them up.

## Queries and counts

- `qc-duplicate-exact-synonym-no-abbrev` (197) — flags when the same string appears (case-insensitively) as a label or exact synonym on two different non-deprecated MONDO terms, excluding ABBREVIATION synonyms.
- `qc-syn-equal-label` (164) — flags when a related/narrow/broad/close synonym on a term matches the term's own label (case-insensitively), meaning the synonym duplicates the label.
- `qc-related-exact-synonym` (204) — flags when a related synonym matches an exact synonym on the same term (case-insensitively), meaning the synonym is recorded twice with conflicting scopes.

Violation tables (one tab per query): https://docs.google.com/spreadsheets/d/1-7GD2qoymKEsCFFohCwsbx78g9gyGTBP8RaqXomSNeg/edit

sabrinatoro:
@matentzn I am done reviewing the duplicate-synonyms spreadsheet (except for 5 cases that are a bit more complicated and require a bit more research). 

The majority of the duplicate exact synonyms represented 
- terms that should probably be merged (I am still reviewing these - there are only 3-4 of these)
- legacy synonyms from terms that referred to a general disease, but later became a subtype when another gene was discovered (mainly originating in OMIM). These synonyms came from xrefs referring to OMIM, for example NORD xref/preferred terms. I have reported and shared a spreadsheet of the NORD terms to be reviewed with Gioconda and asked her to update the NORD table we import. 
- synonyms for terms "included" in a database entry (a lot from NCIT), but are more specific and are their own terms in Mondo.

PR #10288, opened 2026-05-27
remove related synonyms that are the same as term label

sabrinatoro:
This PR removes all the related synonyms that are the same as the term label. The list is on the ["syn-equal-label" sheet here](https://docs.google.com/spreadsheets/d/1-7GD2qoymKEsCFFohCwsbx78g9gyGTBP8RaqXomSNeg/edit?gid=2025999783#gid=2025999783)

This spreadsheet was reviewed. Out of 164 rows:
  - 3 had x-ref to OMIM, all "formerly" names
  - 5 had x-ref to GARD (probably a remnant from when we added all GARD name as synonym)
  - 39 had x-ref to MESH (I couldn't find how these synonyms were added as related in Mondo)
  - 59 had x-ref to UMLS (I couldn't find how these synonyms were added as related in Mondo)
  - 61 had no x-ref

After review, I think all these synonyms can be removed.

NOTE:
1- there are several "syntaxic" related synonym that could also be removed. for example:

REMOVED: synonym: "Vascular Brain injury" RELATED [MESH:D020214] 
**synthaxic synonyms that could also be removed:**
synonym: "Vascular injury, Brain" RELATED [MESH:D020214] synonym: "Vascular trauma, Brain" RELATED [MESH:D020214] synonym: "Vascular Traumas, Brain" RELATED [MESH:D020214]


matentzn (approved):
Always happy to see conflicting related synonyms go away :)

sabrinatoro:
addressed https://github.com/monarch-initiative/mondo/issues/10270

Issue #9976, opened 2026-02-24
Request for new synonym [Add  Congenital myopathy 3 with rigid spine to SELENON-related myopathy]


galyea123:
### Mondo term

SELENON-related myopathy

### Synonyms

Congenital myopathy 3 with rigid spine

### Synonym type

exact

### Additional comments

Note that OMIM list several alternete numbers and is also linked to several Orphanet entries. Thank you, Gioconda Alyea

MeeSiing:
@galyea123, thanks for the request. 

[OMIM:602771](https://omim.org/entry/602771) shows that Congenital myopathy 3 with rigid spine represents the same concept as MONDO:0011271 rigid spine muscular dystrophy 1, not SELENON-related myopathy. Therefore, MONDO:0011271 will be relabeled to Congenital myopathy 3 with rigid spine. 

As you've mentioned, multiple Orphanet entries are linked to OMIM:602771. Based on that, I think that MONDO:0019398 'desmin-related myopathy with Mallory body-like inclusions' and MONDO:0017939 'classic multiminicore myopathy' may need to be merged with MONDO:0011271. But investigation needs to be done to confirm that.

galyea123:
Hello,

After reviewing this further, the only true synonym for SELENON-related myopathy is SEPN1-related myopathy, as this simply reflects the older gene nomenclature (SEPN1 was renamed SELENON).

SELENON-related myopathy is an umbrella term encompassing a spectrum of early-onset muscle disorders caused by loss-of-function mutations in the SELENON (formerly SEPN1). The following are histopathological subtypes of SELENON-related myopathy, distinguished by muscle biopsy findings:

SELENON-related multiminicore disease 

SELENON-related congenital fiber-type disproportion 

Rigid spine muscular dystrophy 1 (RSMD1) 

Desmin-related myopathy with Mallory body-like inclusions 

In practice, these subtypes share a largely overlapping clinical phenotype (axial weakness, early respiratory insufficiency, spinal rigidity, scoliosis), and the histopathological pattern on biopsy can even vary over time or between muscles in the same patient. 
This is precisely why the unifying term "SELENON-related myopathy" was proposed,  to emphasize that the genetic etiology, not the biopsy pattern, defines the disease.    

[...]

PR #10292, opened 2026-05-28
add exact synonym to 'SELENON-related myopathy'

MeeSiing:
address #9976


sabrinatoro:
@MeeSiing I think we should wait until we have confirmed the merge.
Right now, "congenital myopathy 3 with rigid spine" is also an exact synonym to OMIM:602771 which is equivalent to another MONDO term. This will bring confusion if we decide (for whatever reason) that we don't want to merge.

MeeSiing:
replace by #10586

PR #10295, opened 2026-05-28
add venom inflammatory diseases

katiermullen:
addresses #5726

spreadsheet is here: https://docs.google.com/spreadsheets/d/14J9ZGjZjNGhXaRov7csdWrnXKPdIeSpo2O-3zfatv8c/edit?gid=302212382#gid=302212382


sabrinatoro (approved):
no red flag. 
However, I think we could (in a future PR) create definitions based on location (e.g. all the "abscess" terms)

katiermullen:
@sabrinatoro where possible I did try to include a second parent for the location of the abscess, but I agree a logical definition would be better in the future :)

katiermullen:
@sabrinatoro it seems like what happened is the update I made to add the single child exclusion to tick bite granuloma did not persist and now there is a single child violation.

PR #10299, opened 2026-05-29
add venom foreign body disorders

katiermullen:
addresses #5726

spreadsheet is here: https://docs.google.com/spreadsheets/d/1c3LqsvlsrY_5BUvDoKQi18XAkYUqKLY91BraB7ugsoM/edit?gid=335554050#gid=335554050


sabrinatoro (approved):
@katiermullen I am approving this PR, but could you please give a comment about why you think these terms should be in Mondo (they don't look like diseases to me).

katiermullen:
@sabrinatoro foreign body ingestions/inhalations/injuries are commonly encountered disorders in veterinary patients. Depending on the type and location these disorders can necessitate a series of diagnostic procedures and treatments. For example, according to PMID: [42088454](https://pubmed.ncbi.nlm.nih.gov/42088454/) foreign body ingestion accounts for approximately 80% of all mechanical intestinal obstructions in dogs. 

In the future, I think we need to write definitions to make it clear that these are disorders/diseases e.g., "intestinal obstruction due to foreign body ingestion."

PR #10319, opened 2026-06-02
split/create new term MONDO:0100585 myopathy, centronuclear, 1

### LINKED ISSUE #10318: [Revise xref] MONDO:0008048 'autosomal dominant centronuclear myopathy'
MONDO:0008048 'autosomal dominant centronuclear myopathy'

The [OMIM xref](https://omim.org/entry/160150) for this term is referring to the monogenic form of the disease (based on DNM2 gene) and the label should be MYOPATHY, CENTRONUCLEAR, 1. The xref to [Orphanet](https://www.orpha.net/en/disease/detail/169189) is a disease that has 5 genes implicated.

I think we should create a new term for myopathy, centronuclear 1. 

I haven't split a term in a while but I think this is a case where we could just create a new term that is equiv to OMIM without obsoleting the existing class. I'll do the PR but @sabrinatoro let me know if you disagree, thanks!

nicolevasilevsky:
close #10318


src/ontology/mondo-edit.obo
```diff
@@ -181207,28 +181207,18 @@ subset: prototype_pattern
 subset: rare
 synonym: "AD-CNM" EXACT ABBREVIATION [DOID:0111217, Orphanet:169189]
 synonym: "autosomal dominant centronuclear myopathy" EXACT [DOID:0111217, Orphanet:169189] {OMO:0002001="https://w3id.org/information-resource-registry/clingen"}
-synonym: "autosomal dominant centronuclear myopathy caused by mutation in MYF6" EXACT [MONDO:design_pattern]
```
nicolevasilevsky, line 181210: I removed this as I don't think it is correct. should I deprecate it instead?

src/ontology/mondo-edit.obo
```diff
@@ -181207,28 +181207,18 @@ subset: prototype_pattern
...
-synonym: "centronuclear myopathy 1" EXACT [DOID:0111223, NCIT:C126689]
 synonym: "centronuclear myopathy, autosomal dominant" EXACT [MONDO:patterns/autosomal_dominant]
 synonym: "centronuclear myopathy, autosomal, modifier of" EXACT []
-synonym: "CNM1" EXACT ABBREVIATION [DOID:0111223, NCIT:C126689, OMIM:160150]
 synonym: "CNM3" RELATED ABBREVIATION []
-synonym: "DNM2-related centronuclear myopathy" RELATED [GARD:0012719]
 synonym: "myopathy, centronuclear, 1" EXACT [GARD:0012719, OMIM:160150]
 synonym: "myopathy, centronuclear, 3" EXACT []
```
nicolevasilevsky, line 181213: I think this should be removed too, not sure where it came from and it doesn't make sense to have it here

src/ontology/mondo-edit.obo
```diff
@@ -181207,28 +181207,18 @@ subset: prototype_pattern
...
 synonym: "myopathy, centronuclear, autosomal dominant" EXACT [OMIM:160150]
-synonym: "myopathy, centronuclear, type 1" EXACT []
 synonym: "myopathy, centronuclear, type 3" EXACT []
```
nicolevasilevsky, line 181215: same with this, I think this should be removed

PR #10326, opened 2026-06-04
NT for AIP-related pituitary adenoma predisposition

MeeSiing:
close #10131

It was stated in PMID that this predisposition is a low penetrance and does not have not strong family history. Is there a better parent term than 'hereditary neoplastic syndrome'?

While researching for this issue, I noticed that OMIM merged OMIM:600634 with OMIM:102200, so we should probably merge MONDO:0010911 'prolactin-producing pituitary gland adenoma' with MONDO:0007052 'growth hormone secreting pituitary adenoma 1' [Note](https://docs.google.com/presentation/d/1E4STssSmMnXrg1aZuW2oBq1QMaI4fd1dTzSs2IEhOxE/edit?slide=id.p#slide=id.p)


sabrinatoro:
@MeeSiing I am not sure about the parentage. 
I think we should message the requester and ask if, in their opinion, this term should be considered to be 
- a hereditary neoplastic syndrome MONDO:0015356 (is i
- a cancer term (ie should be in the neoplasm branch of mondo) or a predisposition (not in the cancer branch, only in the "susceptibility" branch

Also, is this the same as OMIM:102200 (the included entry)?

MeeSiing:
@sabrinatoro thanks for reviewing. I will reach out to the requester. yes, this term is the same as OMIM include entry, and it is not a separate Mondo label yet. I forgot to ask an important question. I know we follow the OMIM naming, but should we use the ClinGen label because they request for the term first?

sabrinatoro (commented):
@MeeSiing This looks good, but I think you can add a logical definition. 
Also, there is one file that I didn't expect to be changed (and I don't know how to remove these changes from this PR). 

My proposal (just to be safe) is: would you mind creating a new PR for this term (you can copy/paste the stanza for this new term, paste it in the mondo-edit.obo file, and run the normalization. It should be "quick"). + add the logical definition

MeeSiing:
@sabrinatoro I have added the logical definition and got rid of unexpected file change.

sabrinatoro (approved):
looks good

PR #10376, opened 2026-06-19
Remove some cruft from src/ontology

ptgolden:
I was puttering around and I noticed these as low-hanging fruit 

* 9 release files from 2020-12 to 2021-08
* An errantly committed `mondo-edit` OBO file
* Two files that seem like they may have been errantly committed in the 2021-11-01 release
* `get-new-classes.sh`, which was a temporary script until https://github.com/ontodev/robot/issues/432 was merged in 2020. (it does not appear anywhere in the `Makefile`).


matentzn (approved):
Its best to use "git rm" rather than bash rm, but in any case, THANKS for the cleanup. Would be nice to get many more of these :)

ptgolden:
I did use `git rm`!

Also, `git rm FILE` is equivalent to `rm FILE && git add FILE`, is it not?

PR #10405, opened 2026-07-01
reparent incorrect children of Mondo:0700105

katiermullen:
Noticed some terms were incorrectly classified as children of difference of sexual differentiation, non-human animal.

addresses #5726

Robot template: https://docs.google.com/spreadsheets/d/1p_AytMHI4VW4aMHIRIj3ExnO352WcgAKL3U2_J0jXwo/edit?gid=577290559#gid=577290559


sabrinatoro (approved):
looks ok.

PR #10408, opened 2026-07-02
add venom wound terms

katiermullen:
addresses #5726

ROBOT template: https://docs.google.com/spreadsheets/d/1IFaEjcBmU5EWphPW69VKgFAynO_z08M4ax29IDBqJqY/edit?gid=1172407281#gid=1172407281


sabrinatoro (approved):
@katiermullen I approve this PR, 
BUT, I think we should put the non-human animal injury branch outside the disease branch (as we do for human injury). Because these injuries are not diseases. 
We can to this in a different branch.

PR #10465, opened 2026-07-21

ptgolden:


src/ontology/mondo-edit.obo
```diff
@@ -4615,7 +4621,7 @@ synonym: "Jarcho-Levin syndrome" RELATED DEPRECATED [https://rarediseases.org/ra
 synonym: "SCD" RELATED ABBREVIATION [https://rarediseases.org/rare-diseases/jarcho-levin-syndrome/]
 synonym: "SCDO" RELATED ABBREVIATION [https://rarediseases.org/rare-diseases/jarcho-levin-syndrome/]
 synonym: "spondylocostal dysostosis" EXACT [DOID:0050568, https://rarediseases.org/rare-diseases/jarcho-levin-syndrome/, NCIT:C125598, OMIMPS:277300]
-synonym: "Spondylocostal Dysplasia" EXACT [NORD:1308] {OMO:0002001="https://w3id.org/information-resource-registry/nord"}
+synonym: "Spondylocostal Dysplasia" EXACT [NORD:1308, NORD:704] {OMO:0002001="https://w3id.org/information-resource-registry/nord"}
```
sabrinatoro, line 4624: @ptgolden it looks like the preferred names are additive instead of being updated. Can you please check/confirm that it is representative of the information we got in the ClinGen document? Thank you!

ptgolden: I will look at it, but can you explain more what the situation is?

sabrinatoro:
@ptgolden  Before I get in touch with Gioconda (and/or the NORD team), could you please check the following in the file we got from NORD? I think it is the source of the “weird” things and QC failures we are getting here. 

1) Some NORD IDs are associated with 2 different Mondo terms. As a consequence, the preferred names are exact synonyms of 2 different Mondo terms, which triggers a QC failure. 
Can you please confirm that the following NORD IDs are indeed associated with 2 different Mondo terms?

NORD:675 associated with MONDO:0019142 and MONDO:0037939
NORD:278 associated with MONDO:0016098 and MONDO:0019127
NORD:466 associated with MONDO:0010029 and MONDO:0022968

Extra question: I thought we had a check to flag and exclude these cases where one ID is associated with >1 Mondo terms. Am I incorrect, or is it in place and not working?


2) The following Mondo term has a NORD preferred name but no NORD x-ref. Can you please confirm? (This preferred name would have to be removed in the NORD’s file)

“Brown Séquard Syndrome” is a NORD preferred name for MONDO:0008853 but there is no NORD record associated with this Mondo ID.
(Note that this name is currently also a preferred name for MONDO:0003754, which is associated with NORD:950)


3)  There are examples in the diff where the NORD ID that is associated with a Mondo term has been updated, however, the x-ref for the NORD preferred name still has the previous NORD id (in addition to the new one).  
Are the “preferred name” associated with the previous NORD id still in the NORD file? (eg they updated the x-ref but not the preferred name)

For example: 
Screenshot 2026-07-21 at 4 51 34 PM

Screenshot 2026-07-21 at 4 51 16 PM

src/ontology/mondo-edit.obo
sabrinatoro: Please see my comment on the main conversation; it has more explanations.

ptgolden:
> Some NORD IDs are associated with 2 different Mondo terms. As a consequence, the preferred names are exact synonyms of 2 different Mondo terms, which triggers a QC failure. Can you please confirm that the following NORD IDs are indeed associated with 2 different Mondo terms?
>
> NORD:675 associated with MONDO:0019142 and MONDO:0037939
> NORD:278 associated with MONDO:0016098 and MONDO:0019127
> NORD:466 associated with MONDO:0010029 and MONDO:0022968

In the latest dump, yes I confirm that I see all of these.

> There are examples in the diff where the NORD ID that is associated with a Mondo term has been updated, however, the x-ref for the NORD preferred name still has the previous NORD id (in addition to the new one). Are the “preferred name” associated with the previous NORD id still in the NORD file? (eg they updated the x-ref but not the preferred name)

Ah, I see what you mean now. This is, yes, how the pipeline currently works: synonym attributions are additive. Note that I am almost certain that this would be fixed by running the synonym evidence sync workflow.

> The following Mondo term has a NORD preferred name but no NORD x-ref. Can you please confirm? (This preferred name would have to be removed in the NORD’s file)
>
> “Brown Séquard Syndrome” is a NORD preferred name for MONDO:0008853 but there is no NORD record associated with this Mondo ID. (Note that this name is currently also a preferred name for MONDO:0003754, which is associated with NORD:950)

I'm going to guess this stems from the same issue: NORD previously (incorrectly?) added "Brown Séquard Syndrome" as a synonym to `MONDO:0008853`. In the latest release, they have (correctly?) associated it with `MONDO:0003754`. Until synonyms are synced, the dangling synonym will remain.

ptgolden:
For posterity, here's the history of `NORD:875`, via `obohog`:

 Found 7 events matching "NORD:875" across 1 terms and 4 commits

MONDO:0008853 — Barber-Say syndrome
dba5ef2  2023-12-09  Nico Matentzoglu  Add NORD subset and cross-references (#7016)
    → https://github.com/monarch-initiative/mondo/pull/7016
    + xref: NORD:875 {source="MONDO:NORD"}

b4aadc4  2024-03-29  Nico Matentzoglu  Update NORD subset March 2024 (#7507)
    → https://github.com/monarch-initiative/mondo/pull/7507
    + subset: nord_rare {source="MONDO:NORD", source="NORD:875"}
    + synonym: "Brown Séquard Syndrome" EXACT [NORD:875]

c722bd9  2025-03-19  Trish Whetzel  Externally managed content update (#8870)
    → https://github.com/monarch-initiative/mondo/pull/8870
    - subset: nord_rare {source="MONDO:NORD", source="NORD:875"}
    - xref: NORD:875 {source="MONDO:NORD"}

3d44df3  2025-04-24  Trish Whetzel  [create-pull-request] automated change (#8981)
    → https://github.com/monarch-initiative/mondo/pull/8981
    + subset: nord_rare {source="MONDO:NORD", source="NORD:875"}
    + xref: NORD:875 {source="MONDO:NORD"}

ptgolden:
So there are two issues here.

1. NORD IDs associated with multiple MONDO terms
2. Leftover synonym evidence from errant data in previous NORD releases'

I just tested it, and (2) isn't actually fixed by a synonym sync. Let me investigate some more.

ptgolden:
Two things on the dangling synonyms.

1. The `update-nord` target *only* removes `xref` and `subset`. It does not remove synonyms or synonym evidence.

https://github.com/monarch-initiative/mondo/blob/6b4782bf95fd703f78053cb3f08309f018038441/src/ontology/mondo.Makefile#L710-L712

2. NORD is not part of the synonym-sync workflow.


I'm going to fix (1) immediately, and leave (2) for future work.

matentzn:
> NORD is not part of the synonym-sync workflow.

Its fine for now - dont include it. Its good you flagged it, but EMC beats synonym sync, and it can be only either/or

sabrinatoro:
>Its fine for now - dont include it. Its good you flagged it, but EMC beats synonym sync, and it can be only either/or

@matentzn, for my own understanding, can you confirm that the EMC **does include** updates of the synonyms as per the information in the source's file?
In other words: if NORD has (correctly) updated their files, these preferred names would have been updated through the EMC update. (is that correct?)

ptgolden:
@sabrinatoro

> for my own understanding, can you confirm that the EMC does include updates of the synonyms as per the information in the source's file?

Yes, that is correct. I should not have muddied the waters by talking about synonym sync-- it was just an aside and it's not important.

The issue (which #10474 addresses) is that before merging the new external source file, the synonyms from the previous source file should have been taken out.

ptgolden:
I'm going to close this because it needs to be re-run when fixes are in order.

PR #10471, opened 2026-07-22
merged MONDO:0600011 mild hypophosphatasia with MONDO:1010154 adult hypophosphatasia

### LINKED ISSUE #9867: [Merge]MONDO:0600011 (Mild / adult hypophosphatasia) & MONDO:1010154
**Mondo term (ID and Label)**
MONDO:0600011 mild hypophosphatasia

**Reason for deprecation**
OMIM uses these as synonyms. Perhaps they should be merged?

**Term to be merged with**
MONDO:1010154  adult hypophosphatasia

MeeSiing:
close #9867

MONDO:0600011 mild hypophosphatasia and MONDO:1010154 adult hypophosphatasia represent the same concept according to [OMIM:](https://omim.org/entry/146300)


src/ontology/mondo-edit.obo
```diff
@@ -619276,28 +619265,36 @@ intersection_of: MONDO:0700097 MONDO:0007417 {source="https://orcid.org/0000-000
 [Term]
 id: MONDO:1010154
 name: adult hypophosphatasia
-def: "Adult hypophosphatasia (A-HPP) is a mildform of hypophosphatasia characterized by osteomalacia, chondrocalcinosis, osteoarthropathy, stress fractures duringmiddle age, and dental anomalies." [Orphanet:247676]
+def: "A moderate form of hypophosphatasia characterized by adult onset osteomalacia, chondrocalcinosis, osteoarthropathy, stress fractures , and dental anomalies." [Orphanet:247676]
+comment: The most common form of hypophosphatasia characterized by low alkaline phosphatase, unspecific clinical signs, and typically presents in individuals in adulthood. It is autosomal dominantly inherited.
```
sabrinatoro, line 619269: add [https://www.clinicalgenome.org/affiliation/40065/, PMID:32973344] at the end of the comment to keep provenance for this information

sabrinatoro (approved):
I was questioning the need for "single child" exclusion from QC for adult hypophosphatasia - MONDO:1010154. 
This term is a single child of 'ALPL-related autosomal dominant hypophosphatasia' - MONDO:0100608 because it is (1) autosomal dominant, (2) caused by variation in ALPL (which is the case for all hypophosphatasia)

I think all adult hypophosphatasia, however, other types could be dominant OR recessive, which is why adult hypophosphatasia is the only child

PR #10474, opened 2026-07-22
When updating NORD, strip all synonym axioms before merging

ptgolden:
Before this commit, `update-nord` would strip xref and subset values from the OBO file, but it would leave synonyms, synonym evidence, and community label annotations untouched.  This led to the bug described in <https://github.com/monarch-initiative/mondo/pull/10465>.

This commit adds a SPARQL update query that drops all of these axioms.

We may want to think about a more generic way to achieve this in case any of our other EMC sources contribute (or especially, change) synonyms, but this works for now.

### DIFF
--- a/src/ontology/mondo.Makefile
+++ b/src/ontology/mondo.Makefile
@@ -711,7 +711,9 @@ update-nord:
 	make $(TMPDIR)/external/processed-nord.robot.owl -B
 	grep -vE '^(xref: NORD:|subset: nord_rare)' $(SRC) > $(TMPDIR)/mondo-edit.tmp || true
 	mv $(TMPDIR)/mondo-edit.tmp mondo-edit.obo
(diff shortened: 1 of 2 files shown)


src/ontology/mondo.Makefile
```diff
@@ -711,7 +711,9 @@ update-nord:
 	make $(TMPDIR)/external/processed-nord.robot.owl -B
 	grep -vE '^(xref: NORD:|subset: nord_rare)' $(SRC) > $(TMPDIR)/mondo-edit.tmp || true
 	mv $(TMPDIR)/mondo-edit.tmp mondo-edit.obo
-	$(ROBOT) merge -i $(SRC) -i $(TMPDIR)/external/processed-nord.robot.owl --collapse-import-closure false \
+	$(ROBOT) merge -i $(SRC) --collapse-import-closure false \
```
matentzn, line 719: please make a comment saying that "exclusion/deletion logic is now in two places (grep -v, sparql).

matentzn (approved):

PR #10496, opened 2026-07-29

ptgolden:
This updates all EMC except for NORD. There were issues with their latest dump that we are going to address in the next release cycle.


sabrinatoro (approved):
@ptgolden, thank you for this PR. I approve and merge!

Issue #10205, opened 2026-04-28
[NTR/gene]


galyea123:
**Preferred gene-related syndrome label**
SCN2A-Related Disorders
**Synonyms**
SCN2A related conditions
SCN2A disorders

**Parent term (use [OLS](https://www.ebi.ac.uk/ols/ontologies/mondo), or your favorite ontology browser)**
Mendelian neurodevelopmental disorder

**Definition**

SCN2A-related disorders (SCN2A-RD) are a spectrum of neurodevelopmental conditions caused by variants in the SCN2A gene, which encodes the Nav1.2 sodium channel critical for neuronal function. Initially linked to a mild inherited epilepsy (self-limited familial neonatal–infantile epilepsy, SeLFNIE), SCN2A variants are now known to cause a wide range of phenotypes, from severe early-onset epileptic encephalopathies—such as epilepsy in infancy with migrating focal seizures (EIMFS) and infantile spasms—to later-onset epilepsy or autism spectrum disorder and intellectual disability (ASD/ID) without seizures. Most affected individuals, except those with SeLFNIE, exhibit moderate to severe intellectual disability and significantly impaired adaptive functioning, often accompanied by additional neurological and medical complications such as hypotonia, movement disorders, visual impairment, and feeding difficulties. Clinical severity is broadly associated with the functional effect of the variant on the Nav1.2 channel, with gain-of-function changes typically linked to early severe epilepsy and loss-of-function changes more often associated with ASD/ID, although these relationships are complex and not fully understood.  

**Definition source PMID: 38651838  and PMID: 31924505 and PMID: 28379373
Other relevant references: https://scn2a.org/pdf/scn2arelated_disorders.pdf and https://www.rarechromo.org/media/information/Chromosome%20%202/SCN2A%20related%20conditions%20FTNW.pdf

**Children terms (if applicable)*[seizures, benign familial infantile, 3](https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_0011904?lang=en)
[developmental and epileptic encephalopathy, 11](https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_0013388?lang=en)
[...]

PR #10499, opened 2026-07-30
update definition of SCN2A-related disorder

MeeSiing:
adress #10205

User agreed with the classification of SCN2A-related disorder under 'nervous system disorder'. The definition is updated to reflect that this presentation of this disorder is **predominantly neurodevelopmental**.


sabrinatoro (approved):
I feel like some of the information in the definition should be in the comment instead of the definition. But I don't want to block this PR, we can update later.

PR #10530, opened 2026-08-11
ICD-10-CM chapter XVII (Q00-Q99) alignment: analysis, 9 predicate fixes, 16 new mappings

cmungall:
Aligns Mondo with ICD-10-CM chapter XVII (Q00–Q99, congenital malformations). Three commits: an analysis, a defect fix, and a batch of new mappings. **25 xref changes** in `mondo-edit.obo`.

## Why

Mondo references 494 of the 1067 Q codes, but that overstates the alignment badly. Of ~2880 `xref: ICD10CM:Q*` assertions, only **161 were curated equivalences**. The rest carry inherited Orphanet/DOID provenance (`{source="Orphanet:2695"}`) with *no mapping predicate at all*. The dominant real relationship is many-Mondo-diseases → one coarse ICD bucket, i.e. `skos:broadMatch`.

### 1. `4ca57c324` — analysis (no ontology changes)

`src/ontology/analysis/icd10cm-q-alignment/` — a README plus five TSVs and the generating script. Classifies all 1067 Q codes into scope buckets:

| Bucket | n | In scope |
|---|---:|---|
| DISEASE | 439 | yes |
| RESIDUAL | 357 | no — "unspecified"/"other specified" has no stable extension |
| LATERALITY | 107 | no — Mondo doesn't model laterality |
| CONTAINER | 80 | broadMatch only |
| PHENOTYPE | 72 | policy decision needed (HPO covers these) |
| BLOCK | 12 | broadMatch only |

Codes are bucketed by **label, not hierarchy position**: `Q35 Cleft palate` has five children and is still exactly `MONDO:0016064`.

### 2. `da9438c99` — 9 mapping-predicate defects

Eight `equivalentTo` assertions demoted to `mondoIsNarrowerThanSource` (exports as `skos:broadMatch`), each targeting a block, residual or container code that cannot have an exact equivalent — e.g. `Q00-Q99 Congenital malformations, deformations and chromosomal abnormalities` was asserted equivalent to `MONDO:0021147 disorder of development or morphogenesis`.

Plus `MONDO:0031322 triopia`, which carried `source="MONDO:mondoIsNarrowerThanSource."` — a **trailing period**. The malformed qualifier is unparseable and silently dropped the mapping from the broadMatch SSSOM export.

### 3. `5432246be` — 13 exactMatches + 3 promotions

New `MONDO:equivalentTo` xrefs for 13 Q codes that had none, all matching on the ICD label or an official ICD inclusion term (`Q87.83` Bardet-Biedl, `Q85.83` von Hippel-Lindau, `Q93.52` Phelan-McDermid, `Q44.71` Alagille, …). Plus 3 promotions of provenance-only xrefs already 1:1 with label agreement and carrying Orphanet's `/e` exact qualifier.

Q-code `MONDO:equivalentTo` count: **161 → 169**.

## New convention: xref labels

`xref: ICD10CM:Q87.83 {source="..."}` isn't checkable without a code lookup. Three styles were tested through the real NORM pipeline on the full 44 MB edit file:

| Style | Survives NORM? |
|---|---|
| `xref: ICD10CM:Q87.83 {source="..."} ! Bardet-Biedl syndrome` | ❌ **silently stripped** |
| `xref: ICD10CM:Q87.86 "Kleefstra syndrome"` | ✅ preserved |
| `xref: ICD10CM:Q93.52 "Phelan-McDermid syndrome" {source="..."}` | ✅ both preserved |

The trailing `!` comment familiar from `is_a` lines does **not** work on xrefs. The OBO-native quoted description does, round-tripping to `Annotation(rdfs:label "...")` on the `hasDbXref` axiom.

This repo had **zero** prior uses of the quoted form, so this PR introduces it (25 uses). Flagging explicitly in case reviewers want a different convention before it spreads.

## Worth knowing

**Superseded-parent pattern.** 10 of the 13 newly-mapped terms already xreffed the *parent* ICD code. ICD-10-CM has been carving named syndromes out of residual buckets like `Q87.8` into their own codes, and Mondo's inherited mappings still point at the superseded parent. Those parent xrefs are **left untouched** here — being unqualified they export as `oboInOwl:hasDbXref`, not a semantic mapping, so they don't contradict the new exactMatch. Retiring or demoting them belongs with the bulk predicate pass.

**Not addressed:** the 2412 provenance-only xrefs that should carry `broadMatch`, 142 new-term candidates, and the phenotype policy question (72 codes where HPO has the term and Mondo doesn't — e.g. `Q13.5 Blue sclera` = `HP:0000592`). The phenotype decision gates the largest downstream chunk.

## Validation

`make NORM` (run natively as `owltools --merge-axiom-annotations` + `robot convert`, since docker was unavailable) applied cleanly at each step; `robot convert` parses; stanza count unchanged at 36005; the diff is exactly the intended lines with no reformatting churn.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cmungall:
I wasn't done with this and made the PR too early closing for now to avoid confusion [human]

PR #10534, opened 2026-08-11
Add SNAP25-related developmental and epileptic encephalopathy

### LINKED ISSUE #10532: MONDO term request: no class exists for SNAP25-related developmental and epileptic encephalopathy
## Request

MONDO has no class for **SNAP25-related developmental and epileptic encephalopathy**. dismech curates this entity (`kb/disorders/SNAP25_Encephalopathy.yaml`) and currently has to leave `disease_term` unbound.

Requesting a new class, child of `MONDO:0100062` genetic developmental and epileptic encephalopathy, gene-anchored via `RO:0004003` to `HGNC:11132` SNAP25.

## Why nothing existing fits

Verified with OAK against the local `sqlite:obo:mondo` build:

```
$ uv run runoak -i sqlite:obo:mondo relationships --direction down HGNC:11132
subject         predicate    object      subject_label                      object_label
MONDO:0014590   RO:0004003   HGNC:11132  congenital myasthenic syndrome 18  SNAP25
```

`MONDO:0014590` (CMS18) is the **only** SNAP25-anchored term in MONDO. It is not a substitute: the entity curated here is a DEE defined by intellectual disability, early-onset epilepsy, movement disorder, cerebral visual impairment and brain atrophy, and it does not curate myasthenia. Binding a DEE to a congenital myasthenic syndrome would misstate the disease.

The two candidate near-misses and why each is wrong:

| Term | Why not |
|---|---|
| `MONDO:0014590` congenital myasthenic syndrome 18 | SNAP25-anchored, but a presynaptic CMS (`is_a MONDO:0700466`), not a DEE |
| `MONDO:0032678` developmental and epileptic encephalopathy, 71 | **Not a SNAP25 entity.** `is_a MONDO:0600001` glutaminase deficiency; xref `OMIM:618328`, `Orphanet:557064` — this is the *GLS* entity |

`MONDO:0032678` is worth flagging on its own. It was previously bound to the dismech SNAP25 entry, almost certainly by matching the shared label phrase "developmental and epileptic encephalopathy" — a Named Entity Confusion that no anti-hallucination check catches, because the term is real, its label matches, and it validates. That entry is now correctly anchored on `kb/disorders/Neonatal_Epileptic_Encephalopathy_Due_to_Glutaminase_Deficiency.yaml` (fixed in monarch-initiative/dismech#8143).

## Supporting literature

`PMID:33299146` — Klöckner et al., *De novo variants in SNAP25 cause an early-onset developmental and epileptic encephalopathy.* Establishes the entity and its phenotype.

## Note on OMIM

OMIM currently represents the SNAP25 phenotype only as **#616330 (CMS18)**. So this may need to be a MONDO-native class rather than an OMIM-derived one, unless OMIM splits the DEE presentation separately. Flagging in case that affects how you'd prefer to mint it.

A claim that #616330 has been relabelled "DEE117" circulated in a deep-research report during curation of monarch-initiative/dismech#7673; MONDO does not corroborate it and it was rejected there. Repeating it here only so it doesn't resurface as evidence.

## Interim state in dismech

`SNAP25_Encephalopathy.yaml` now carries `disease_term` with a `preferred_term` and **no** `term:` binding. An empty mapping is preferable to a placeholder that overstates specificity — the previous stand-in was `MONDO:0100062`, a correct but non-specific parent shared with the umbrella entry `Genetic_Developmental_and_Epileptic_Encephalopathy`.

Once a class exists, rebind. Until then, please do **not** rebind to `MONDO:0014590` merely because it is the only SNAP25-specific term available.

/cc @matentzn — assigning to you for MONDO triage.

matentzn:
Fixes #10532.

I checked:

1. the pattern (synonyms, definition, logical axiom)
2. The PMID (in support)

# Areas of uncertainty:

1. I added some clarifying remarks into the definition that are not part of the standard pattern but help distinguishing the disease
2. I added `relationship: has_material_basis_in_germline_mutation_in http://identifiers.org/hgnc/11132 {source="PMID:33299146"} ! SNAP25` separately from the logical definition - not sure if this is SOP.


sabrinatoro:
@matentzn (I also updated the main GH issue)

I don't think the proposed change is correct. 
I am not convinced that the 2 diseases ("congenital myasthenic syndrome 18" and "developmental and epileptic encephalopathy-117") are distinct and therefore require 2 distinct terms.

Also, if I were to "agree" to a new term, this term should have the OMIM xref

sabrinatoro (changes requested):
see other comment

matentzn:
Thank you @sabrinatoro - I will close the PR for now until we sort out what needs doing on the issue.

PR #10544, opened 2026-08-13
Add ClinGen requested terms as exact synonyms

MeeSiing:
address #9994 close #10547

added requested terms as exact synonym
updated definitions
added more parent terms to MONDO:0060583 'platelet abnormalities with eosinophilia and immune-mediated inflammatory disease'
[Note](https://docs.google.com/spreadsheets/d/1FWjUMueTR1FBkVwWUwL0XEFoae9f4_YVyPmF4U1kfKA/edit?gid=1007216806#gid=1007216806)

MONDO:0100083 has two ClinGen preferred labels by different GCEP

PR #10546, opened 2026-08-13
 update qc-clingen sparql

### LINKED ISSUE #10545: QC check for single ClinGen preferred label failed
The SPARQL quality control test ([qc-clingen.sparql](https://mondo.readthedocs.io/en/latest/editors-guide/quality-control-tests/#qc-clingensparql)) is designed to ensure that each Mondo term has no more than one ClinGen preferred label. However, in PR #10544, MONDO:0100083 contains two ClinGen preferred labels, but the QC check failed to trigger.

MeeSiing:
close #10545

I asked Claude to update the qc-clingen sparql and check it using my PR term. The updated sparql seems to work now and it found a few more terms that have more than one preferred label. I updated those terms to make sure there are only only preferred label per term. Details below:

The corrected query ran successfully over the full PR #10544 file (exit 0) — it works. Here are the results.

✅ Confirmation for your PR term

MONDO:0100083 → 2 violations — both ClinGen preferred labels are correctly flagged:
- "RUNX1-related hereditary thrombocytopenia and hematological cancer predisposition syndrome"
- "hereditary thrombocytopenia and hematological cancer predisposition syndrome associated with RUNX1"

So the fix does exactly what #10545 asks: it now catches the two-ClinGen-label case that PR #10544 slipped through.

Full blast radius — 7 violations across 4 terms

Running it over the whole ontology also surfaced 3 pre-existing violations the broken check never caught:
Screenshot 2026-08-13 at 3 38 40 PM

This also confirms both rules of the check now fire: block 1 (two ClinGen labels) and block 2 (ClinGen preferred label must be an exact synonym — MONDO:0975761 has it on a narrow synonym).

⚠️ Key implication before committing/merging the QC fix

These 3 other terms are already in master, so once this corrected check is active, CI's ontology_qc will fail on them too — the QC-fix PR cannot go green until those pre-existing violations are also resolved (or the terms fixed in the same/parallel PR). So the fix "works" — arguably too well until the backlog is cleaned up. Worth noting in #10545 and coordinating with the Mondo team on how to handle the 3 pre-existing cases.

### DIFF
--- a/src/sparql/qc/mondo/qc-clingen.sparql
+++ b/src/sparql/qc/mondo/qc-clingen.sparql
@@ -10,22 +10,26 @@ prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
 
 # description: This QC check ensures that no class ever has two clingen labels, and that there are only exact preferred labels
 
-SELECT ?entity ?property ?value
+SELECT DISTINCT ?entity ?property ?value
 WHERE {
  {
   ?ax owl:annotatedSource ?entity ;
          owl:annotatedProperty ?property ;
-         OMO:0002001 <https://w3id.org/information-resource-registry/clingen> .
+         OMO:0002001 ?clingen1 .
   ?ax2 owl:annotatedSource ?entity ;
          owl:annotatedProperty ?property ;
          owl:annotatedTarget ?value ;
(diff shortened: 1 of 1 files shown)


ptgolden (approved):
Good catch!

PR #10567, opened 2026-08-19

matentzn:


ptgolden:
@sabrinatoro The biggest change here is that NORD content here now (finally) reflects the updated dump, with changes made since November.

One thing worth noting, however, is that there are a number of removals from the NORD subset because of a QC issue that must be fixed upstream by NORD. These are the cases we identified last month where one NORD term maps to multiple MONDO terms.

All QC issues are detailed at https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/external/nord-qc-failures.md, but I am attaching a file here that just describes this case.

[nord-mondo-collisions.tsv](https://github.com/user-attachments/files/31235426/nord-mondo-collisions.tsv)

sabrinatoro (approved):

sabrinatoro:
Thank you @ptgolden. I approved and merged this PR. 
Have you already communicated the QC failures to the NORD team?

ptgolden:
I will send an email this evening.

I'm glad we are getting back in sync 👍

PR #10594, opened 2026-08-28
Fix casing in template names

MeeSiing:
updated all template names to sentence case, shortened 2 add new term templates


sabrinatoro (approved):
Thank you @MeeSiing

Issue #10604, opened 2026-08-31
[Merge]MONDO:0100307 adult Refsum disease due to PEX7 defect with MONDO:0013945 peroxisome biogenesis disorder 9B


MeeSiing:
**Mondo term (ID and Label)**
MONDO:0100307 adult Refsum disease due to PEX7 defect

**Reason for deprecation**
duplication based on OMIM: https://omim.org/entry/614879, related to #10377 

**Term to be merged with**
MONDO:0013945 peroxisome biogenesis disorder 9B

**Your nano-attribution (ORCID)**
If you don't have an ORCID, you can sign up for one [here](https://orcid.org/)

PR #10607, opened 2026-08-31
NT for adult Refsum disease and relabel its child term MONDO:0009958

### LINKED ISSUE #10377: Request for new term [Adult Refsum Disease] PARENT CONCEPT
### Label

Adult Refsum Disease

### Synonyms

Classic Refsum Disease

### Synonym type

exact

### Definition

"Adult Refsum disease (ARD) is associated with elevated plasma phytanic acid levels, late childhood-onset (or later) retinitis pigmentosa, and variable combinations of anosmia, polyneuropathy, deafness, ataxia, and ichthyosis. Onset of symptoms ranges from age seven months to older than age 50 years. Cardiac arrhythmia and heart failure caused by cardiomyopathy are potentially severe health problems that develop later in life." PMID: [20301527](https://pubmed.ncbi.nlm.nih.gov/20301527)

### Parent term

MONDO_0019053 Peroxisomal disease (or something under this??)

### Children term(s)

MONDO:0009958 (RENAME TO adult Refsum disease due to PHYH or other current exact syn); MONDO:0100307 adult Refsum disease due to PEX7 defect

### ORCID Identifier

_No response_

### Website URL

_No response_

### Additional comments

Requesting this action to align Mondo/MedGen/GeneReviews disease concepts.

- [ ] Rename MONDO:0009958 to indicate the monogenic nature of this concept.
- [ ] Create a new concept for adult refsum disease parent to PHYH and PEX7 subtypes according to https://www.ncbi.nlm.nih.gov/books/NBK1353/#refsum.Molecular_Genetics

MeeSiing:
close #10377

1. Created MONDO:1060256 adult Refsum disease as a gene-agnostic parent under MONDO:0019053 peroxisomal disease.
2. Relabelled MONDO:0009958 to "Refsum disease, classic" because it's with a more specific concept (OMIM:266500)
3. Moved the gene-agnostic mappings and their associated subsets and synonyms to MONDO:1060256
4. Retained OMIM:266500, OMIM:600964 and NCIT:C85043 on MONDO:0009958. NCIt defines Refsum Disease as "caused by mutation(s) in the PHYH gene"
5. updated classification of MONDO:0013945 'peroxisome biogenesis disorder 9B' under MONDO:0019609 Zellweger spectrum disorders. Zellweger is referring the most serious form where the symptoms appear right at birth, and it is often life-threatening in infancy. 'peroxisome biogenesis disorder 9B' is the mild/atypical adult-onset PEX7 phenotype, therefore should be under non-Zellweger
6. Corrected the curator note on MONDO:0009958 to point to omim.org/entry/601539 (peroxisome biogenesis disorder 1B, NALD/IRD); it previously pointed at 266500


sabrinatoro (changes requested):
@MeeSiing I agree with your analysis, but not with the execution. 

In this PR, you have changed the definition, ie the concept of MONDO:0009958 and made it more specific. Therefore, if someone used this Mondo ID in their annotations, their annotations might be incorrect, because they might not have meant to use the PHYH-gene specific term. 

If is OK (in some cases) to change a definition to make it more general (the annotations using the term would be less specific, but not incorrect), but not the opposite. 

What you need to do here is keep MONDO:0009958 as is, and update the x-refs (and the SC. The new term to create would be the more specific term. 

IF the logical and textual definitions of MONDO:0009958 conflict (I don't think it is the case here), then we would need to "split" the term - ie obsolete and create 2 new terms to replace this one term.

MeeSiing:
@sabrinatoro Thank you for the reminder. I've reverted MONDO:0009958 adult Refsum disease and created new term for PHYH-gene specific term.

Other changes:
- updated the synonym scope of MONDO:0009958 adult Refsum disease
- removed the definition and SCof of MONDO:0013945, to avoid the duplication problem (this will be addressed in the merge issue #10604 )

sabrinatoro (approved):
This looks good. I approve and merge this PR

PR #10610, opened 2026-09-01
release September 2026

sabrinatoro:


MeeSiing:
@sabrinatoro Everything looks good except I have a question for **/src/ontology/reports/mondo_obsoletioncandidates.tsv**. Should this file contain all the terms that are tagged with the merge comment? I remember we have 3 obsoletion PRs with a total of 6 terms but I only see 1 term here. 
There are 6 terms in this file :/reports/obsoletion-candidates.tsv

MeeSiing (approved):
Thank you for doing this. Approve and will merge.

PR #10613, opened 2026-09-02
reparent coccydynia to vertebral column disorder

### LINKED ISSUE #9511: [Revise subclass] MONDO:0022792 coccygodynia
**Mondo term (ID and Label)**
MONDO:0022792 coccygodynia

no evidence to support 'neurological pain disorder' as a parent term for MONDO:0022792 coccygodynia. Mondo does not currently have "spinal pain disorder" which is the direct parent of coccygodynia in ICD foundation #9339 

MeeSiing:
close #9511

- reparented coccydynia to vertebral column disorder using SNOMED concept which placed SCIT:34789001  "Pain in coccyx") under SCIT:48926009 "Pain in spine "
- excluded neurological pain disorder as a parent because its common cause is injury, and there is no evidence to support that neurological pain disorder as parent.
-- only ORCID is used as source because there was no provenance for this parent assertion


sabrinatoro (approved):
This looks great. 
Thank you @MeeSiing

PR #10622, opened 2026-09-05
obsolete autosomal dominant optic atrophy and reparent the child terms

### LINKED ISSUE #9831: MONDO:0020250 autosomal dominant optic atrophy
We are reviewing terms with Orphanet IDs that have been obsoleted #7693 

We noticed that MONDO:0014753 autosomal recessive optic atrophy is already obsolete, while MONDO:0020250 autosomal dominant optic atrophy is still active. Autosomal dominant optic atrophy is used as a grouping term.

We are considering obsoleting autosomal dominant optic atrophy as well, for consistency.

@kanems and Eric @GARDInfoCenter, we would appreciate your thoughts.

@courtneythaxton, could you please share this with the retinal or eye working group and let us know their feedback.

Thank you.

@sabrinatoro

MeeSiing:
close #9831


sabrinatoro (approved):
This looks good. 
Note that the "in_subset" and "conformsTo" annotations are still there, but these will be updated automatically through the next subset update, so we can leave it as is.

Approving and merging

PR #10669, opened 2026-09-17
merge 'mirror movements 1' - MONDO:0008002 and 'mirror movements 1 an…

### LINKED ISSUE #9764: [Merge]mirror movements 1
**Mondo term (ID and Label)**
MONDO:0008002  mirror movements 1 

**Reason for deprecation**
OMIM treats these two as synonyms. Also mirror movements 1 is the only child to mirror movements 1 and/or agenesis of the corpus callosum.

**Term to be merged with**
mirror movements 1 and/or agenesis of the corpus callosum
 MONDO_0100515

sabrinatoro:
…d/or agenesis of the corpus callosum' - MONDO:0100515

Closes #9764

Note that the logical definition was not added because (1) the "familial congenital mirror movements" genus is not correct anymore, (2) there are other nervous system disease caused by germline variation in DCC


MeeSiing (approved):
This looks great. I updated the synonym scope. Thank you.

PR #10670, opened 2026-09-17
relabel MONDO:0035694 to immune deficiency and dysregulation due to RELA insufficiency

### LINKED ISSUE #10493: Rename MONDO:0035694 request from ClinGen SCID-CID GCEP
Mondo term (ID and Label): combined immunodeficiency due to RELA haploinsufficiency (MONDO:0035694)

Suggested new label: immune deficiency and dysregulation due to RELA insufficiency

Your nano-attribution (ORCID) 0000-0001-5684-5457

_Optional_: Any additional information (like supporting evidence, PubMed ID, etc.):  This request is being made by the ClinGen SCID-CID GCEP.  Do NOT merge with 'MONDO:0032659 mucocutaneous ulceration, chronic'', which is icaused by the underlying immune dysregulation.  Inclusion of the term 'dysregulation' will capture this basis.  Also, literature (PMID: [37273177](https://pubmed.ncbi.nlm.nih.gov/37273177/)) indicates that dominant-negative variants can cause the same disease, so please remove 'haplo' and just leave the term as 'insufficiency'.

MeeSiing:
close #10493

relabeled MONDO:0035694 to ClinGen request name because the original label is from Orphanet which has already been updated


sabrinatoro (commented):
Looks good! 
Approving and merging.

sabrinatoro (approved):

